PR #7, opened 2024-01-14
Schema cleanup

kcsongor:
This PR cleans up the schema a little bit. The commits are best reviewed separately.

### DIFF
--- a/src/EndpointManager.sol
+++ b/src/EndpointManager.sol
@@ -32,8 +32,8 @@ abstract contract EndpointManager is IEndpointManager, OwnableUpgradeable, Reent
 
     uint64 _sequence;
 
-    constructor(address token, Mode mode, uint16 chainId) {
-        _token = token;
+    constructor(address tokenAddress, Mode mode, uint16 chainId) {
+        _token = tokenAddress;
         _mode = mode;
         _chainId = chainId;
         _evmChainId = block.chainid;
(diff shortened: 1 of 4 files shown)


src/libraries/EndpointStructs.sol
```diff
@@ -16,11 +16,8 @@ struct EndpointManagerMessage {
 struct NativeTokenTransfer {
     /// @notice Amount being transferred (big-endian uint256)
     uint256 amount;
-    /// @notice Address of the token. Left-zero-padded if shorter than 32 bytes
-    // TODO: this field is not needed.
-    bytes32 tokenAddress;
-    /// @notice Address of the recipient. Left-zero-padded if shorter than 32 bytes
-    bytes32 to;
+    /// @notice Address of the recipient.
+    bytes to;
```
nik-suri, line 20: Nice adding the payload and recipient address lengths. Perhaps we should add them in the structs as well?

I imagine people will reference these structs when building additional services -- so add `NativeTokenTransfer.recipientLength` and `EndpointManagerMessage.payloadLength`?

kcsongor: That's a good point. I think we should instead include a comment explaining what the wire format looks like -- I don't feel very good adding that field into the struct because that introduces newly representable invalid states that we then have to think about (for example one could create a struct where the length field doesn't actually match the length of the address)

src/EndpointManager.sol
```diff
@@ -207,9 +208,8 @@ abstract contract EndpointManager is IEndpointManager, OwnableUpgradeable, Reent
         pure
         returns (bytes memory encoded)
     {
-        // TODO -- should we check payload length here?
-        // for example, CCTP integration checks payload is <= max(uint16)
-        return abi.encodePacked(m.chainId, m.sequence, m.msgType, m.payload);
+        uint16 payloadLength = uint16(m.payload.length);
+        return abi.encodePacked(m.chainId, m.sequence, m.msgType, payloadLength, m.payload);
```
nik-suri, line 212: Should we validate the payload length is `<= max(uint16)`, similar to the CCTP contracts?

src/EndpointManager.sol
```diff
@@ -236,7 +237,8 @@ abstract contract EndpointManager is IEndpointManager, OwnableUpgradeable, Reent
         pure
         returns (bytes memory encoded)
     {
-        return abi.encodePacked(m.amount, m.tokenAddress, m.to, m.toChain);
+        uint16 toLength = uint16(m.to.length);
```
nik-suri, line 240: same here -- should we enforce a max length on recipient addresses?

src/EndpointManager.sol
kcsongor: yes good point

kcsongor: done

nik-suri (approved):

PR #8, opened 2024-01-14
Improve standalone arch

kcsongor:
This makes `_executeMsg` perform replay protection, so that the different configurations of endpoints and manager don't have to independently reimplement this logic.

Also moved the struct util functions into a library, so the commits are best view separately

### DIFF
--- a/src/EndpointManager.sol
+++ b/src/EndpointManager.sol
@@ -25,6 +25,17 @@ abstract contract EndpointManager is IEndpointManager, OwnableUpgradeable, Reent
         BURNING
     }
 
+    // @dev Information about attestations for a given message.
+    struct AttestationInfo {
+        // whether this message has been executed
+        bool executed;
+        // bitmap of endpoints that have attested to this message (NOTE: might contain disabled endpoints)
+        uint64 attestedEndpoints;
+    }
+
+    // Maps are keyed by hash of EndpointManagerMessage.
+    mapping(bytes32 => AttestationInfo) public managerMessageAttestations;
+
     address immutable _token;
     Mode immutable _mode;
     uint16 immutable _chainId;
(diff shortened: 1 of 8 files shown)


src/EndpointManager.sol
```diff
@@ -149,12 +160,27 @@ abstract contract EndpointManager is IEndpointManager, OwnableUpgradeable, Reent
         return amount;
     }
 
+    // @dev Mark a message as executed.
+    // This function will revert if the message has already been executed.
+    function _replayProtect(bytes32 digest) internal {
+        // check if this message has already been executed
+        if (managerMessageAttestations[digest].executed) {
```
nik-suri, line 167: nit -- reuse `isMessageExecuted(digest)` here?

src/EndpointManager.sol
```diff
@@ -25,6 +25,17 @@ abstract contract EndpointManager is IEndpointManager, OwnableUpgradeable, Reent
...
+    // @dev Information about attestations for a given message.
+    struct AttestationInfo {
+        // whether this message has been executed
+        bool executed;
+        // bitmap of endpoints that have attested to this message (NOTE: might contain disabled endpoints)
+        uint64 attestedEndpoints;
+    }
+
+    // Maps are keyed by hash of EndpointManagerMessage.
+    mapping(bytes32 => AttestationInfo) public managerMessageAttestations;
```
nik-suri, line 37: The `EndpointManager` doesn't use `AttestationInfo.attestedEndpoints` -- I'm wondering whether the replay protection in the `EndpointManager` should be a `mapping(bytes32 => bool)` to just track which digests have been executed, and then the standalone contract could implement multi-sig threshold logic on top of this by keeping another storage variable to track `mapping(bytes32 => uint64) digestAttestedEndpoints`.

Agreed we need better naming for these contracts. `EnpointManagerStandalone` should really be named `EndpointManager`, and `EndpointManager` should be named something like `BaseAbstractMessenger` since it's really becoming responsible for `sendMessage`, `quoteDeliveryPrice`, and `_executeMsg` (these methods will also become external once we pull out the business logic for transferring tokens into a separate contract).

src/EndpointManager.sol
kcsongor: done in https://github.com/wormhole-foundation/example-native-token-transfers/compare/dc94fdb2a22d275407c11dde109f6d8cbcb393bf..84af9b429c9282f8edc910f7d5a2382d58634eed

src/EndpointManager.sol
kcsongor: Yeah, this is a little annoying, and separating the two would definitely be the cleanest. It would however add an extra 20k gas overhead on each message in the threshold case (since the endpoint attestations and the replay protection would now be stored in different slots).

nik-suri: Ok, we can also revisit this later once we complete more refactoring and think on whether this trade-off is worth it.

nik-suri (approved):

PR #9, opened 2024-01-15
EndpointManager: custom storage slots

kcsongor:
The main purpose of this PR is to use deterministic storage slots everywhere in `EndpointManager`, but it includes some bug fixes and refactoring too.

### DIFF
--- a/src/EndpointManager.sol
+++ b/src/EndpointManager.sol
@@ -1,7 +1,6 @@
 // SPDX-License-Identifier: Apache 2
 pragma solidity >=0.6.12 <0.9.0;
 
-import "openzeppelin-contracts/contracts/security/ReentrancyGuard.sol";
 import "openzeppelin-contracts/contracts/token/ERC20/IERC20.sol";
 import "openzeppelin-contracts/contracts/token/ERC20/utils/SafeERC20.sol";
 import "openzeppelin-contracts/contracts/token/ERC20/extensions/ERC20Burnable.sol";
@@ -10,6 +9,7 @@ import "wormhole-solidity-sdk/Utils.sol";
 import "wormhole-solidity-sdk/libraries/BytesParsing.sol";
 
 import "./libraries/external/OwnableUpgradeable.sol";
+import "./libraries/external/ReentrancyGuardUpgradeable.sol";
 import "./libraries/EndpointStructs.sol";
 import "./libraries/EndpointHelpers.sol";
 import "./interfaces/IEndpointManager.sol";
@@ -17,7 +17,7 @@ import "./interfaces/IEndpointToken.sol";
 import "./Endpoint.sol";
 
 // TODO: rename this (it's really the business logic)
-abstract contract EndpointManager is IEndpointManager, OwnableUpgradeable, ReentrancyGuard {
+abstract contract EndpointManager is IEndpointManager, OwnableUpgradeable, ReentrancyGuardUpgradeable {
     using BytesParsing for bytes;
     using SafeERC20 for IERC20;
 
@@ -54,6 +54,7 @@ abstract contract EndpointManager is IEndpointManager, OwnableUpgradeable, Reent
     function initialize() public initializer {
         // TODO: shouldn't be msg.sender but a separate (contract) address that's passed in the initializer
         __Ownable_init(msg.sender);
+        __ReentrancyGuard_init();
     }
 
     /// @dev This will either cross-call or internal call, depending on whether the contract is standalone or not.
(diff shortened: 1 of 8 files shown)


src/EndpointManagerStandalone.sol
```diff
@@ -4,55 +4,68 @@ pragma solidity >=0.6.12 <0.9.0;
...
-            revert CallerNotEndpoint(msg.sender);
-        }
-        _;
+        _Threshold storage _threshold = _getThresholdStorage();
+        // We increase the threshold here. This might not be what the user
+        // wants, in which case they can call setThreshold() afterwards.
+        // However, this is the most sensible default behaviour, since
+        // this makes the system more secure in the event that the user forgets
+        // to call setThreshold().
+        _threshold.num += 1;
```
djb15, line 53: I like this default behaviour!

src/EndpointManagerStandalone.sol
```diff
@@ -4,55 +4,68 @@ pragma solidity >=0.6.12 <0.9.0;
...
     }
 
-    constructor(address token, Mode mode, uint16 chainId) EndpointManager(token, mode, chainId) {
-        _checkEndpointsInvariants();
+    function removeEndpoint(address endpoint) external onlyOwner {
```
djb15, line 56: Do we want to prevent removing an endpoint if 0 endpoints remain after the action? Not sure if that's something we should care about, seems like an admin responsibility.

src/EndpointRegistry.sol
```diff
@@ -0,0 +1,261 @@
...
+        uint8 index;
+    }
+
+    /// @dev Bitmap encoding the enabled endpoints.
+    /// invariant: forall (i: uint8), enabledEndpointBitmap & i == 1 <=> endpointInfos[i].enabled
+    struct _EnabledEndpointBitmap {
+        uint64 bitmap;
+    }
+
+    /// @dev Total number of registered endpoints. This number can only increase.
```
djb15, line 25: This comment is not true since we can remove endpoints?

src/EndpointRegistry.sol
```diff
@@ -0,0 +1,261 @@
...
+    /// invariant: numRegisteredEndpoints <= MAX_ENDPOINTS
+    /// invariant: forall (i: uint8),
+    ///   i < numRegisteredEndpoints <=> exists (a: address), endpointInfos[a].index == i
+    struct _NumRegisteredEndpoints {
+        uint8 num;
+    }
+
+    // TODO: should we just increase this to 255?
+    uint8 constant MAX_ENDPOINTS = 64;
```
djb15, line 34: At first glance I don't see a problem increasing to 255? But tbh even 64 endpoints sounds kind of crazy so I'm happy to leave this as is

src/EndpointManager.sol
```diff
@@ -33,16 +40,40 @@ abstract contract EndpointManager is IEndpointManager, OwnableUpgradeable, Reent
...
+
+    bytes32 public constant SEQUENCE_SLOT = bytes32(uint256(keccak256("ntt.sequence")) - 1);
+
+    function _getMessageAttestationsStorage()
+        internal
+        pure
+        returns (mapping(bytes32 => AttestationInfo) storage $)
+    {
+        uint256 slot = uint256(MESSAGE_ATTESTATIONS_SLOT);
+        assembly {
```
djb15, line 60: Make these slots "memory-safe"?

djb15 (approved):
Couple of small comments, most non-specific to the custom storage slots. Happy with the updated storage slot logic though

src/EndpointRegistry.sol
kcsongor: this is "registered" but not necessarily "enabled" (the latter of which can go down indeed). we keep the disabled ones to ensure their index is not reused

src/EndpointManagerStandalone.sol
nik-suri: Imo it's okay -- say someone wants to fully replace their endpoints. They might want to reduce to 0, then re-add them.

I think we want to minimize cases where someone could shoot themselves in the foot and brick the contract, but 0 endpoints should be okay and easily fixable.

src/EndpointRegistry.sol
kcsongor: will revisit this later

src/EndpointManagerStandalone.sol
kcsongor: this is actually a good point, because if the endpoints are all removed, then the threshold becomes 0, and any message is immediately accepted. Agree with @nik-suri that we should allow removing endpoints first then re-introducing them.
But we don't want to open up a time period where any message is accepted, so I added check https://github.com/wormhole-foundation/example-native-token-transfers/pull/9/commits/3b07332f6b2638c1e24735f929296917c57e9423 to prevent this

nik-suri: nice catch @kcsongor !

djb15 (approved):

PR #10, opened 2024-01-15
EndpointManager: Add global outbound rate limits

nik-suri:
Partially addresses https://github.com/wormhole-foundation/example-native-token-transfers/issues/15

### DIFF
--- a/src/EndpointManager.sol
+++ b/src/EndpointManager.sol
@@ -51,6 +51,9 @@ abstract contract EndpointManager is
 
     bytes32 public constant SEQUENCE_SLOT = bytes32(uint256(keccak256("ntt.sequence")) - 1);
 
+    bytes32 public constant OUTBOUND_LIMIT_PARAMS_SLOT =
+        bytes32(uint256(keccak256("ntt.outboundLimitParams")) - 1);
+
     function _getMessageAttestationsStorage()
         internal
         pure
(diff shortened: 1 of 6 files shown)


src/EndpointManager.sol
```diff
@@ -64,6 +78,76 @@ abstract contract EndpointManager is IEndpointManager, OwnableUpgradeable, Reent
...
+    }
+
+    /**
+     * @dev Updates the current capacity
+     *
+     * @param newLimit The new limit
+     * @param oldLimit The old limit
+     * @param currentLimit The current limit
+     */
+    function _calculateNewCurrentCapacity(
```
djb15, line 126: Is this method copy pasted from xERC20 or new logic? I'd like to see tests for updating the limits in both directions when the current limit isn't 100% of the oldLimit (i.e. a transfer has consumed some of the limit)

src/EndpointManager.sol
```diff
@@ -43,6 +43,20 @@ abstract contract EndpointManager is IEndpointManager, OwnableUpgradeable, Reent
...
+    uint256 private constant _RATE_LIMIT_DURATION = 1 days;
+
+    struct RateLimitParams {
+        uint256 limit;
+        uint256 currentCapacity;
+        uint256 lastTxTimestamp;
+        uint256 ratePerSecond;
+    }
+
+    RateLimitParams outboundLimitParams;
```
djb15, line 58: Deterministic storage slot?

djb15 (changes requested):
Should this PR also include the ability to enqueue a transfer if the outgoing rate limit has been hit, or is that coming in a future PR?

src/EndpointManager.sol
nik-suri: Pretty much copy-pasted, just changed some variable names (imo using `currentCapacity` is a bit more descriptive than `currentLimit`).

I'll add tests for updating limits.

djb15: Cool thanks. I just want to be sure that at the new rate the limits are filling to the correct max. Since it's mainly copy pasted there should hopefully be some tests from them you can repurpose as well

nik-suri:
@djb15 I will write the outbound queuing feature in a follow up PR.

djb15 (approved):

PR #11, opened 2024-01-16
Upgradeability

kcsongor:


src/libraries/external/ReentrancyGuardUpgradeable.sol
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    // keccak256(abi.encode(uint256(keccak256("openzeppelin.storage.ReentrancyGuard")) - 1)) & ~bytes32(uint256(0xff))
+    bytes32 private constant ReentrancyGuardStorageLocation =
+        0x9b779b17422d0df92223018b32b4d1fa46e071723d6817e2486d003becc55f00;
+
+    function _getReentrancyGuardStorage() private pure returns (ReentrancyGuardStorage storage $) {
+        assembly {
+            $.slot := ReentrancyGuardStorageLocation
+        }
```
RahulMaganti47, line 51: maybe useful to have the `memory-safe` annotation here?

kcsongor: ah this is actually not added in this PR, just need to rebase

RahulMaganti47: ah o

src/wormhole/Governance.sol
```diff
@@ -0,0 +1,92 @@
...
+    IWormhole immutable wormhole;
+
+    error InvalidGovernanceChainId(uint16 chainId);
+    error InvalidGovernanceContract(bytes32 contractAddress);
+    error GovernanceActionAlreadyConsumed(bytes32 digest);
+
+    error NotRecipientChain(uint16 chainId);
+    error NotRecipientContract(address contractAddress);
+
+    bytes32 constant CONSUMED_GOVERNANCE_ACTIONS_SLOT =
```
djb15, line 16: Why is this slot hash calculated differently from every other slot?

src/wormhole/Governance.sol
```diff
@@ -0,0 +1,92 @@
...
+    function performGovernance(bytes calldata vaa) external {
+        IWormhole.VM memory verified = _verifyGovernanceVAA(vaa);
+        GovernanceMessage memory message = abi.decode(verified.payload, (GovernanceMessage));
+
+        if (message.chainId != wormhole.chainId()) {
+            revert NotRecipientChain(message.chainId);
+        }
+        if (message.governanceContract != address(this)) {
+            revert NotRecipientContract(message.governanceContract);
+        }
```
djb15, line 50: Can we get rid of this check and instead add a "module" hash to the `GovernanceMessage` struct. And then verify that `module == keccak256("NativeTokenTransfers")` or something like that?

My thinking is we could then get the guardians to sign a single message with `chainId` of 0 to perform an action that is relevant for every chain (e.g. calling `pause`). We could restrict this to specific actions by checking the first 4 bytes of calldata once we've finalised all the allowed actions of the Manager owner.

Maybe this isn't too important since we won't have loads of deployments, lmk what you think?

src/EndpointAndManager.sol
```diff
@@ -42,7 +43,7 @@ abstract contract EndpointAndManager is Endpoint, EndpointManager {
 
     // override this function to avoid a storage lookup. The endpoint is always enabled
     function _getEnabledEndpointsBitmap() internal pure override returns (uint64) {
-        return 0x1;
+        return uint64(1 << ENDPOINT_INDEX);
```
djb15, line 46: Can you explain the logic behind this change pls?

kcsongor: technically it's identical to before (since `ENDPOINT_INDEX` is 0) but now more consistent with the use of that constant. i.e. if we change `ENDPOINT_INDEX` to 200, it will calculate the bit correctly still

src/wormhole/Governance.sol
kcsongor: That's a good point, there might be actions that make sense on all chains. My original reason for putting the address in was that if this contract ever needs to be migrated to another one that works similarly (perhaps a bug fix), then with the address in, we wouldn't need to worry about replays. Maybe that's just too paranoid?

src/wormhole/Governance.sol
kcsongor: that was a bit of an oversight, fixed here: https://github.com/wormhole-foundation/example-native-token-transfers/compare/48c246a92d11421e1d545711815739f8d61f02e6..2c7096ebe08e650b3b57394d6fa9d3706795a116

the reason it doesn't mention `ntt` is because it's technically independent of `ntt` and this exact same contract could be used to govern other contracts too, since it's fully generic.

djb15: Nice, change looks good. Yeah happy that it's not `ntt`, I agree it's independent

src/wormhole/Governance.sol
djb15: Hmmm good point, that's the downside of not making this contract upgradable. Let's leave it as is then, I'd rather have there be a little more effort for governors than have the potential of replaying these actions.

Or we make this upgradable to solve both? I originally thought this contract would look a bit different but since it's making arbitrary external calls anyway I don't think making this upgradable would introduce any new attack vectors?

src/EndpointManager.sol
```diff
@@ -97,9 +99,10 @@ abstract contract EndpointManager is
         _rateLimitDuration = rateLimitDuration;
     }
 
-    function initialize() public initializer {
+    function __EndpointManager_init() internal onlyInitializing {
         // TODO: shouldn't be msg.sender but a separate (contract) address that's passed in the initializer
         __Ownable_init(msg.sender);
+        // TODO: check if it's safe to not initialise reentrancy guard
```
kcsongor, line 105: this is more of a note to self (we should definitely initialise it, I was just curious if behaves sanely without initialisation)

djb15 (approved):
I'm ignoring the outstanding TODOs for now, the actual changes look good to me

PR #19, opened 2024-01-19
EndpointManager: add outbound transfers queue

nik-suri:
https://github.com/wormhole-foundation/example-native-token-transfers/issues/15

### DIFF
--- a/src/EndpointManager.sol
+++ b/src/EndpointManager.sol
@@ -44,6 +44,10 @@ abstract contract EndpointManager is
         uint64 num;
     }
 
+    struct _QueueSequence {
+        uint64 num;
+    }
+
     /// =============== STORAGE ===============================================
 
     bytes32 public constant MESSAGE_ATTESTATIONS_SLOT =
@@ -53,6 +57,12 @@ abstract contract EndpointManager is
 
     bytes32 public constant OUTBOUND_LIMIT_PARAMS_SLOT =
         bytes32(uint256(keccak256("ntt.outboundLimitParams")) - 1);
+
+    bytes32 public constant OUTBOUND_QUEUE_SLOT =
+        bytes32(uint256(keccak256("ntt.outboundQueue")) - 1);
+
+    bytes32 public constant OUTBOUND_QUEUE_SEQUENCE_SLOT =
+        bytes32(uint256(keccak256("ntt.outboundQueueSequence")) - 1);
 
     function _getMessageAttestationsStorage()
         internal
@@ -74,6 +84,24 @@ abstract contract EndpointManager is
 
     function _getOutboundLimitParamsStorage() internal pure returns (RateLimitParams storage $) {
         uint256 slot = uint256(OUTBOUND_LIMIT_PARAMS_SLOT);
+        assembly ("memory-safe") {
+            $.slot := slot
+        }
+    }
+
+    function _getOutboundQueueStorage()
+        internal
+        pure
+        returns (mapping(uint64 => OutboundQueuedTransfer) storage $)
+    {
+        uint256 slot = uint256(OUTBOUND_QUEUE_SLOT);
+        assembly ("memory-safe") {
+            $.slot := slot
+        }
+    }
+
+    function _getOutboundQueueSequenceStorage() internal pure returns (_QueueSequence storage $) {
+        uint256 slot = uint256(OUTBOUND_QUEUE_SEQUENCE_SLOT);
         assembly ("memory-safe") {
             $.slot := slot
         }
@@ -156,6 +184,14 @@ abstract contract EndpointManager is
         return _getCurrentCapacity(getOutboundLimitParams());
     }
 
+    function getOutboundQueuedTransfer(uint64 queueSequence)
+        public
+        view
+        returns (OutboundQueuedTransfer memory)
+    {
+        return _getOutboundQueueStorage()[queueSequence];
+    }
+
     /**
      * @dev Gets the current capacity for a parameterized rate limits struct
      */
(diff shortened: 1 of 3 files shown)


src/EndpointManager.sol
```diff
@@ -241,8 +319,29 @@ abstract contract EndpointManager is
...
+                _getOutboundQueueStorage()[queueSequence] = OutboundQueuedTransfer({
+                    amount: amount,
+                    recipientChain: recipientChain,
+                    recipient: recipient,
+                    txTimestamp: block.timestamp,
+                    isSet: true
+                });
+
+                // refund the price quote back to sender
+                payable(msg.sender).transfer(totalPriceQuote);
```
djb15, line 337: There's an existing refund if `msg.value` exceeds the quote. I'm thinking that should be moved to later in this method because otherwise we're making 2 refunds which costs extra gas

djb15 (changes requested):
I think the `_transfer` logic should be changed so that tokens are still locked/burned when a transfer is queued. Otherwise I could just queue up an infinite value of transfers at an arbitrary point in time and then I can redeem these queued entries to avoid rate limits.

The current logic also has the side effect that a user could accidentally send their funds to an unintended recipient if they finalise the wrong queued sequence number. Keeping outbound queued transfers non-permissioned is fine provided we have already seized the funds at the time queueing starts.

src/EndpointManager.sol
```diff
@@ -79,6 +89,24 @@ abstract contract EndpointManager is
         }
     }
 
+    function _getOutboundQueueStorage()
+        internal
+        pure
+        returns (mapping(uint64 => OutboundQueuedTransfer) storage $)
```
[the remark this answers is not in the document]
nik-suri: No particular reason -- I had used `uint64` for sequence number since that's the sequence number type used by the Wormhole token bridge.

Perhaps the token bridge used it for compatibility reasons with other components? Don't know all the history here, cc @kcsongor

src/EndpointManager.sol
```diff
@@ -216,8 +283,19 @@ abstract contract EndpointManager is
...
         uint256 amount,
         uint16 recipientChain,
-        bytes32 recipient
+        bytes32 recipient,
+        bool shouldQueue
     ) external payable nonReentrant returns (uint64 msgSequence) {
+        return _transfer(amount, recipientChain, recipient, false, shouldQueue);
+    }
+
+    function _transfer(
```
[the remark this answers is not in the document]
nik-suri: Created this issue -- https://github.com/wormhole-foundation/example-native-token-transfers/issues/24

Will address this in a follow up PR.

src/EndpointManager.sol
```diff
@@ -241,34 +313,62 @@ abstract contract EndpointManager is
...
+
+                _getOutboundQueueStorage()[queueSequence] = OutboundQueuedTransfer({
+                    amount: amount,
+                    recipientChain: recipientChain,
+                    recipient: recipient,
+                    txTimestamp: block.timestamp
+                });
+
+                // refund the price quote back to sender
+                payable(msg.sender).transfer(totalPriceQuote);
```
djb15, line 359: This should refund `msg.value` now

src/interfaces/IEndpointManager.sol
```diff
@@ -19,12 +21,25 @@ interface IEndpointManager {
         uint256 ratePerSecond;
     }
 
+    struct OutboundQueuedTransfer {
```
djb15, line 24: Can you reorder this struct to have the `recipientChain` last please?

djb15 (dismissed):

src/EndpointManager.sol
```diff
@@ -241,34 +313,62 @@ abstract contract EndpointManager is
...
+            bool didConsumeAmount = _consumeOutboundAmount(amount);
+            if (shouldQueue && !didConsumeAmount) {
+                // queue up and return
+                uint64 queueSequence = useOutboundQueueSequence();
+
+                _getOutboundQueueStorage()[queueSequence] = OutboundQueuedTransfer({
+                    amount: amount,
+                    recipientChain: recipientChain,
+                    recipient: recipient,
+                    txTimestamp: block.timestamp
```
RahulMaganti47, line 355: would it better to use ` _getOutboundLimitParamsStorage().lastTxTimestamp` here? or add an `assert` to check that block.timestamp is equivalent to  ` _getOutboundLimitParamsStorage().lastTxTimestamp` when the transfer is added to the queue? if `didConsume` is true, this variable is set here: https://github.com/wormhole-foundation/example-native-token-transfers/tree/0099b82e1daf519efdf1b1748aa014e3cf7b8356/src/EndpointManager.sol

src/EndpointManager.sol
```diff
@@ -241,34 +313,62 @@ abstract contract EndpointManager is
...
+                });
+
+                // refund the price quote back to sender
+                payable(msg.sender).transfer(msg.value);
+
+                // return the sequence in the queue
+                return queueSequence;
+            } else if (!didConsumeAmount) {
+                revert NotEnoughOutboundCapacity(getCurrentOutboundCapacity(), amount);
```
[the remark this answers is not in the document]
nik-suri: Ah, so the way I wrote this was to have `didConsumeAmount == true` when the transaction can go through, and have `didConsumeAmount == false` when the transaction can't go through (hits a rate limit).

Maybe naming is the issue here -- instead of `didConsumeAmount` perhaps we could name something like `isAmountRateLimited`

src/EndpointManager.sol
```diff
@@ -241,34 +313,62 @@ abstract contract EndpointManager is
...
-
-            // transfer tokens
-            IERC20(_token).safeTransferFrom(msg.sender, address(this), amount);
-
-            // query own token balance after transfer
-            uint256 balanceAfter = getTokenBalanceOf(_token, address(this));
-
-            // correct amount for potential transfer fees
-            amount = balanceAfter - balanceBefore;
-        } else {
```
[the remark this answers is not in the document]
nik-suri: Good callout -- I'll add a `revert` for the else case if the mode is neither locking nor burning.

src/EndpointManager.sol
djb15: > But it's an interesting callout. Should we update only  _getOutboundLimitParamsStorage().lastTxTimestamp to be block.timestamp when we queue up a transfer? (That would effectively reset the rate-limit recharge)

That would basically make it a stepwise linear recharge right? Where the gradient decreases each time you queue up a transfer? Imo I don't think we should do that unless there's a strong argument for it

src/EndpointManager.sol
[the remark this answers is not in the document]
nik-suri: Adding context here -- this is best for compatibility since on most other chains the native word size is `u64`.

src/EndpointManager.sol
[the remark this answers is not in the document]
RahulMaganti47: got it - yea i think `isAmountRateLimited` or `isAtCapacity` would be good. Also, an early return may be good if `shouldQueue` is `false` given the queuing control flow shouldn't happen in that case

src/EndpointManager.sol
RahulMaganti47, line 355: can you move this out to a separate function please?

RahulMaganti47: seems like it's stepwise constant for each set of transfers added to the queue until the capacity is reached. agreed that this is probably not a good idea

djb15 (approved):

RahulMaganti47 (approved):

PR #20, opened 2024-01-19
EndpointManager: add inbound rate limiting logic with queue

nik-suri:
https://github.com/wormhole-foundation/example-native-token-transfers/issues/14

### DIFF
--- a/src/EndpointManager.sol
+++ b/src/EndpointManager.sol
@@ -13,6 +13,7 @@ import "./libraries/external/ReentrancyGuardUpgradeable.sol";
 import "./libraries/EndpointStructs.sol";
 import "./libraries/EndpointHelpers.sol";
 import "./interfaces/IEndpointManager.sol";
+import "./interfaces/IEndpointManagerEvents.sol";
 import "./interfaces/IEndpointToken.sol";
 import "./Endpoint.sol";
 import "./EndpointRegistry.sol";
@@ -20,6 +21,7 @@ import "./EndpointRegistry.sol";
 // TODO: rename this (it's really the business logic)
 abstract contract EndpointManager is
     IEndpointManager,
+    IEndpointManagerEvents,
     EndpointRegistry,
     OwnableUpgradeable,
     ReentrancyGuardUpgradeable
@@ -564,6 +566,8 @@ abstract contract EndpointManager is
                 recipient: transferRecipient,
                 txTimestamp: block.timestamp
             });
+
+            emit InboundTransferQueued(queueSequence);
 
             // end execution early
             return;
--- a/src/EndpointManager.sol
+++ b/src/EndpointManager.sol
@@ -573,7 +573,7 @@ abstract contract EndpointManager is
                 txTimestamp: block.timestamp
             });
 
-            emit InboundTransferQueued(queueSequence);
+            emit InboundTransferQueued(queueSequence, message.chainId);
 
             // end execution early
             return;
--- a/src/EndpointManager.sol
+++ b/src/EndpointManager.sol
@@ -247,36 +247,6 @@ abstract contract EndpointManager is
         return true;
     }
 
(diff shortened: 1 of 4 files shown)


src/EndpointManager.sol
```diff
@@ -458,39 +534,89 @@ abstract contract EndpointManager is
...
+            });
+
+            // end execution early
+            return;
+        }
+
+        _mintOrUnlockToRecipient(transferRecipient, nativeTransferAmount);
+    }
+
+    function completeInboundQueuedTransfer(uint64 queueSequence) external payable nonReentrant {
```
djb15, line 556: This doesn't have to be payable

src/interfaces/IEndpointManager.sol
```diff
@@ -28,6 +28,12 @@ interface IEndpointManager {
         uint256 txTimestamp;
     }
 
+    struct InboundQueuedTransfer {
```
djb15, line 31: Order struct based on size of element types

src/interfaces/IEndpointManager.sol
```diff
@@ -40,6 +46,8 @@ interface IEndpointManager {
         payable
         returns (uint64 msgSequence);
 
+    function completeInboundQueuedTransfer(uint64 queueSequence) external payable;
```
djb15, line 49: Same here re payable

djb15 (commented):
Are there any tests for the inbound transfers being rate limited/queued yet?

nik-suri:
@djb15 added a unit test for the inbound rate limit queue

djb15 (dismissed):

src/EndpointManager.sol
```diff
@@ -476,39 +561,94 @@ abstract contract EndpointManager is
...
+            // queue up the transfer
+            uint64 queueSequence = _useInboundQueueSequence();
+
+            _getInboundQueueStorage()[queueSequence] = InboundQueuedTransfer({
+                amount: nativeTransferAmount,
+                recipient: transferRecipient,
+                txTimestamp: block.timestamp
+            });
+
+            emit InboundTransferQueued(queueSequence);
```
djb15, line 576: I'm just wondering whether from a monitoring point of view we want to include the sending chain id (i.e. the chain id of the origin chain) in this event?

djb15 (dismissed):

src/EndpointManager.sol
```diff
@@ -476,39 +561,94 @@ abstract contract EndpointManager is
...
+        bool isRateLimited = _isInboundAmountRateLimited(nativeTransferAmount, message.chainId);
+        if (isRateLimited) {
+            // queue up the transfer
+            uint64 queueSequence = _useInboundQueueSequence();
+
+            _getInboundQueueStorage()[queueSequence] = InboundQueuedTransfer({
+                amount: nativeTransferAmount,
+                recipient: transferRecipient,
+                txTimestamp: block.timestamp
+            });
```
RahulMaganti47, line 574: it may be good add a check here that the same message has not been added to the queue more than once. this function does have replay protection above, but this is a good defense in depth measure regardless

nik-suri: This means we'd need to have additional replay protection for adding messages to queues, right?
Or perhaps we could simply replay-protect by ensuring that a number less than the `queueSequence` isn't being added to the queue? Since `queueSequence` monotonically increases

djb15: Personally feel like the check is unnecessary because we're incrementing the `queueSequence` in advance right?

nik-suri: That's true -- the check would be redundant.

I also made the outbound enqueue logic safer by bundling the `queueSequence` increment inside of the function -- so that's equivalent to making the check.

RahulMaganti47: ok that sounds good to me

RahulMaganti47 (approved):

PR #26, opened 2024-01-25
Add msg.sender to wire format

nik-suri:
https://github.com/wormhole-foundation/example-native-token-transfers/issues/12

### DIFF
--- a/foundry.toml
+++ b/foundry.toml
@@ -2,7 +2,7 @@
 solc_version = "0.8.19"
 optimizer = true
 optimizer_runs = 200
-via_ir = false
+via_ir = true
 evm_version = "london"
 src = "src"
 out = "out"
--- a/src/Manager.sol
+++ b/src/Manager.sol
@@ -502,8 +502,11 @@ abstract contract Manager is
 
         // construct the ManagerMessage payload
         uint64 sequence = _useMessageSequence();
+        bytes memory senderBytes = abi.encodePacked(msg.sender);
         bytes memory encodedManagerPayload = EndpointStructs.encodeManagerMessage(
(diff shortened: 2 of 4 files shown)


foundry.toml
```diff
@@ -2,7 +2,7 @@
 solc_version = "0.8.19"
 optimizer = true
 optimizer_runs = 200
-via_ir = false
+via_ir = true
```
kcsongor, line 5: why do we need this? makes compilation a *lot* slower :(

kcsongor (approved):

foundry.toml
nik-suri: Yeah, makes compilation slower but won't compile without it:
```
Error:
Compiler run failed:
Error: Compiler error (/solidity/libsolidity/codegen/LValue.cpp:52):Stack too deep. Try compiling with `--via-ir` (cli) or the equivalent `viaIR: true` (standard JSON) while enabling the optimizer. Otherwise, try removing local variables.
   --> src/Manager.sol:513:21:
    |
513 |         sendMessage(recipientChain, encodedManagerPayload);
    |                     ^^^^^^^^^^^^^^

make: *** [Makefile:8: build] Error 1
```

Not sure if this is a machine-specific error or happens across the board.

djb15: Can you refactor the relevant logic to block scope some variables instead?

djb15 (approved):

RahulMaganti47 (approved):

foundry.toml
nik-suri: Can address this in a follow up PR -- https://github.com/wormhole-foundation/example-native-token-transfers/issues/29

PR #27, opened 2024-01-25
Interface cleanup

nik-suri:
- Remove receiver interfaces from Endpoint
- Remove sibling interfaces -- these are Endpoint specific

### DIFF
--- a/src/WormholeEndpoint.sol
+++ b/src/WormholeEndpoint.sol
@@ -29,6 +29,7 @@ abstract contract WormholeEndpoint is Endpoint {
     error InvalidSibling(uint16 chainId, bytes32 siblingAddress);
     error TransferAlreadyCompleted(bytes32 vaaHash);
     error InvalidSiblingZeroAddress();
+    error InvalidSiblingChainIdZero();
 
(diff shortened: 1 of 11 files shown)


kcsongor (dismissed):

src/WormholeEndpoint.sol
```diff
@@ -110,4 +122,18 @@ abstract contract WormholeEndpoint is Endpoint {
...
     }
+
+    /// @notice Get the corresponding Endpoint contract on other chains that have been registered via governance.
+    ///         This design should be extendable to other chains, so each Endpoint would be potentially concerned with Endpoints on multiple other chains
+    ///         Note that siblings are registered under wormhole chainID values
+    function getSibling(uint16 chainId) public view returns (bytes32) {
+        return _siblings[chainId];
+    }
+
+    function _setSibling(uint16 chainId, bytes32 siblingContract) internal {
```
djb15, line 133: We should also check that `chainId != 0` now that this is WH specific

djb15 (approved):

RahulMaganti47 (approved):

PR #28, opened 2024-01-26
Rate Limiting Refactor

nik-suri:
Changes:
- revert if dust amount included
- refactor to make outbound rate limiting clearer

https://github.com/wormhole-foundation/example-native-token-transfers/issues/6

### DIFF
--- a/src/Manager.sol
+++ b/src/Manager.sol
@@ -447,6 +447,9 @@ abstract contract Manager is
 
         // now check rate limits
         bool isAmountRateLimited = _isOutboundAmountRateLimited(amount);
+        if (!shouldQueue && isAmountRateLimited) {
+            revert NotEnoughCapacity(getCurrentOutboundCapacity(), amount);
+        }
         if (shouldQueue && isAmountRateLimited) {
             // queue up and return
             uint64 queueSequence = _enqueueOutboundTransfer(amount, recipientChain, recipient);
--- a/src/Manager.sol
+++ b/src/Manager.sol
@@ -314,9 +314,6 @@ abstract contract Manager is
         uint256 capacity,
         RateLimitParams storage rateLimitParams
     ) internal {
-        if (capacity < amount) {
-            revert NotEnoughCapacity(capacity, amount);
-        }
         rateLimitParams.lastTxTimestamp = block.timestamp;
         rateLimitParams.currentCapacity = capacity - amount;
     }
--- a/src/Manager.sol
+++ b/src/Manager.sol
@@ -174,7 +174,7 @@ abstract contract Manager is
     function quoteDeliveryPrice(uint16 recipientChain) public view virtual returns (uint256);
 
     /// @dev This will either cross-call or internal call, depending on whether the contract is standalone or not.
-    function sendMessage(uint16 recipientChain, bytes memory payload) internal virtual;
+    function _sendMessageToEndpoint(uint16 recipientChain, bytes memory payload) internal virtual;
 
     // TODO: do we want additional information (like chain etc)
     function isMessageApproved(bytes32 digest) public view virtual returns (bool);
@@ -315,7 +315,7 @@ abstract contract Manager is
         RateLimitParams storage rateLimitParams
     ) internal {
         if (capacity < amount) {
-            revert NotEnoughOutboundCapacity(capacity, amount);
+            revert NotEnoughCapacity(capacity, amount);
         }
         rateLimitParams.lastTxTimestamp = block.timestamp;
         rateLimitParams.currentCapacity = capacity - amount;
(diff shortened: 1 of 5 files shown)


src/Manager.sol
```diff
@@ -405,79 +401,78 @@ abstract contract Manager is
...
-            }
+            // correct amount for potential burn fees
+            amount = balanceAfter - balanceBefore;
+        } else {
+            revert InvalidMode(uint8(_mode));
+        }
+
+        // now check rate limits
+        bool isAmountRateLimited = _isOutboundAmountRateLimited(amount);
+        if (shouldQueue && isAmountRateLimited) {
```
djb15, line 450: Looks like we've lost the revert if `shouldQueue = false`?

nik-suri: The revert is handled in `_consumeOutboundAmount`

djb15: Ok nice I didn't see it in this diff and was worried we'd lost that path

djb15 (approved):

src/Manager.sol
```diff
@@ -405,79 +401,78 @@ abstract contract Manager is
...
-                payable(msg.sender).transfer(msg.value);
+            // return the sequence in the queue
+            return queueSequence;
+        }
 
-                // return the sequence in the queue
-                return queueSequence;
-            }
+        // otherwise, consume the outbound amount
+        _consumeOutboundAmount(amount);
```
RahulMaganti47, line 462: this may be stylistic but since the `false` code path for `shouldQueue` reverts, imo makes more sense to check that condition first and then proceed to the enqueuing logic

RahulMaganti47 (changes requested):

src/Manager.sol
nik-suri: This is technically checked up front before the entire if-statement with
```
bool isAmountRateLimited = _isOutboundAmountRateLimited(amount);
```

We then only proceed to the enqueuing logic if that check passes. I would say this is more stylistic but I'm curious to get @djb15 's thoughts on whether one structuring is more readable than the other.

RahulMaganti47: yea logically i think its fine. more of a question of how structure early returns in the control flow

djb15: As I was reviewing this PR I actually got confused by this and thought there was a bug, so I agree with Rahul here. Can we have an explicit `if (!shouldQueue && isAmountRateLimited)` before the queuing logic or something similar?

src/Manager.sol
```diff
@@ -486,6 +481,10 @@ abstract contract Manager is
             payable(msg.sender).transfer(excessValue);
         }
 
+        // query tokens decimals
+        (, bytes memory queriedDecimals) = _token.staticcall(abi.encodeWithSignature("decimals()"));
+        uint8 decimals = abi.decode(queriedDecimals, (uint8));
+
         // normalize amount decimals
         uint256 normalizedAmount = normalizeAmount(amount, decimals);
```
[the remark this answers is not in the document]
nik-suri: In `transfer` we do this:
```
uint256 newAmount = deNormalizeAmount(normalizeAmount(amount, decimals), decimals);
```

This is to cut off any dust and we revert if the transfer amount includes any dust amount of tokens. But we denormalize back to the correct number of decimals so that we can call the `safeTransferFrom` and `burnFrom` methods properly.

Then in `_transfer` we perform the actual normalization to 8 decimals for the cross-chain transfer.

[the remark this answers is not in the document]
djb15: Ah yeah makes sense, thanks for clarifying!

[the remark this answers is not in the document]
djb15: I was just thinking that maybe to save gas we could do this normalisation in `transfer` instead (and store the normalised amount in the queue as well) to avoid having to fetch the decimals again?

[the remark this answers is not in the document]
nik-suri: We can definitely do that to save gas. I had originally kept the normalization logic in `_transfer` to be sure that it happens every time and isolate the normalization assumption. If you think the trade-off to save gas is worth it I can certainly refactor.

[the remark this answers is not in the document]
djb15: I've not checked all the implications of this so maybe there's another blocker but generally I'd vote for saving gas here. We'd have to make sure the getter of a queued transfer denormalises to avoid confusion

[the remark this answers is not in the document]
nik-suri: Synced offline with @djb15 on this -- we verified the two cases:
1. Transfer isn't queued -- it would obviously be cheaper not to double call, but because the slot is hot the extra gas is pretty minimal.
2. Transfer is queued -- It's cheaper to call again rather than use extra storage for the token decimals

So we will keep this code as-is.

RahulMaganti47 (approved):

PR #31, opened 2024-01-29
Add sibling registration and validation to manager contract

nik-suri:
https://github.com/wormhole-foundation/example-native-token-transfers/issues/22

### DIFF
--- a/src/Manager.sol
+++ b/src/Manager.sol
@@ -54,6 +54,8 @@ abstract contract Manager is
     bytes32 public constant MESSAGE_SEQUENCE_SLOT =
         bytes32(uint256(keccak256("ntt.messageSequence")) - 1);
 
+    bytes32 public constant SIBLINGS_SLOT = bytes32(uint256(keccak256("ntt.siblings")) - 1);
+
     bytes32 public constant OUTBOUND_LIMIT_PARAMS_SLOT =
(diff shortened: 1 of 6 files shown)


src/Manager.sol
```diff
@@ -681,4 +696,44 @@ abstract contract Manager is
...
+            return false;
+        }
+
+        for (uint256 i = 0; i < a.length; i++) {
+            if (a[i] != b[i]) {
+                return false;
+            }
+        }
+
+        return true;
```
RahulMaganti47, line 737: how expensive is this? is there a way to reduce the gas cost by batch loading word by word instead of byte by bytes?

RahulMaganti47 (dismissed):

src/Manager.sol
nik-suri: Synced offline -- we'll open a follow up PR to use the [BytesLib equal method](https://github.com/GNSPS/solidity-bytes-utils/blob/master/contracts/BytesLib.sol#L396-L503) to compare words, with the assumption that memory is word aligned (which it should be since we don't have any weird assembly blocks).

Opened an issue for this -- https://github.com/wormhole-foundation/example-native-token-transfers/issues/36

RahulMaganti47 (approved):

djb15 (approved):

PR #32, opened 2024-01-29
Endpoint upgradeability via manager

nik-suri:
https://github.com/wormhole-foundation/example-native-token-transfers/issues/30

### DIFF
--- a/src/EndpointStandalone.sol
+++ b/src/EndpointStandalone.sol
@@ -5,14 +5,12 @@ import "./Endpoint.sol";
 import "./interfaces/IManagerStandalone.sol";
 import "./interfaces/IEndpointStandalone.sol";
 import "./libraries/Implementation.sol";
-import "./libraries/ImmutableMigrator.sol";
 import "./libraries/external/ReentrancyGuardUpgradeable.sol";
 
 abstract contract EndpointStandalone is
     IEndpointStandalone,
     Endpoint,
     Implementation,
-    ImmutableMigrator,
     ReentrancyGuardUpgradeable
 {
     /// @dev updating bridgeManager requires a new Endpoint deployment.
--- a/src/EndpointStandalone.sol
+++ b/src/EndpointStandalone.sol
@@ -37,12 +37,12 @@ abstract contract EndpointStandalone is
 
     function _migrate() internal override {}
 
+    function _checkImmutables() internal view override {
+        assert(this.manager() == manager);
+    }
+
     function upgrade(address newImplementation) external onlyManager {
         _upgrade(newImplementation);
-        if (!this.getMigratesImmutables()) {
-            assert(this.manager() == manager);
-        }
-        _setMigratesImmutables(false);
     }
 
     /// @notice Called by the BridgeManager contract to send a cross-chain message.
--- a/src/EndpointStandalone.sol
+++ b/src/EndpointStandalone.sol
@@ -37,6 +37,7 @@ abstract contract EndpointStandalone is
 
     function _migrate() internal override {}
 
+    /// @dev When we add new immutables, this function should be updated
     function _checkImmutables() internal view override {
         assert(this.manager() == manager);
     }
--- a/src/EndpointStandalone.sol
+++ b/src/EndpointStandalone.sol
@@ -4,28 +4,54 @@ pragma solidity >=0.6.12 <0.9.0;
 import "./Endpoint.sol";
 import "./interfaces/IManagerStandalone.sol";
 import "./interfaces/IEndpointStandalone.sol";
+import "./libraries/Implementation.sol";
+import "./libraries/external/ReentrancyGuardUpgradeable.sol";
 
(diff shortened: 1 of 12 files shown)


src/EndpointStandalone.sol
```diff
@@ -4,28 +4,54 @@ pragma solidity >=0.6.12 <0.9.0;
...
-abstract contract EndpointStandalone is IEndpointStandalone, Endpoint {
-    /// updating bridgeManager requires a new Endpoint deployment.
+abstract contract EndpointStandalone is
+    IEndpointStandalone,
+    Endpoint,
+    Implementation,
+    ReentrancyGuardUpgradeable
+{
+    bytes32 public constant ENDPOINT_MANAGER_SLOT =
+        bytes32(uint256(keccak256("endpoint.manager")) - 1);
```
djb15, line 17: Can you make this `ntt.endpointManager` instead please to stick with the ntt theme

src/EndpointStandalone.sol
```diff
@@ -4,28 +4,54 @@ pragma solidity >=0.6.12 <0.9.0;
...
     /// Projects should implement their own governance to remove the old Endpoint contract address and then add the new one.
-    address immutable _manager;
+    function _getEndpointManagerStorage() internal pure returns (_Address storage $) {
+        uint256 slot = uint256(ENDPOINT_MANAGER_SLOT);
+        assembly ("memory-safe") {
+            $.slot := slot
+        }
+    }
 
     constructor(address manager) {
```
djb15, line 28: This won't do anything behind a proxy. We probably want to set the manager in the initialiser and remove the constructor

src/WormholeEndpoint.sol
```diff
@@ -31,10 +18,85 @@ abstract contract WormholeEndpoint is Endpoint {
     error InvalidWormholeSiblingZeroAddress();
     error InvalidWormholeSiblingChainIdZero();
 
+    struct _UInt256 {
```
djb15, line 21: Nit pick but I prefer `_Uint256`

src/WormholeEndpoint.sol
```diff
@@ -8,19 +8,6 @@ import "./interfaces/IWormhole.sol";
 import "./Endpoint.sol";
 
 abstract contract WormholeEndpoint is Endpoint {
-    // TODO -- fix this after some testing
-    uint256 constant _GAS_LIMIT = 500000;
-
-    address immutable _wormholeCoreBridge;
-    address immutable _wormholeRelayerAddr;
-    uint256 immutable _wormholeEndpoint_evmChainId;
```
djb15, line 16: Why have you modified these constants/immutables to now be in storage? I don't see why we'd need to do this + there are no setters to update these anyway

src/Manager.sol
```diff
@@ -154,10 +154,10 @@ abstract contract Manager is
 
     // TODO: should we store this in storage? otherwise they can change during
     // each upgrade
-    address immutable _token;
-    Mode immutable _mode;
-    uint16 immutable _chainId;
-    uint256 immutable _evmChainId;
+    address public immutable _token;
```
kcsongor, line 157: let's add a comment here to remind ourselves in the future that when adding new immutables, the `upgrade` function should be updated

src/WormholeEndpoint.sol
```diff
@@ -31,6 +18,43 @@ abstract contract WormholeEndpoint is Endpoint {
...
+        uint256 slot = uint256(WORMHOLE_SIBLINGS_SLOT);
+        assembly ("memory-safe") {
+            $.slot := slot
+        }
+    }
+
+    // TODO -- fix this after some testing
+    uint256 constant _GAS_LIMIT = 500000;
+
+    address immutable _wormholeCoreBridge;
```
kcsongor, line 54: should we keep immutables at the top of the contract?

src/libraries/ImmutableMigrator.sol
```diff
@@ -0,0 +1,4 @@
+// SPDX-License-Identifier: Apache 2
+pragma solidity >=0.6.12 <0.9.0;
+
+abstract contract ImmutableMigrator {}
```
kcsongor, line 4: we can delete this now

src/WormholeEndpoint.sol
nik-suri: We can, I defaulted to keeping them above the constructor where they were initialized but can move all of them to the top of the contract.

Might actually make most sense to keep them right above the `_checkImmutables()` function.

src/Manager.sol
djb15, line 157: Nit pick but for public variables can we get rid of the leading underscores?

src/WormholeEndpoint.sol
nik-suri: I actually moved them all to the top of the contract for consistency, since not all of the contracts with immutables have the `_checkImmutables()` function (e.g. `Manager.sol`)

kcsongor (approved):

djb15 (approved):

PR #37, opened 2024-01-30
swap `sender` and `msgType` in `ManagerMessage`

kcsongor:
It makes more sense for the discriminator to directly precede the payload


nik-suri:
This is no longer relevant after https://github.com/wormhole-foundation/example-native-token-transfers/pull/33

PR #40, opened 2024-02-01
Add Prefix and source token address to NativeTokenTransfer struct

nik-suri:
Closes https://github.com/wormhole-foundation/example-native-token-transfers/issues/35
Closes https://github.com/wormhole-foundation/example-native-token-transfers/issues/39

### DIFF
--- a/src/libraries/EndpointStructs.sol
+++ b/src/libraries/EndpointStructs.sol
@@ -91,12 +91,16 @@ library EndpointStructs {
     /// Token Transfer payload corresponding to type == 1
     /// @dev The wire format is as follows:
     ///    - amount - 32 bytes
+    ///    - sourceTokenLength - 2 bytes
+    ///    - sourceToken - `sourceTokenLength` bytes
     ///    - toLength - 2 bytes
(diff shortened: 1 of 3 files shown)


src/libraries/EndpointStructs.sol
```diff
@@ -7,6 +7,9 @@ library EndpointStructs {
     using BytesParsing for bytes;
 
     error PayloadTooLong(uint256 size);
+    error IncorrectPrefix(bytes4 prefix);
+
+    bytes4 constant NTT_PREFIX = 0x994E5454;
```
djb15, line 12: Can you add a comment to explain the prefix. I assume it's the first 4 bytes of the hash of some relevant string?

djb15 (dismissed):

src/libraries/EndpointStructs.sol
nik-suri: Ah my comment got accidentally removed. It's meant to be `0x99'N''T''T'`

RahulMaganti47 (approved):

PR #41, opened 2024-02-01
Current Capacity Logic refactor

nik-suri:


djb15 (dismissed):

src/Manager.sol
```diff
@@ -267,18 +267,12 @@ abstract contract Manager is
...
-            capacity = calculatedCapacity > rateLimitParams.limit
-                ? rateLimitParams.limit
-                : calculatedCapacity;
-        }
+        uint256 timePassed = block.timestamp - rateLimitParams.lastTxTimestamp;
+        uint256 calculatedCapacity =
+            rateLimitParams.currentCapacity + (timePassed * rateLimitParams.ratePerSecond);
+
+        return
+            calculatedCapacity > rateLimitParams.limit ? rateLimitParams.limit : calculatedCapacity;
```
RahulMaganti47, line 275: might be good to abstract this out into a `min` or `max` function to make it more consistent with the other implementations. also makes the clamping explicit.

RahulMaganti47: also imo a bit more readable if its `calculatedCapacity < rateLimitParams.limit ? calculatedCapacity: rateLimitParams.limit`

RahulMaganti47 (approved):

djb15 (approved):

PR #43, opened 2024-02-01
Refactor Inbound Queue Keying

nik-suri:
Closes https://github.com/wormhole-foundation/example-native-token-transfers/issues/42

Refactor inbound queue to key by source chain ID and manager message sequence.

### DIFF
--- a/src/Manager.sol
+++ b/src/Manager.sol
@@ -144,7 +144,7 @@ abstract contract Manager is
     function _getInboundQueueStorage()
         internal
         pure
-        returns (mapping(uint16 => mapping(uint64 => InboundQueuedTransfer)) storage $)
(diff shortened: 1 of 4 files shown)


src/Manager.sol
```diff
@@ -147,21 +144,14 @@ abstract contract Manager is
     function _getInboundQueueStorage()
         internal
         pure
-        returns (mapping(uint64 => InboundQueuedTransfer) storage $)
+        returns (mapping(uint16 => mapping(uint64 => InboundQueuedTransfer)) storage $)
```
RahulMaganti47, line 147: how about keying by the message hash instead like @kcsongor suggested?

RahulMaganti47 (approved):

PR #45, opened 2024-02-02
Fix Wormhole Standard Relaying Integration

nik-suri:
Fix Wormhole standard relaying integration to properly receive messages and flexibly adapt to new evm chains.

This PR:
1. Uses Wormhole standard relaying when the target chain is EVM and supported by standard relaying.
2. Directly emits Wormhole messages via the core bridge when the conditions for (1) are not met.

A couple important things to note about the integration:
1. This change does not implement the case to handle a caller requesting for a transfer to not be emitted via Standard Relaying even if the target chain is EVM + relaying-supported. I'm playing around with local changes on how exactly to implement this, it requires some new structs and interface changes.
2. I've disallowed relayer deliveries with additionalMessages populated, since this contract never expects to receive a delivery from the relayer with additional messages to verify and deliver.

### DIFF
--- a/src/WormholeEndpoint.sol
+++ b/src/WormholeEndpoint.sol
@@ -98,19 +98,19 @@ abstract contract WormholeEndpoint is Endpoint, IWormholeEndpoint, IWormholeRece
         wormholeEndpoint_evmChainId = block.chainid;
     }
 
+    function shouldRelayViaStandardRelaying(uint16 chainId) public view returns (bool) {
+        return isWormholeRelayingEnabled(chainId) && isWormholeEvmChain(chainId);
+    }
+
     function _quoteDeliveryPrice(uint16 targetChain)
         internal
         view
         override
         returns (uint256 nativePriceQuote)
     {
-        if (isWormholeRelayingEnabled(targetChain)) {
-            if (isWormholeEvmChain(targetChain)) {
-                (uint256 cost,) = wormholeRelayer.quoteEVMDeliveryPrice(targetChain, 0, GAS_LIMIT);
-                return cost;
-            } else {
-                revert RelayingNotImplemented(targetChain);
-            }
+        if (shouldRelayViaStandardRelaying(targetChain)) {
+            (uint256 cost,) = wormholeRelayer.quoteEVMDeliveryPrice(targetChain, 0, GAS_LIMIT);
+            return cost;
         } else {
             return 0;
         }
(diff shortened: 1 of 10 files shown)


src/WormholeEndpoint.sol
```diff
@@ -105,23 +130,26 @@ abstract contract WormholeEndpoint is Endpoint {
...
+            }
         } else {
-            wormholeRelayer().sendPayloadToEvm{value: msg.value}(
-                recipientChain,
-                fromWormholeFormat(getWormholeSibling(recipientChain)),
-                encodedEndpointPayload,
-                0,
-                _GAS_LIMIT
-            );
+            wormhole.publishMessage(0, encodedEndpointPayload, 1);
```
RahulMaganti47, line 152: this logic is flipped? this should revert relaying is not enabled and emit via the core bridge when its not an evm chain.

src/WormholeEndpoint.sol
```diff
@@ -238,4 +289,26 @@ abstract contract WormholeEndpoint is Endpoint {
...
+
+    function isWormholeRelayingEnabled(uint16 chainId) public view returns (bool) {
+        return _getWormholeRelayingEnabledChainsStorage()[chainId];
+    }
+
+    function _setIsWormholeRelayingEnabled(uint16 chainId, bool isEnabled) internal {
+        if (chainId == 0) {
+            revert InvalidWormholeChainIdZero();
+        }
+        _getWormholeRelayingEnabledChainsStorage()[chainId] = isEnabled;
```
RahulMaganti47, line 301: may be good to add an event here

src/WormholeEndpoint.sol
```diff
@@ -238,4 +289,26 @@ abstract contract WormholeEndpoint is Endpoint {
...
+
+    function isWormholeEvmChain(uint16 chainId) public view returns (bool) {
+        return _getWormholeEvmChainIdsStorage()[chainId];
+    }
+
+    function _setIsWormholeEvmChain(uint16 chainId) internal {
+        if (chainId == 0) {
+            revert InvalidWormholeChainIdZero();
+        }
+        _getWormholeEvmChainIdsStorage()[chainId] = true;
```
RahulMaganti47, line 312: add an event here as well

src/WormholeEndpoint.sol
RahulMaganti47: maybe better to define a predicate for when to emit via core bridge and then another predicate for emitting via AR. `relayViaAR` and `emitViaCoreBridge`

RahulMaganti47 (approved):

PR #46, opened 2024-02-02
fix: wrap Manager msg in Endpoint msg

### LINKED ISSUE #23: Use `EndpointMessage` struct
Currently `EndpointStructs.EndpointMessage` is unused.

The `Endpoint` contract should emit this struct instead of simply re-emitting the payload sent to it from the manager contract.

RahulMaganti47:
Fixes #23

### DIFF
--- a/src/libraries/EndpointStructs.sol
+++ b/src/libraries/EndpointStructs.sol
@@ -160,12 +160,29 @@ library EndpointStructs {
     }
 
     struct EndpointMessage {
-        /// @notice Magic string (constant value set by messaging provider) that idenfies the payload as an endpoint-emitted payload.
-        ///         Note that this is not a security critical field. It's meant to be used by messaging providers to identify which messages are Endpoint-related.
-        bytes32 endpointId;
+        /// @notice
+        bytes4 prefix;
         /// @notice Payload provided to the Endpoint contract by the Manager contract.
         bytes managerPayload;
-        /// @notice Custom payload which messaging providers can use to pass bridge-specific information, if needed.
-        bytes endpointPayload;
+    }
+
+    /*
+     * @dev Encodes an Endpoint message for communication between the Manager and the Endpoint.
+     *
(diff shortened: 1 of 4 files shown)


src/libraries/EndpointStructs.sol
```diff
@@ -160,12 +160,29 @@ library EndpointStructs {
...
+     * @throws PayloadTooLong if the length of endpointId, managerPayload, or endpointPayload exceeds the allowed maximum.
+     */
+    function encodeEndpointMessage(EndpointMessage memory m)
+        public
+        pure
+        returns (bytes memory encoded)
+    {
+        // if (m.managerPayload.length > type(uint16).max) {
+        //     revert PayloadTooLong(m.managerPayload.length);
+        // }
```
nik-suri, line 183: Let's leave this comment in for now for security review to consider.
Not sure whether we should perform checks on `managerPayload.length`

test/Manager.t.sol
```diff
@@ -342,7 +356,7 @@ contract TestManager is Test, IManagerEvents {
             0,
             abi.encodePacked(address(manager)),
             abi.encodePacked(address(0)),
-            abi.encode(EndpointStructs.EndpointMessage("hello", "world", "payload"))
+            abi.encode(EndpointStructs.EndpointMessage(0x9945FF10, "payload"))
```
nik-suri, line 359: can you please update the tests to wrap the `ManagerMessage` in the `EndpointMessage` and not the other way around?
You'll need to update the `DummyEndpoint.receiveMessage` logic.

RahulMaganti47: will update this in a separate PR along with other test refactors. currently tests are not setup to test the wrapping logic because it skips the `deliverToManager` path

nik-suri: Sounds good and thanks for creating https://github.com/wormhole-foundation/example-native-token-transfers/issues/48

nik-suri (approved):

PR #47, opened 2024-02-05
emit events to notify clients

### LINKED ISSUE #18: Add Events
Descriptive events that get logged when various actions happen are useful for indexers and generally analyzing data on-chain.

We should add event logging throughout the codebase. For example, when:
- State gets updated (siblings, threshold, Endpoint management, etc)
- Transfers are rate-limited or queued
- Transfers are sent
- Transfers are redeemed

RahulMaganti47:
Fixes #18

### DIFF
--- a/src/Manager.sol
+++ b/src/Manager.sol
@@ -655,7 +655,10 @@ abstract contract Manager is
         if (_isAllZeros(siblingContract)) {
             revert InvalidSiblingZeroBytes();
         }
+
         _getSiblingsStorage()[chainId_] = siblingContract;
+
+        emit SiblingUpdated(chainId_, siblingContract);
     }
 
     function _isAllZeros(bytes memory payload) internal pure returns (bool) {
--- a/src/Manager.sol
+++ b/src/Manager.sol
@@ -494,6 +494,8 @@ abstract contract Manager is
         // send the message
         _sendMessageToEndpoint(recipientChain, encodedManagerPayload);
 
+        emit TransferSent(recipient, recipientChain, sequence);
+
         // return the sequence number
         return sequence;
     }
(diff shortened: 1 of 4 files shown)


src/Manager.sol
```diff
@@ -497,6 +505,8 @@ abstract contract Manager is
         // send the message
         _sendMessageToEndpoint(recipientChain, encodedManagerPayload);
 
+        emit TransferSent(recipient, recipientChain, sequence);
```
nik-suri, line 508: Let's add the amount to this event? 
Both normalized and non-normalized

src/Manager.sol
```diff
@@ -673,7 +683,10 @@ abstract contract Manager is
         if (_isAllZeros(siblingContract)) {
             revert InvalidSiblingZeroBytes();
         }
+
         _getSiblingsStorage()[chainId_] = siblingContract;
+
+        emit SiblingUpdated(chainId_, siblingContract);
```
nik-suri, line 689: Should we include the old sibling contract address there too?
If unset, it'd be zero -- just whatever was in the mapping before we set the new value.

src/ManagerStandalone.sol
```diff
@@ -67,6 +67,8 @@ contract ManagerStandalone is IManagerStandalone, Manager, Implementation {
         _Threshold storage _threshold = _getThresholdStorage();
         _threshold.num = threshold;
         _checkThresholdInvariants();
+
+        emit ThresholdChanged(threshold);
```
nik-suri, line 71: Same here -- let's include the old threshold value as well?

src/WormholeEndpoint.sol
```diff
@@ -163,5 +165,7 @@ abstract contract WormholeEndpoint is Endpoint {
             revert InvalidWormholeSiblingZeroAddress();
         }
         _getWormholeSiblingsStorage()[chainId] = siblingContract;
+
+        emit SetWormholeSibling(chainId, siblingContract);
```
nik-suri, line 169: Same here -- what do you think about including the old value?

nik-suri (approved):
lgtm, couple of minor suggestions

nik-suri (approved):

PR #52, opened 2024-02-05
Narrow compiler range to 0.8.x

nik-suri:
Closes https://github.com/wormhole-foundation/example-native-token-transfers/issues/38


kcsongor:
Yeah, upon further thinking, there are a bunch of compiler features that are not controlled by the hardfork version. For example, 0.8.0 introduced overflow checking by default on all arithmetic operations, so it's definitely a good idea to require at least that version.

RahulMaganti47 (approved):

djb15:
@nik-suri To support custom errors we need at least 0.8.4 fyi

PR #61, opened 2024-02-06
NormalizedAmount value type

kcsongor:
The commits are best reviewed separately (and should be rebased instead of squashed)

### DIFF
--- a/src/libraries/NormalizedAmount.sol
+++ b/src/libraries/NormalizedAmount.sol
@@ -3,6 +3,40 @@ pragma solidity >=0.8.0 <0.9.0;
 
 type NormalizedAmount is uint64;
 
+using {gt as >, lt as <, sub as -, add as +, div, mul, min, unwrap} for NormalizedAmount global;
+
+function gt(NormalizedAmount a, NormalizedAmount b) pure returns (bool) {
+    return NormalizedAmount.unwrap(a) > NormalizedAmount.unwrap(b);
+}
+
+function lt(NormalizedAmount a, NormalizedAmount b) pure returns (bool) {
+    return NormalizedAmount.unwrap(a) < NormalizedAmount.unwrap(b);
+}
+
+function sub(NormalizedAmount a, NormalizedAmount b) pure returns (NormalizedAmount) {
+    return NormalizedAmount.wrap(NormalizedAmount.unwrap(a) - NormalizedAmount.unwrap(b));
+}
+
+function add(NormalizedAmount a, NormalizedAmount b) pure returns (NormalizedAmount) {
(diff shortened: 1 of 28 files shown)


src/libraries/NormalizedAmount.sol
```diff
@@ -3,6 +3,40 @@ pragma solidity >=0.8.0 <0.9.0;
...
+
+function mul(NormalizedAmount a, uint64 b) pure returns (NormalizedAmount) {
+    return NormalizedAmount.wrap(NormalizedAmount.unwrap(a) * b);
+}
+
+function min(NormalizedAmount a, NormalizedAmount b) pure returns (NormalizedAmount) {
+    return a > b ? b : a;
+}
+
+function unwrap(NormalizedAmount a) pure returns (uint64) {
```
djb15, line 32: Do we need this `unwrap` method? I thought it was implicit?

src/libraries/NormalizedAmount.sol
```diff
@@ -3,6 +3,40 @@ pragma solidity >=0.8.0 <0.9.0;
 
 type NormalizedAmount is uint64;
```
djb15, line 4: Btw we'll need minimum solidity version 0.8.8 for user defined types

djb15 (dismissed):
With these changes we'd need to operationally ensure the limits are correctly set to an appropriate normalised value, but otherwise it looks good

src/libraries/NormalizedAmount.sol
kcsongor: it just allows `foo.unwrap()` instead of `NormalizedAmount.unwrap(foo)` (but it's never done implicitly, that's the point!)

kcsongor:
@djb15 hmm, maybe we should change the public set functions to take unnormalised amounts and perform normalisation in the function?

src/libraries/NormalizedAmount.sol
djb15: Sorry yeah implicit was the wrong word! That's nice though, it looks a bit cleaner this way

djb15:
> @djb15 hmm, maybe we should change the public set functions to take unnormalised amounts and perform normalisation in the function?

Ah yeah I hadn't considered that as an option. I think for the integrators using this code it's much easier to reason in actual token amounts rather than in normalised amounts that they might not be as familiar with...so I'd vote for what you've suggested (can't think of any major downsides either)

kcsongor:
@djb15 changed it so all external facing stuff now shows unnormalised amounts

djb15 (approved):

RahulMaganti47 (approved):

PR #66, opened 2024-02-08
Bugfix: remove burn fee calculation

nik-suri:


djb15 (commented):
I think we're safer querying balances for this code path still. For example a token might be non-standard and return false instead of reverting if a burn action fails. Probably not necessary but good defence in depth

mdulin2 (approved):

nik-suri:
@djb15 how do you imagine implementing this defense in depth?

Querying balance before and after and verifying that the difference equals the amount that the user is attempting to transfer?

djb15:
@nik-suri I would keep everything the same (as it was before this PR) but change `amount = balanceAfter - balanceBefore` to `amount = balanceBefore - balanceAfter`

nik-suri:
@djb15 my concern with that is it could result in weird implicit errors if a token is non-standard and implements a burn fee.

For example, the case I mention in the comment -- imagine the user transfers 20 and the burn fee is 1, and the burn fee is taken out of the user's balance in addition to the burn amount. This could result in the following scenario:
```
- balanceBefore = 100
- balanceAfter = 79
- Amount corrected from 20 -> 21
```
But this would be incorrect, since amount should only be 20. And it'll result in 21 tokens getting minted on the destination chain, since there are no more checks around the transfer amount after this block of code.

I think we should implement the defense in depth by adding a check that `balanceBefore - balanceAfter == amount`.
Lmk if I'm missing something though.

djb15:
@nik-suri Gotcha yeah so rather than just documenting these fee-on-burn tokens aren't supported, we explicitly add a check there to make sure they aren't accidentally used. Sounds good to me

djb15 (approved):

PR #79, opened 2024-02-08
WormholeEndpoint: check EndpointMessage is consumed on parsing

### LINKED ISSUE #72: Check the whole bytearray is consumed when parsing EndpointMessage

RahulMaganti47:
Fixes #72


djb15 (commented):
Probably cleaner to use the `checkLength` method from the library? Example: https://github.com/wormhole-foundation/example-native-token-transfers/blob/e98c69d10d4ace01e19c717403fa8a851867d1a1/src/libraries/EndpointStructs.sol#L74

djb15 (approved):

PR #85, opened 2024-02-08
Add runtime invariants for incoming and outgoing transfers

djb15:
Add the following runtime invariants:

1. The token balance of msg.sender should always decrease when initiating a transfer
    1. The token balance should decrease by the amount specified in the cross chain message
2. In burning mode the token balance of the contract should never change
    1. The change in supply of the token should be equal to what’s specified in the message and should be no greater than the total supply of the canonical token
3. In locking mode, the token balance of the contract should always increase when initiating a transfer
    1. The token balance should increase by the same amount that has left the user and by the same amount as in the cross chain message
4. In locking mode, the token balance of the contract should always decrease when completing a transfer
    1. This should be the same amount as specified in the message
5. The ETH balance of the contract should never change during a transaction
6. The rate limit should always be between 0 and maxRateLimit inclusive


nik-suri:
Discussed offline, let's add these to the testing suite instead of to the runtime.

PR #87, opened 2024-02-09
implement pausing functionality

RahulMaganti47:
Adds #84

### DIFF
--- a/src/Endpoint.sol
+++ b/src/Endpoint.sol
@@ -2,8 +2,9 @@
 pragma solidity >=0.8.8 <0.9.0;
 
 import "./libraries/EndpointStructs.sol";
+import "./libraries/Pausable.sol";
 
-abstract contract Endpoint {
+abstract contract Endpoint is Pausable {
     function _sendMessage(uint16 recipientChain, bytes memory managerMessage) internal virtual;
 
     function _deliverToManager(EndpointStructs.ManagerMessage memory payload) internal virtual;
@@ -15,4 +16,9 @@ abstract contract Endpoint {
         pure
         virtual
         returns (EndpointStructs.EndpointMessage memory endpointMessage);
+
+    /// @notice pause the endpoint
+    function _pauseEndpoint() internal {
+        _pause();
+    }
 }
--- a/src/Endpoint.sol
+++ b/src/Endpoint.sol
@@ -16,9 +16,4 @@ abstract contract Endpoint is Pausable {
         pure
         virtual
         returns (EndpointStructs.EndpointMessage memory endpointMessage);
-
-    /// @notice pause the endpoint
-    function _pauseEndpoint() internal {
-        _pause();
-    }
 }
--- a/src/EndpointAndManager.sol
+++ b/src/EndpointAndManager.sol
@@ -38,6 +38,14 @@ abstract contract EndpointAndManager is Endpoint, Manager, Implementation {
         _upgrade(newImplementation);
     }
 
+    /// @notice This function is used to pause the manager and the endpoint
+    function pause() public override onlyOwner {
+        // pause the manager
+        super.pause();
+        // pause the endpoint
+        _pauseEndpoint();
+    }
+
     function quoteDeliveryPrice(uint16 recipientChain) public view override returns (uint256) {
         return _quoteDeliveryPrice(recipientChain);
     }
--- a/src/EndpointAndManager.sol
+++ b/src/EndpointAndManager.sol
@@ -38,14 +38,6 @@ abstract contract EndpointAndManager is Endpoint, Manager, Implementation {
         _upgrade(newImplementation);
     }
 
-    /// @notice This function is used to pause the manager and the endpoint
-    function pause() public override onlyOwner {
-        // pause the manager
-        super.pause();
-        // pause the endpoint
-        _pauseEndpoint();
-    }
-
     function quoteDeliveryPrice(uint16 recipientChain) public view override returns (uint256) {
         return _quoteDeliveryPrice(recipientChain);
     }
--- a/src/EndpointStandalone.sol
+++ b/src/EndpointStandalone.sol
@@ -44,6 +44,10 @@ abstract contract EndpointStandalone is
         _upgrade(newImplementation);
     }
 
+    /// @notice pause the endpoint
+    /// TODO: add in pauser role
+    function pauseEndpoint() external virtual {}
+
     /// @notice Called by the BridgeManager contract to send a cross-chain message.
     function sendMessage(
         uint16 recipientChain,
--- a/src/Manager.sol
+++ b/src/Manager.sol
@@ -19,7 +19,7 @@ import "./interfaces/IManagerEvents.sol";
 import "./interfaces/IEndpointToken.sol";
 import "./Endpoint.sol";
 import "./EndpointRegistry.sol";
-import "./Pausable.sol";
+import "./libraries/Pausable.sol";
 
 // TODO: rename this (it's really the business logic)
 abstract contract Manager is
--- a/src/Manager.sol
+++ b/src/Manager.sol
@@ -136,6 +136,9 @@ abstract contract Manager is
         emit MessageAttestedTo(digest, endpoint, _getEndpointInfosStorage()[endpoint].index);
     }
 
+    /*
+     * @dev pause the Manager contract. Only the Manager should be able to do this
+     */
     function pauseManager() external onlyOwner {
         _pause();
     }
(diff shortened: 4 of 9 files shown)


nik-suri (commented):
We have pauseEndpoint in the WormholeEndpointStandalone but but don't have any of the Wormhole functions as pausable (like receiveMessage and receiveWormholeMessages). Should we make WormholeEndpoint Pausable instead and then we can guard the proper methods in the derived classes as well?

djb15 (commented):
Any reason we're not using the OZ implementation rather than spinning our own? https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/master/contracts/utils/PausableUpgradeable.sol

RahulMaganti47:
yea im extending this to include the pauser role in a later PR. that also doesn't have custom errors

src/Pausable.sol
```diff
@@ -0,0 +1,108 @@
+// SPDX-License-Identifier: UNLICENSED
```
djb15, line 1: Can we move this file to the libraries folder?

src/Manager.sol
```diff
@@ -163,6 +172,7 @@ abstract contract Manager is
         external
         payable
         nonReentrant
+        whenNotPaused
```
djb15, line 175: When we pause this path we're effectively locking user funds in the protocol...just checking this is what we want?

Maybe we should only pause starting new transfers but in-flight/queued transfers can be completed?

src/Pausable.sol
```diff
@@ -0,0 +1,108 @@
...
+    function _unpause() internal virtual whenPaused {
+        // this can only be set to NOTPAUSED when the state is PAUSED
+        _setPauseStorage(NOT_PAUSED);
+        emit NotPaused(false);
+    }
+
+    /**
+     * @dev Returns true if the method is paused, and false otherwise.
+     */
+    function _isPaused() public view returns (bool) {
```
djb15, line 104: No leading underscore for public method

src/Pausable.sol
```diff
@@ -0,0 +1,108 @@
...
+
+    struct PauseStorage {
+        uint256 _pauseFlag;
+    }
+
+    // @dev Emitted when the contract is paused
+    event Paused(bool paused);
+    event NotPaused(bool notPaused);
+
+    bytes32 public constant PAUSE_SLOT = bytes32(uint256(keccak256("Pause.pauseFlag")) - 1);
```
djb15, line 39: We need to initialise the storage slot to `NOT_PAUSED` in an initialiser. Right now the slot will be 0 by default

src/WormholeEndpointStandalone.sol
```diff
@@ -14,6 +14,12 @@ contract WormholeEndpointStandalone is WormholeEndpoint, EndpointStandalone, Own
         address wormholeRelayerAddr
     ) EndpointStandalone(manager) WormholeEndpoint(wormholeCoreBridge, wormholeRelayerAddr) {}
 
+    /// @notice This function is used to pause the endpoint
+    /// Only the implementor (deployer) can call this function
```
djb15, line 18: This comment won't be true soon so probably worth removing/updating

djb15 (changes requested):

src/Pausable.sol
RahulMaganti47: i believe it is initialized: https://github.com/wormhole-foundation/example-native-token-transfers/blob/42e0c3bcf0558dcba56a3ea5c56586b37d00888a/src/Pausable.sol#L94

src/Manager.sol
RahulMaganti47: yea we discussed this earlier. the idea was to add pausing to every entrypoint but this could be better

src/WormholeEndpointStandalone.sol
RahulMaganti47: true

src/Manager.sol
djb15: Although actually in an exploit scenario we'd also want to pause queued transfers.

How about we don't pause on the receiving side though so that everything in-flight can get redeemed on the destination chain? To avoid having to keep track of VAAs and keep retrying them?

RahulMaganti47: let's keep it in for now and then we can remove later?

djb15: Sure, I've created https://github.com/wormhole-foundation/example-native-token-transfers/issues/99 to keep track of that future discussion

src/libraries/external/Pausable.sol
```diff
@@ -0,0 +1,108 @@
+// SPDX-License-Identifier: UNLICENSED
```
djb15, line 1: It's not an external library so should go in the root /libraries folder imo

RahulMaganti47: put it there b/c it replicates lots of the OZ functionality but yea can move it to the root

src/WormholeEndpoint.sol
```diff
@@ -8,8 +8,9 @@ import "./interfaces/IWormhole.sol";
 import "./interfaces/IWormholeEndpoint.sol";
 import "./Endpoint.sol";
 import "wormhole-solidity-sdk/libraries/BytesParsing.sol";
+import "./libraries/Pausable.sol";
 
-abstract contract WormholeEndpoint is Endpoint, IWormholeEndpoint, IWormholeReceiver {
+abstract contract WormholeEndpoint is Endpoint, IWormholeEndpoint, IWormholeReceiver, Pausable {
```
djb15, line 13: Should we inherit `Pausable` in the Endpoint contract instead? Pausing an endpoint shouldn't be Wormhole specific if pausing a Manager isn't WH specific right?

RahulMaganti47: yea that's fair

src/EndpointAndManager.sol
```diff
@@ -38,6 +38,14 @@ abstract contract EndpointAndManager is Endpoint, Manager, Implementation {
...
     }
 
+    /// @notice This function is used to pause the manager and the endpoint
+    function pause() public override onlyOwner {
+        // pause the manager
+        super.pause();
+        // pause the endpoint
+        _pauseEndpoint();
+    }
+
```
djb15, line 48: I've just realised that maybe we don't need this virtual + override behaviour? The inheritance of Pausable will be linearised so here we're basically calling `_pause` twice and the second pause will fail?

RahulMaganti47: yea i think it will be linearized. we just need to call one of these functions to pause. The second one won't fail, however. pausing idempotent, so the behavior is the same but yea we don't need to call this twice

djb15: I think the `whenNotPaused` modifier on `_pause` will cause a revert, no?

RahulMaganti47: ah actually we may want to consider removing the modifier on _pause

src/EndpointStandalone.sol
```diff
@@ -44,6 +44,10 @@ abstract contract EndpointStandalone is
         _upgrade(newImplementation);
     }
 
+    /// @notice pause the endpoint
+    /// TODO: add in pauser role
+    function pauseEndpoint() external virtual {}
```
nik-suri, line 49: This should call into `Pausable._pause` right?

src/WormholeEndpointStandalone.sol
```diff
@@ -28,4 +28,8 @@ contract WormholeEndpointStandalone is WormholeEndpoint, EndpointStandalone, Own
     function setIsWormholeEvmChain(uint16 chainId) external onlyOwner {
         _setIsWormholeEvmChain(chainId);
     }
+
+    function pauseEndpoint() external override onlyOwner {
+        _pauseEndpoint();
+    }
```
nik-suri, line 34: This method should only be implemented in `EndpointStandalone.sol`

src/EndpointAndManager.sol
nik-suri: I think this method should only have a single call to `_pause`, which would override the `pause()` functions on both the Endpoint and Manager contracts.

src/Endpoint.sol
```diff
@@ -15,4 +16,9 @@ abstract contract Endpoint {
         pure
         virtual
         returns (EndpointStructs.EndpointMessage memory endpointMessage);
+
+    /// @notice pause the endpoint
+    function _pauseEndpoint() internal {
+        _pause();
+    }
```
nik-suri, line 23: Do we need this function? It's effectively just a pass-through to `_pause` since it's also internal.
Endpoint implementations such as `EndpointStandalone` can call directly into `Pausable._pause` instead of calling `_pauseEndpoint`.

nik-suri (changes requested):

djb15 (approved):

nik-suri (approved):

PR #98, opened 2024-02-09
implement configurable pauser role

RahulMaganti47:
Adds #93


src/libraries/Pausable.sol
```diff
@@ -26,17 +41,37 @@ abstract contract Pausable is Initializable {
...
+        uint256 slot = uint256(PAUSER_ROLE_SLOT);
+        assembly ("memory-safe") {
+            $.slot := slot
+        }
+    }
+
+    /**
+     * @dev Returns the current pauser account address.
+     */
+    function _pauser() private view returns (address) {
```
djb15, line 72: Worth having this method public as `pauser()` (without the leading underscore)? I feel like being able to read from elsewhere could be useful

src/libraries/Pausable.sol
```diff
@@ -71,19 +118,27 @@ abstract contract Pausable is Initializable {
...
+        if (_pauser() != msg.sender) {
+            revert InvalidPauser(msg.sender);
+        }
     }
 
     /**
      * @dev pauses the function and emits the `Paused` event
      */
-    function _pause() internal virtual whenNotPaused {
+    function _pause() internal virtual whenNotPaused onlyPauser {
```
djb15, line 141: I thought we wanted both the owner and pauser to be able to pause? You'll need to create a helper modifier and probably use that with the `external` pause methods in the inheriting contracts

src/libraries/Pausable.sol
```diff
@@ -105,4 +160,28 @@ abstract contract Pausable is Initializable {
         PauseStorage storage $ = _getPauseStorage();
         return $._pauseFlag == PAUSED;
     }
+
+    /**
+     * @dev Transfers the ability to pause to a new account (`newPauser`).
+     */
+    function _transferPauserCapability(address newPauser) internal virtual onlyPauser {
```
djb15, line 167: Not sure whether we want the owner or pauser to control the next pauser?

src/libraries/Pausable.sol
```diff
@@ -105,4 +160,28 @@ abstract contract Pausable is Initializable {
...
+        $._pauser = newPauser;
+        emit PauserTransferred(oldPauser, newPauser);
+    }
+
+    /**
+     * @dev Leaves the contract without a Pauser
+     * and all the capabilities afforded with that role.
+     * TODO: do we need this?
+     */
+    function renouncePauser() public virtual onlyPauser {
```
djb15, line 179: If the owner has the ability to update the pauser then I'm fine with keeping this in. This PR will need some changes to the inheriting contracts

src/libraries/Pausable.sol
RahulMaganti47: my thinking for now was just to collapse the pauser role to owner for now given we don't expect the pauser to be distinct from the owner. (although it is a simple change - we could extend it now).

src/libraries/Pausable.sol
RahulMaganti47: i think it could be either. if the owner and the pauser are distinct, the owner already trusts the pauser to guarantee liveness.

src/libraries/Pausable.sol
djb15: I would personally vote for extending that now as it's not a heavy lift but we can discuss in the standup later if needed

src/libraries/Pausable.sol
djb15: Yeah same as above, with the two roles distinct I think it makes sense to allow both to specify the next pauser

src/libraries/Pausable.sol
RahulMaganti47: sure

src/libraries/Pausable.sol
nik-suri: Agreed, I think both owner and pauser should be able to specify the next pauser (in case the pauser specifies a new pauser that's malicious).

src/libraries/Pausable.sol
nik-suri: I'm also for having the owner and pauser be able to pause.
We could have an `onlyPauser` modifier and another `onlyPauserAndOwner(address owner)` modifier?

src/libraries/Pausable.sol
nik-suri: Agreed, if owner has ability to update pauser we should be good to keep this.

nik-suri (commented):
Looks good -- could you please also add the proper modifiers to the pause methods that are exposed in the inheriting contracts?

src/libraries/Pausable.sol
```diff
@@ -71,19 +121,44 @@ abstract contract Pausable is Initializable {
...
+        if (pauser() != msg.sender) {
+            revert InvalidPauser(msg.sender);
+        }
+    }
+
+    /*
+     * @dev Modifier to allow only the Pauser to access some functionality
+     */
+    function _checkOwnerOrPauser(address owner) internal view {
+        if (pauser() != msg.sender || owner != msg.sender) {
```
nik-suri, line 153: This should be an `&&` right?

src/libraries/Pausable.sol
```diff
@@ -49,6 +84,21 @@ abstract contract Pausable is Initializable {
         _getPauseStorage()._pauseFlag = pauseFlag;
     }
 
+    function __Paused_init(address initialPauser, address owner) internal onlyInitializing {
```
djb15, line 87: I don't think we're calling this initialiser from anywhere, or at least I can't see find it?

src/libraries/Pausable.sol
```diff
@@ -49,6 +84,21 @@ abstract contract Pausable is Initializable {
...
+        __Paused_init_unchained(initialPauser, owner);
+    }
+
+    function __Paused_init_unchained(
```
djb15, line 91: On init I don't think we need to worry about the owner do we? Can we just set the initial pauser directly rather than calling `_transferPauserCapability`?

src/libraries/Pausable.sol
```diff
@@ -71,19 +121,44 @@ abstract contract Pausable is Initializable {
         _;
     }
 
-    function __Paused_init() internal onlyInitializing {
-        __Paused_init_unchained();
+    /*
+     * @dev Modifier to allow only the Pauser to access pausing functionality
+     */
+    modifier onlyPauser() {
```
djb15, line 127: We're not using this anywhere? Or do you want to keep this around for integrators to use?

RahulMaganti47: yea just keeping it for integrators

src/libraries/Pausable.sol
RahulMaganti47: its in now

src/libraries/Pausable.sol
```diff
@@ -105,4 +179,31 @@ abstract contract Pausable is Initializable {
...
+        $._pauser = newPauser;
+        emit PauserTransferred(oldPauser, newPauser);
+    }
+
+    /**
+     * @dev Leaves the contract without a Pauser
+     * and all the capabilities afforded with that role.
+     * TODO: do we need this?
+     */
+    function renouncePauser(address owner) public virtual onlyOwnerOrPauser(owner) {
```
djb15, line 201: Anyone can now pause the contract by specifying their own address as the pauser :(

djb15 (commented):
Instead of passing through the owner down to this contract can we instead have the `onlyOwnerOrPauser` (can be a modifier or a function) logic in the inheriting contracts that also inherit the Ownable logic? You can then move the modifiers from the internal methods in Pausable.sol to the external methods in the contracts that inherit it

src/Manager.sol
```diff
@@ -29,7 +29,7 @@ abstract contract Manager is
     RateLimiter,
     OwnableUpgradeable,
```
djb15, line 30: We don't have to inherit OwnableUpgradeable here anymore right?

src/libraries/Pausable.sol
```diff
@@ -12,7 +12,7 @@ pragma solidity >=0.8.8 <0.9.0;
  */
 import {Initializable} from "./external/Initializable.sol";
 
-abstract contract Pausable is Initializable {
+abstract contract PausableUpgradeable is Initializable {
```
djb15, line 15: Can you rename the file as well?

src/libraries/PausableOwnable.sol
```diff
@@ -0,0 +1,52 @@
...
+
+    function __PausedOwnable_init(address initialPauser, address owner) internal onlyInitializing {
+        __Paused_init(initialPauser);
+        __Ownable_init(owner);
+    }
+
+    /**
+     * @dev Transfers the ability to pause to a new account (`newPauser`).
+     */
+    function _transferPauserCapability(address newPauser) internal virtual onlyOwnerOrPauser {
```
djb15, line 34: This needs to be external/public now I think?

src/libraries/PausableOwnable.sol
```diff
@@ -0,0 +1,52 @@
...
+        PauserStorage storage $ = _getPauserStorage();
+        address oldPauser = $._pauser;
+        $._pauser = newPauser;
+        emit PauserTransferred(oldPauser, newPauser);
+    }
+
+    /**
+     * @dev Leaves the contract without a Pauser
+     */
+    function _renouncePauser() public virtual onlyOwnerOrPauser {
```
djb15, line 44: No leading underscore for external/public methods please.

Or alternatively if this is supposed to be internal then it should sit behind the `renouncePauser` virtual method...it's only defined in EndpointStandalone so far but should probably be defined here instead so it can be used by the manager and overriden if needed

RahulMaganti47: yea let's leave it public. removed the virtual method from EndpointStandalone. Manager directly call now.

PR #102, opened 2024-02-12
Normalize amounts correctly based on number of decimals

### LINKED ISSUE #91: normalize amounts to 8 decimals in all cases
There is no requirement for the number of decimals on a token to be the same on every chain. (e.g. a token amount with less than 8 decimals needs to send along the number of decimals so the denormalization can be handled properly).

RahulMaganti47:
Fixes #91

### DIFF
--- a/foundry.toml
+++ b/foundry.toml
@@ -2,7 +2,7 @@
 solc_version = "0.8.19"
 optimizer = true
 optimizer_runs = 200
-via_ir = true
+via_ir = false
 evm_version = "london"
 src = "src"
 out = "out"
--- a/foundry.toml
+++ b/foundry.toml
@@ -2,7 +2,7 @@
 solc_version = "0.8.19"
 optimizer = true
 optimizer_runs = 200
-via_ir = false
+via_ir = true
 evm_version = "london"
 src = "src"
 out = "out"
--- a/src/libraries/EndpointStructs.sol
+++ b/src/libraries/EndpointStructs.sol
@@ -6,6 +6,7 @@ import "./NormalizedAmount.sol";
 
 library EndpointStructs {
     using BytesParsing for bytes;
+    using NormalizedAmountLib for NormalizedAmount;
 
     error PayloadTooLong(uint256 size);
     error IncorrectPrefix(bytes4 prefix);
@@ -75,7 +76,7 @@ library EndpointStructs {
     ///      - to - 32 bytes
     ///      - toChain - 2 bytes
     struct NativeTokenTransfer {
-        /// @notice Amount being transferred (big-endian uint256)
+        /// @notice Amount being transferred (big-endian u64 and u8 for decimals)
         NormalizedAmount amount;
         /// @notice Source chain token address.
         bytes32 sourceToken;
@@ -90,7 +91,15 @@ library EndpointStructs {
         pure
         returns (bytes memory encoded)
     {
-        return abi.encodePacked(NTT_PREFIX, m.amount, m.sourceToken, m.to, m.toChain);
+        NormalizedAmount memory transferAmount = m.amount;
+        return abi.encodePacked(
+            NTT_PREFIX,
+            transferAmount.getDecimals(),
+            transferAmount.getAmount(),
+            m.sourceToken,
+            m.to,
+            m.toChain
+        );
     }
 
     /*
@@ -109,9 +118,13 @@ library EndpointStructs {
         if (prefix != NTT_PREFIX) {
             revert IncorrectPrefix(prefix);
         }
+
+        uint8 numDecimals;
+        (numDecimals, offset) = encoded.asUint8Unchecked(offset);
         uint64 amount;
         (amount, offset) = encoded.asUint64Unchecked(offset);
-        nativeTokenTransfer.amount = NormalizedAmount.wrap(amount);
+        nativeTokenTransfer.amount = NormalizedAmountLib.wrap(amount, numDecimals);
+
         (nativeTokenTransfer.sourceToken, offset) = encoded.asBytes32Unchecked(offset);
         (nativeTokenTransfer.to, offset) = encoded.asBytes32Unchecked(offset);
         (nativeTokenTransfer.toChain, offset) = encoded.asUint16Unchecked(offset);
--- a/src/libraries/EndpointStructs.sol
+++ b/src/libraries/EndpointStructs.sol
@@ -71,8 +71,8 @@ library EndpointStructs {
     /// @dev Native Token Transfer payload.
     ///      The wire format is as follows:
     ///      - NTT_PREFIX - 4 bytes
-    ///      - amount - 8 bytes
     ///      - numDecimals - 1 byte
+    ///      - amount - 8 bytes
     ///      - sourceToken - 32 bytes
     ///      - to - 32 bytes
     ///      - toChain - 2 bytes
--- a/src/libraries/EndpointStructs.sol
+++ b/src/libraries/EndpointStructs.sol
@@ -72,6 +72,7 @@ library EndpointStructs {
     ///      The wire format is as follows:
     ///      - NTT_PREFIX - 4 bytes
     ///      - amount - 8 bytes
+    ///      - numDecimals - 1 byte
     ///      - sourceToken - 32 bytes
     ///      - to - 32 bytes
     ///      - toChain - 2 bytes
--- a/src/libraries/EndpointStructs.sol
+++ b/src/libraries/EndpointStructs.sol
@@ -123,7 +123,7 @@ library EndpointStructs {
         (numDecimals, offset) = encoded.asUint8Unchecked(offset);
         uint64 amount;
         (amount, offset) = encoded.asUint64Unchecked(offset);
-        nativeTokenTransfer.amount = NormalizedAmountLib.wrap(amount, numDecimals);
+        nativeTokenTransfer.amount = NormalizedAmount(amount, numDecimals);
 
         (nativeTokenTransfer.sourceToken, offset) = encoded.asBytes32Unchecked(offset);
         (nativeTokenTransfer.to, offset) = encoded.asBytes32Unchecked(offset);
--- a/src/libraries/NormalizedAmount.sol
+++ b/src/libraries/NormalizedAmount.sol
@@ -133,14 +133,15 @@ library NormalizedAmountLib {
     }
 
     /// @dev scale the amount from origDecimals to normDecimals (base 10)
-    function scalingFactor(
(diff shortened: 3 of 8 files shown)


src/libraries/RateLimiter.sol
```diff
@@ -165,30 +171,35 @@ abstract contract RateLimiter is IRateLimiter, IRateLimiterEvents {
...
         } else {
-            difference = newLimit - oldLimit;
-            newCurrentCapacity = currentCapacity + difference;
+            difference = newLimit.sub(oldLimit);
+            newCurrentCapacity = currentCapacity.add(difference);
+        }
+
+        // TODO: change this to min
+        if (newCurrentCapacity.gt(newLimit)) {
+            revert CapacityCannotExceedLimit(newCurrentCapacity, newLimit);
```
nik-suri, line 192: The TODO here says to change this to use `min` instead.
Why not keep the invariant check?

nik-suri (commented):
Generally looks good to me, though will take another pass through the new normalization logic.

src/libraries/NormalizedAmount.sol
```diff
@@ -1,58 +1,137 @@
...
+    }
 
-function min(NormalizedAmount a, NormalizedAmount b) pure returns (NormalizedAmount) {
-    return a > b ? b : a;
-}
+    function gt(
+        NormalizedAmount memory a,
+        NormalizedAmount memory b
+    ) internal pure returns (bool) {
+        return a.amount > b.amount;
```
kcsongor, line 37: let's assert that the decimals are the same

src/libraries/NormalizedAmount.sol
```diff
@@ -1,58 +1,137 @@
...
+    }
 
-    function normalize(uint256 amount, uint8 decimals) internal pure returns (NormalizedAmount) {
-        if (decimals > 8) {
-            amount /= 10 ** (decimals - 8);
+    function sub(
+        NormalizedAmount memory a,
+        NormalizedAmount memory b
+    ) internal pure returns (NormalizedAmount memory) {
+        if (a.amount < b.amount) {
```
kcsongor, line 55: this is redundant, solidity inserts underflow checks so now we're doing it twice

src/libraries/NormalizedAmount.sol
```diff
@@ -1,58 +1,137 @@
...
+            revert AmountUnderflows(a.amount, b.amount);
         }
-        // amount after normalization must fit into uint64 (that's the point of
-        // normalization, as Solana only supports uint64 for token amounts)
-        if (amount > type(uint64).max) {
-            revert AmountTooLarge(amount);
+
+        // on initialization
```
kcsongor, line 59: not sure if we need this check?

src/libraries/NormalizedAmount.sol
```diff
@@ -1,58 +1,137 @@
...
+            revert NumberOfDecimalsNotEqual(a.decimals, b.decimals);
+        }
+        return wrap(a.amount + b.amount, a.decimals);
+    }
+
+    function min(
+        NormalizedAmount memory a,
+        NormalizedAmount memory b
+    ) public pure returns (NormalizedAmount memory) {
+        return a.amount < b.amount ? a : b;
```
kcsongor, line 94: check decimals are the same

src/libraries/NormalizedAmount.sol
```diff
@@ -1,58 +1,137 @@
...
+    }
+
+    function minDecimals(uint8 toDecimals, uint8 fromDecimals) internal pure returns (uint8) {
```
kcsongor, line 97: this is just a regular min function on u8s

src/libraries/NormalizedAmount.sol
```diff
@@ -1,58 +1,137 @@
...
 // SPDX-License-Identifier: Apache 2
 pragma solidity >=0.8.8 <0.9.0;
 
-type NormalizedAmount is uint64;
+struct NormalizedAmount {
+    uint64 amount;
+    uint8 decimals;
+}
 
-using {gt as >, lt as <, sub as -, add as +, div, min, unwrap} for NormalizedAmount global;
```
kcsongor, line 6: why remove these? we should keep them, then the downstream code wouldn't have to change at all I think

src/libraries/NormalizedAmount.sol
```diff
@@ -1,58 +1,137 @@
...
-    return NormalizedAmount.unwrap(a) < NormalizedAmount.unwrap(b);
-}
+    function wrap(uint64 amt, uint8 dec) internal pure returns (NormalizedAmount memory) {
+        return NormalizedAmount(amt, dec);
+    }
 
-function sub(NormalizedAmount a, NormalizedAmount b) pure returns (NormalizedAmount) {
-    return NormalizedAmount.wrap(NormalizedAmount.unwrap(a) - NormalizedAmount.unwrap(b));
-}
+    function unwrap(NormalizedAmount memory a) internal pure returns (uint64, uint8) {
```
kcsongor, line 21: we can get rid of wrap/unwrap and just use the constructor directly

src/libraries/NormalizedAmount.sol
RahulMaganti47: yea afaik there's no `min` function defined on uint8's in the standard library.

src/libraries/NormalizedAmount.sol
RahulMaganti47: operators can only be defined for user-defined value types. value-types can only be defined for elementary types

kcsongor: oh you're right. what a nonsensical restriction...
we could do
```
using {gt, lt, sub, add, div, min, unwrap} for NormalizedAmount global;
```
so we can write `foo.gt(bar)` instead of `NormalizedAmountLib.gt(foo, bar)`

RahulMaganti47: i can already do that if the function is part of the library.

src/libraries/NormalizedAmount.sol
RahulMaganti47: i'd prefer to keep the unwrap method given that solidity doesn't have support for destructuring types via pattern matching (just nice to have a single method that cleanly destructures).

src/libraries/NormalizedAmount.sol
RahulMaganti47: which check, the isZero?

foundry.toml
```diff
@@ -2,7 +2,7 @@
 solc_version = "0.8.19"
 optimizer = true
 optimizer_runs = 200
-via_ir = false
+via_ir = true
```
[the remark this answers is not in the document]
RahulMaganti47: yea i think we're getting pretty close to the stack limit. ill play around a little bit to see if we can reduce.

src/libraries/RateLimiter.sol
RahulMaganti47: yea i actually prefer the revert here so it propagates the error up the call stack. if the capacity exceeds the limit, something most likely went wrong

foundry.toml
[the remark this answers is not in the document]
RahulMaganti47: fixed.

src/libraries/NormalizedAmount.sol
```diff
@@ -1,58 +1,171 @@
...
+        }
+        return NormalizedAmount(uint64(amt), toDecimals);
+    }
+
+    function denormalize(
+        NormalizedAmount memory amt,
+        uint8 toDecimals
+    ) internal pure returns (uint256) {
+        (uint256 deNorm, uint8 dec) = unwrap(amt);
+        deNorm *= scalingFactor(toDecimals, dec);
```
djb15, line 167: Correct me if I'm wrong but I don't think this denormalises correctly when `toDecimals` < 8 and `fromDecimals` >= 8?

For example if normalise 1e8 of a token with 8 decimals on the source chain we'd have (1e8, 8) in the wire format. Then when we try to denormalise on the destination chain to `toDecimals = 6` the scaling factor is set to 1 and the amount out is 1e8. So we end up getting 100 tokens on the destination chain instead of 1

RahulMaganti47: another way to say this is that denormalization strictly depends on normalization.

nik-suri: What if we have the following case:
- Token on Eth -- 18 decimals
- Token on Arbitrum -- 6 decimals
Token on Eth will get normalized to 8 decimals and transferred over. Then on the Arbitrum side, the 8 decimals will get scaled by 1 and remain 8 decimals so we get 100x more tokens as @djb15 is saying.

djb15: Yeah my original message is overly constrained as well. I think this is broken when `toDecimals` < `normalised(fromDecimals)`

src/libraries/EndpointStructs.sol
```diff
@@ -71,11 +72,12 @@ library EndpointStructs {
     ///      The wire format is as follows:
     ///      - NTT_PREFIX - 4 bytes
     ///      - amount - 8 bytes
+    ///      - numDecimals - 1 byte
```
djb15, line 75: Can you update the docstring to have the wire format in the correct order pls?

djb15 (approved):

src/libraries/EndpointStructs.sol
RahulMaganti47: it is in the right order.
```struct NormalizedAmount {
    uint64 amount;
    uint8 decimals;
}
```

djb15: Not in the encoded payload it isn't: https://github.com/wormhole-foundation/example-native-token-transfers/blob/a614dd8213fdb47967945f6b7d62e9ee83e2a4c3/src/libraries/EndpointStructs.sol#L96-L102. One or the other should be flipped around I think

src/libraries/NormalizedAmount.sol
```diff
@@ -1,58 +1,172 @@
...
+struct NormalizedAmount {
+    uint64 amount;
+    uint8 decimals;
 }
 
-function lt(NormalizedAmount a, NormalizedAmount b) pure returns (bool) {
-    return NormalizedAmount.unwrap(a) < NormalizedAmount.unwrap(b);
+function minUint8(uint8 a, uint8 b) pure returns (uint8) {
+    return a < b ? a : b;
 }
```
nik-suri, line 12: very nitpicky but what do you think of moving this method to `EndpointHelpers.sol`
There is a `min(uint256,uint256)` method already in there.

nik-suri (approved):

PR #104, opened 2024-02-13
Fix quoting logic for EndpointAndManager

nik-suri:


djb15 (dismissed):
Ideally would like some more tests to verify this is doing what we want, but the changes look good to me

nik-suri:
@djb15 writing tests as part of https://github.com/wormhole-foundation/example-native-token-transfers/issues/95

djb15 (approved):

PR #109, opened 2024-02-13
Do not auto-increment threshold with new endpoint set

nik-suri:


djb15 (approved):

djb15 (approved):

djb15:
Arguably we could also edit the event to either not emit the threshold (since it hasn't changed), or emit the threshold and the number of enabled endpoints (to make it clear what the new quorum requirement is). But logically this PR is good

djb15 (approved):

PR #116, opened 2024-02-15
Rename EndpointToken to NTTToken and include abstract contract

nik-suri:


src/NTTToken.sol
```diff
@@ -0,0 +1,41 @@
+// SPDX-License-Identifier: Apache 2
```
RahulMaganti47, line 1: is this meant to be a mock? if so, can you move it into a mocks/ dir?

PR #119, opened 2024-02-15
restructure ownership logic

### LINKED ISSUE #94: Reflect Manager ownership on the Endpoint contracts
```[tasklist]
### Tasks
- [x] remove OwnableUpgradeable from WormholeEndpointStandalone
- [x] set the owner of the manager and to the owner of the endpoint
- [x] override transferOwnership in ManagerStandalone to loop through all the endpoints and update the owner of all the endpoints to be the owner of the manager
- [x] remove pass through upgrade of endpoints from manager
```

RahulMaganti47:
Fixes #94


src/ManagerStandalone.sol
```diff
@@ -46,6 +46,18 @@ contract ManagerStandalone is IManagerStandalone, Manager, Implementation {
...
 
+    /// @dev Transfer ownership of the Manager contract and all Endpoint contracts to a new owner.
+    function transferOwnership(address newOwner) public override onlyOwner {
+        super.transferOwnership(newOwner);
+        // loop through all the registered endpoints and set the new owner of each endpoint to the newOwner
+        address[] storage _registeredEndpoints = _getRegisteredEndpointsStorage();
+        _checkRegisteredEndpointsInvariants();
+
+        for (uint256 i = 0; i < _registeredEndpoints.length; i++) {
+            IEndpointStandalone(_registeredEndpoints[i]).initializeEndpoint();
```
nik-suri, line 57: Why do we need to re-initialize here? We might run into some unintended behavior.
Why not expose a method that's `onlyManager` in the `EndpointStandalone` that allows transferring ownership directly?

src/EndpointStandalone.sol
```diff
@@ -6,6 +6,7 @@ import "./interfaces/IManagerStandalone.sol";
 import "./interfaces/IEndpointStandalone.sol";
 import "./libraries/Implementation.sol";
 import "./libraries/external/ReentrancyGuardUpgradeable.sol";
+import "./Manager.sol";
```
djb15, line 9: Can we import the interface instead?

src/Manager.sol
```diff
@@ -112,6 +111,8 @@ abstract contract Manager is
         __Ownable_init(msg.sender);
```
djb15, line 111: Think we should get rid of this now

src/ManagerStandalone.sol
djb15: Yeah agreed, can we override `transferOwnership` in the endpoints to only allow calls from the manager?

src/Manager.sol
```diff
@@ -150,6 +148,11 @@ abstract contract Manager is
         _pause();
     }
 
+    /// @notice Returns the address of the owner of this contract.
+    function getOwner() external view returns (address) {
```
djb15, line 152: Any particular reason you've wrapped `owner()` here? Unless I'm missing something I would prefer without the wrapping

src/EndpointStandalone.sol
```diff
@@ -28,11 +29,22 @@ abstract contract EndpointStandalone is
         _;
     }
 
+    /// @dev Returns the owner of the manager contract
+    function getManagerOwner() public view returns (address) {
+        return IManager(manager).getOwner();
+    }
+
+    /// @dev transfer the ownership of the endpoint to a new address
+    function transferEndpointOwnership(address newOwner) external {
```
djb15, line 38: This needs to change to have an `onlyManager` modifier guard and the call should go to `_transferOwnership` internal method. Currently this will always revert on calls from the manager because the manager is not the owner. 

We want the owner of the endpoints to stay in sync with the manager owner so we also need to override `transferOwnership` and `renounceOwnership` to do nothing (to prevent the owner from going out of sync with the manager owner).

We'll also need to override `renounceOwnership` (or get rid of it altogether) in the Manager to prevent the the manager owner from being renounced as I don't think we should support that case, what do you think?

src/Manager.sol
RahulMaganti47: couldn't access owner() directly by just importing the interface

djb15: Ah can you instead add an interface for OwnableUpgradeable and make sure it's accessible that way pls?

src/EndpointStandalone.sol
RahulMaganti47: overrode the methods directly in the concrete implementations

src/ManagerStandalone.sol
```diff
@@ -29,6 +29,13 @@ contract ManagerStandalone is IManagerStandalone, Manager, Implementation {
         _checkEndpointsInvariants();
     }
 
+    /// @dev Override the [`renounceOwnership`] function to ensure
+    /// the manager ownership is not renounced.
+    function renounceOwnership() public override onlyOwner {
+        // do nothing
+        // this method body is empty
+    }
```
nik-suri, line 37: should we revert here instead of doing nothing?

src/WormholeEndpointAndManager.sol
```diff
@@ -29,4 +29,11 @@ contract WormholeEndpointAndManager is EndpointAndManager, WormholeEndpoint {
     function setIsWormholeEvmChain(uint16 chainId) external onlyOwner {
         _setIsWormholeEvmChain(chainId);
     }
+
+    /// @dev Override the [`renounceOwnership`] function to ensure
+    /// the manager ownership is not renounced.
+    function renounceOwnership() public override onlyOwner {
+        // do nothing
+        // this method body is empty
+    }
```
nik-suri, line 38: same here -- should we revert instead?

src/WormholeEndpointStandalone.sol
```diff
@@ -29,4 +29,18 @@ contract WormholeEndpointStandalone is WormholeEndpoint, EndpointStandalone {
...
+        // do nothing
+        // this method body is empty
+    }
+
+    /// @dev Override the [`renounceOwnership`] function to ensure
+    /// the manager ownership is not renounced.
+    function renounceOwnership() public override onlyOwner {
+        // do nothing
+        // this method body is empty
+    }
```
nik-suri, line 45: reverts?

src/Manager.sol
djb15: I still dislike this pattern quite a lot :( The fix is to just wrap the manager address in `getManagerOwner` with the `IOwnable` interface and call `owner()` directly

src/ManagerStandalone.sol
djb15: Agree re reverting on all of these overrides

src/WormholeEndpointStandalone.sol
```diff
@@ -29,4 +29,18 @@ contract WormholeEndpointStandalone is WormholeEndpoint, EndpointStandalone {
     function pauseEndpoint() external override onlyOwnerOrPauser {
         _pause();
     }
+
+    /// @dev Override the [`transferOwnership`] method from OwnableUpgradeable
+    /// to ensure owner cannot transfer ownership
```
djb15, line 34: "..directly. The owner can still transfer ownership via the Manager. This is to ensure the owner of this contract is in sync with the owner of the Manager"

src/WormholeEndpointStandalone.sol
```diff
@@ -29,4 +29,18 @@ contract WormholeEndpointStandalone is WormholeEndpoint, EndpointStandalone {
...
+    function transferOwnership(address newOwner) public override onlyOwner {
+        // do nothing
+        // this method body is empty
+    }
+
+    /// @dev Override the [`renounceOwnership`] function to ensure
+    /// the manager ownership is not renounced.
```
djb15, line 41: **endpoint** ownership

src/Manager.sol
RahulMaganti47: ah yea just haven't pushed this change

PR #123, opened 2024-02-15
Fix: Don't allow a threshold of 0

djb15:
By default when we add our first endpoint the threshold will remain 0. Message from other chains will be received by the endpoint but will not be executed because we do not execute when threshold is 0. However if the endpoint has replay protection (which the WH endpoint does) then this messages is now stuck.

This fix makes sure that we have a threshold of 1 when the first endpoint has been added.


mdulin2:
Made the changes locally and ran tests. This solves the initial problem I was thinking about.

src/ManagerStandalone.sol
```diff
@@ -216,13 +224,14 @@ contract ManagerStandalone is IManagerStandalone, Manager, Implementation {
...
         address[] storage _enabledEndpoints = _getEnabledEndpointsStorage();
+        address[] storage _registeredEndpoints = _getRegisteredEndpointsStorage();
 
         // invariant: threshold <= enabledEndpoints.length
         if (_threshold.num > _enabledEndpoints.length) {
             revert ThresholdTooHigh(_threshold.num, _enabledEndpoints.length);
         }
 
-        if (_enabledEndpoints.length > 0) {
+        if (_registeredEndpoints.length > 0) {
```
mdulin2, line 234: Yeah, I think this makes sense to change. Before, I couldn't perform a migrate because this let's us migrate when none of the endpoints are turned on.

mdulin2 (approved):

RahulMaganti47 (dismissed):

nik-suri (approved):

PR #125, opened 2024-02-15
End to end tests for wormhole integration

mdulin2:
* Standalone with a single endpoint and multiple endpoints
* Standalone while using the relayer
* Upgrade tests to ensure invariant checking and migration works as expected

### DIFF
--- a/test/Upgrades.t.sol
+++ b/test/Upgrades.t.sol
@@ -527,7 +527,13 @@ contract TestUpgrades is Test, IManagerEvents, IRateLimiterEvents {
         {
             uint256 managerBalanceBefore = token1.balanceOf(address(managerChain1));
             uint256 userBalanceBefore = token1.balanceOf(address(userA));
-            managerChain1.transfer(sendingAmount, chainId2, bytes32(uint256(uint160(userB))), false, encodeEndpointInstruction(true));
+            managerChain1.transfer(
+                sendingAmount,
+                chainId2,
+                bytes32(uint256(uint160(userB))),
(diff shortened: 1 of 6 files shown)


mdulin2:
Mostly creating new tests. The only exception is modifying the `migrate()` function to be `virtual` in order to be able to override it during tests for upgrading.

test/Upgrades.t.sol
```diff
@@ -0,0 +1,679 @@
...
+            address(relayer)
+        ); 
+        managerChain1.upgradeEndpoint(address(wormholeEndpointChain1), address(wormholeEndpointChain1Implementation)); 
+
+        basicFunctionality();
+    }
+
+
+    // Confirm that we can handle multiple upgrades as a manager
+    function testDoubleUpgradeManager() public {
```
djb15, line 264: Nit-pick, but can all the test be named like `test_camelCase`

djb15:
Also I guess you need to get all the tests running before we can merge this in 😄

test/Upgrades.t.sol
mdulin2: The linter lied to me. Can do!

mdulin2:
> Also I guess you need to get all the tests running before we can merge this in 😄

I have the frontrunning of `initialize()` as a test case. I'll just comment it out for the time being.

djb15 (approved):

djb15 (approved):

RahulMaganti47 (approved):

Issue #24, opened 2024-01-19
Add docstrings for important methods and variables


nik-suri:
Self-explanatory. Let's make sure components are well-documented.

PR #129, opened 2024-02-16
Add WormholeEndpoint Special Relayer

nik-suri:
Add special relayer quoting logic to handle specialized target chain deliveries

### DIFF
--- a/src/Endpoint.sol
+++ b/src/Endpoint.sol
@@ -6,6 +6,7 @@ import "./libraries/PausableOwnable.sol";
 
 abstract contract Endpoint is PausableOwnable {
     function _sendMessage(
+        address token,
         uint16 recipientChain,
         uint256 deliveryPayment,
         address caller,
@@ -20,6 +21,7 @@ abstract contract Endpoint is PausableOwnable {
     ) internal virtual;
 
     function _quoteDeliveryPrice(
+        address token,
         uint16 targetChain,
         EndpointStructs.EndpointInstruction memory endpointInstruction
     ) internal view virtual returns (uint256);
--- a/src/Endpoint.sol
+++ b/src/Endpoint.sol
@@ -6,7 +6,6 @@ import "./libraries/PausableOwnable.sol";
 
 abstract contract Endpoint is PausableOwnable {
     function _sendMessage(
-        address token,
         uint16 recipientChain,
         uint256 deliveryPayment,
         address caller,
@@ -21,10 +20,11 @@ abstract contract Endpoint is PausableOwnable {
     ) internal virtual;
 
     function _quoteDeliveryPrice(
-        address token,
         uint16 targetChain,
         EndpointStructs.EndpointInstruction memory endpointInstruction
     ) internal view virtual returns (uint256);
+
+    function getManagerToken() public view virtual returns (address);
 
     /// @notice pause the endpoint
     function _pauseEndpoint() internal {
--- a/src/EndpointStandalone.sol
+++ b/src/EndpointStandalone.sol
@@ -23,6 +23,7 @@ abstract contract EndpointStandalone is
 
     constructor(address _manager) {
         manager = _manager;
+        // TODO: ERC-165 check on the manager contract, otherwise revert
         token = IManager(manager).token();
     }
 
--- a/src/EndpointStandalone.sol
+++ b/src/EndpointStandalone.sol
@@ -2,6 +2,7 @@
 pragma solidity >=0.8.8 <0.9.0;
 
 import "./Endpoint.sol";
+import "./interfaces/IManager.sol";
 import "./interfaces/IManagerStandalone.sol";
 import "./interfaces/IEndpointStandalone.sol";
 import "./libraries/Implementation.sol";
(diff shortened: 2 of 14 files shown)


src/EndpointStandalone.sol
```diff
@@ -56,11 +59,11 @@ abstract contract EndpointStandalone is
         EndpointStructs.EndpointInstruction memory instruction,
         bytes memory managerMessage
     ) external payable nonReentrant onlyManager {
-        _sendMessage(recipientChain, msg.value, msg.sender, instruction, managerMessage);
+        _sendMessage(token, recipientChain, msg.value, msg.sender, instruction, managerMessage);
     }
 
     function quoteDeliveryPrice(uint16 targetChain) external view override returns (uint256) {
-        return _quoteDeliveryPrice(targetChain);
+        return _quoteDeliveryPrice(token, targetChain);
```
RahulMaganti47, line 67: what about passing in the token as a function parameter instead of initializing a new state variable?

test/Manager.t.sol
```diff
@@ -53,21 +53,20 @@ contract EndpointAndManagerContract is EndpointAndManager, IEndpointReceiver {
...
-    function _quoteDeliveryPrice(uint16 /* recipientChain */ )
-        internal
-        pure
-        override
-        returns (uint256)
-    {
+    function _quoteDeliveryPrice(
+        address, /* token */
+        uint16 /* recipientChain */
+    ) internal pure override returns (uint256) {
```
RahulMaganti47, line 59: can you make these comments docstrings pls?

test/Manager.t.sol
```diff
@@ -53,21 +53,20 @@ contract EndpointAndManagerContract is EndpointAndManager, IEndpointReceiver {
...
-        uint256 deliveryPayment,
-        address caller,
-        EndpointStructs.EndpointInstruction memory instruction,
-        bytes memory payload
+        address, /* token */
+        uint16, /* recipientChain */
+        uint256, /* deliveryPayment */
+        address, /* caller */
+        EndpointStructs.EndpointInstruction memory, /* instruction */
+        bytes memory /* payload */
```
RahulMaganti47, line 69: same here as well

src/WormholeEndpoint.sol
```diff
@@ -79,6 +84,17 @@ abstract contract WormholeEndpoint is Endpoint, IWormholeEndpoint, IWormholeRece
         }
     }
 
+    function _getSpecialRelayingEnabledChainsStorage()
+        internal
+        pure
+        returns (mapping(uint16 => bool) storage $)
```
RahulMaganti47, line 90: id suggest using the native EVM word size here (uint256) instead of a bool. saves an extra SLOAD when reading from storage later

src/Endpoint.sol
```diff
@@ -6,6 +6,7 @@ import "./libraries/PausableOwnable.sol";
 
 abstract contract Endpoint is PausableOwnable {
     function _sendMessage(
+        address token,
```
djb15, line 9: Why are we extending both of these interfaces to include the token address? Since we have the token address available as an immutable in both the endpoints and the manager I don't think we have to pass it around like this?

nik-suri: The problem is that `token` isn't available in the `WormholeEndpoint` implementation.
We need to find some way to pass this token address to the `WormholeEndpoint`, so I opted for adding the token to the internal `_sendMessage` method.

Perhaps there's another way to do this where we expose a virtual `token()` method in the abstract Endpoint contract that the `WormholeEndpoint` will call, and this'll work when all the contracts are put together?

src/WormholeEndpoint.sol
nik-suri: Can do. Should we also update all the other storage getter to return uint256 instead of bool, such as `_getWormholeRelayingEnabledChainsStorage`?

src/EndpointStandalone.sol
nik-suri: Which is preferable?
Passing the token as a function parameter requires a 2 lookups against the manager contract for every transfer -- 1 lookup during quoting and 1 lookup when sending the message.
Adding the token as a state variable takes up additional state on the Endpoint but enables us to only lookup against the manager once at contract instantiation.

src/Endpoint.sol
djb15: But we're now adding it as an immutable? https://github.com/wormhole-foundation/example-native-token-transfers/blob/quoter/src/EndpointStandalone.sol#L25

And it's already accessible in the Manager so for both cases it's available no?

src/EndpointStandalone.sol
djb15: Fyi I prefer the immutable, discussed in another thread

src/WormholeEndpoint.sol
RahulMaganti47: Yea I think so

test/Manager.t.sol
nik-suri: Discussed offline -- this is just personal preference but I like having the variables commented out on the side for tests, so if someone wants to start using them they easily can just by commenting out and the rest are clearly there. I don't think we really should need to have the `@param` docstrings for the test methods.

But I do agree we should have the @param docstrings generally in the base level interfaces and abstract contracts. That's something to tackle as part of the docstrings issue #24

test/Manager.t.sol
nik-suri: same as above

djb15 (approved):

RahulMaganti47 (approved):

PR #135, opened 2024-02-16
Dependencies fix

djb15:
Annoyingly the default import of the WH solidity SDK is one where the `asBoolUnchecked` method always returns true and was fixed in https://github.com/wormhole-foundation/wormhole-solidity-sdk/commit/d3afb99e3678a97bd769667060c14b26778ccdea.

I've updated the dependency management to allow us to specify specific commits, like how the monorepo does it

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,5 +13,8 @@ docs/
 # Dotenv file
 .env
 
+# Imported libraries
+lib/
+
 # macOS stuff
 .DS_Store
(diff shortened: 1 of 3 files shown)


kcsongor:
good catch -- but why not just bump the submodule?

kcsongor:
like this: https://github.com/wormhole-foundation/example-native-token-transfers/pull/136

djb15:
We need to keep submodules around so that integrators who want to use NTT can easily do so with foundry

PR #140, opened 2024-02-19
Add NttNormalizer abstract contract

nik-suri:
- Shift decimals to match the native token decimals when receiving an NTT transfer.
- Use `NttNormalizer` wrapper lib to enforce all NTT-related normalize/denormalize operations always use the proper decimals.

### DIFF
--- a/src/libraries/NormalizedAmount.sol
+++ b/src/libraries/NormalizedAmount.sol
@@ -132,7 +132,7 @@ library NormalizedAmountLib {
         return a.amount < b.amount ? a : b;
     }
 
-    /// @dev scale the amount from origDecimals to normDecimals (base 10)
+    /// @dev scale the amount from original decimals to target decimals (base 10)
     function scale(
         uint256 amount,
         uint8 fromDecimals,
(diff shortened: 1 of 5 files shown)


src/libraries/NormalizedAmount.sol
```diff
@@ -132,16 +132,18 @@ library NormalizedAmountLib {
...
     ) internal pure returns (uint256) {
         if (fromDecimals > toDecimals) {
             return amount / (10 ** (fromDecimals - toDecimals));
-        } else {
+        } else if (fromDecimals < toDecimals) {
             return amount * (10 ** (toDecimals - fromDecimals));
+        } else {
+            return amount;
         }
     }
```
RahulMaganti47, line 148: it may be a good idea to keep this logic consistent with the solana impl.
https://github.com/wormholelabs-xyz/solana-example-native-token-transfers/blob/ab2ca08b3dd27881a82070ad0e09964aba73d590/solana/programs/example-native-token-transfers/src/normalized_amount.rs#L47

nik-suri: The structure is identical -- does re-ordering have a significant impact on gas or something else?

RahulMaganti47: its more readable.

RahulMaganti47 (approved):

PR #148, opened 2024-02-21
fix: cache the deployer to mitigate front-running on initialization

### LINKED ISSUE #145: Initialize() Owner is Frontrunnable
Currently, the owner of the manager and endpoint is set through whoever calls `initialize()`. Since the deployment of the contracts and call to initialize are separate, an attacker could frontrun this call to become the owner. 

By itself, we would probably just redeploy it after some sort of error occurs and we see that we're no longer the owner. However, I'm worried that there's a nice way to backdoor it (such as adding our own endpoint) then transfer ownership back to the user. 

There's a test demonstrating this under ``TestInitialize`` in ``tests/Upgrades.sol``. I think if we save the deployer then add a check in the call to initialize that the deployer is the caller then this should work fine.

RahulMaganti47:
Fixes #145.

### DIFF
--- a/evm/src/Manager.sol
+++ b/evm/src/Manager.sol
@@ -20,7 +20,6 @@ import "./Endpoint.sol";
 import "./EndpointRegistry.sol";
 import "./NttNormalizer.sol";
 import "./libraries/PausableOwnable.sol";
-import {ContextUpgradeable} from "./libraries/external/ContextUpgradeable.sol";
 
 // TODO: rename this (it's really the business logic)
 abstract contract Manager is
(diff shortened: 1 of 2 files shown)


evm/src/Manager.sol
```diff
@@ -104,9 +107,15 @@ abstract contract Manager is
         mode = _mode;
         chainId = _chainId;
         evmChainId = block.chainid;
+        // save the deployer (check this on iniitialization)
+        deployer = _msgSender();
```
djb15, line 111: Any reason in particular why we're using the Context package here now but still msg.sender everywhere else?

djb15 (approved):

evm/src/Manager.sol
RahulMaganti47: to cover the case in which the deployer is a contract to retrieve the address who initiated the deployment.

djb15: We don't override it anywhere though right so it's still `msg.sender`?

gator-boi: You're correct, I'll make the change.

djb15 (approved):

PR #151, opened 2024-02-21
evm: Wormhole endpoint registration

djb15:
1. Do the messages we're emitting contain all the information we need?
2. Does the payload format with the new prefixes make sense, or would another format be preferable?


djb15:
Cherry picked into new PR with cleaner history

PR #153, opened 2024-02-21
fix: refactor to separate Endpoint and Manager contracts

RahulMaganti47:


evm/test/IntegrationStandalone.t.sol
djb15, line 1: I would like to see these tests pass with the refactor

evm/test/IntegrationRelayer.t.sol
djb15, line 1: Same with these tests

djb15 (commented):
Generally looks sensible to me!

There are quite a few virtual and override methods around still that no longer have to be so I'd suggest cleaning that up.

The biggest thing imo is getting those integration tests working. Shouldn't actually be a huge lift as the entrypoints have all stayed the same, it's just some setup stuff that you'll have to change

evm/src/Endpoint.sol
```diff
@@ -3,8 +3,95 @@ pragma solidity >=0.8.8 <0.9.0;
 
 import "./libraries/EndpointStructs.sol";
 import "./libraries/PausableOwnable.sol";
+import "./interfaces/IManager.sol";
+import "./interfaces/IEndpoint.sol";
+import "./libraries/external/ReentrancyGuardUpgradeable.sol";
+import "./libraries/Implementation.sol";
+
+abstract contract Endpoint is
```
djb15, line 11: Shouldn't be abstract anymore

evm/src/Manager.sol
```diff
@@ -20,6 +20,7 @@ import "./Endpoint.sol";
 import "./EndpointRegistry.sol";
 import "./NttNormalizer.sol";
 import "./libraries/PausableOwnable.sol";
+import "./libraries/Implementation.sol";
 
 // TODO: rename this (it's really the business logic)
 abstract contract Manager is
```
djb15, line 26: Shouldn't be abstract anymore

evm/test/IntegrationStandalone.t.sol
RahulMaganti47: test pass now!

evm/src/Endpoint.sol
djb15: My bad, this one should stay abstract

djb15 (dismissed):
LGTM!

evm/src/Manager.sol
```diff
@@ -191,9 +192,9 @@ contract Manager is
         if (msg.sender != deployer) {
             revert UnexpectedOwner(deployer, msg.sender);
         }
-        // TODO: msg.sender may not be the right address for both
         __PausedOwnable_init(msg.sender, msg.sender);
-        // TODO: check if it's safe to not initialise reentrancy guard
+        // NOTE: the first transfer will be more expensive if this is not done here.
```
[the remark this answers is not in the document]
gator-boi: This comment was removed, thanks.

evm/src/Manager.sol
```diff
@@ -582,7 +718,89 @@ abstract contract Manager is
...
+    ) external onlyEndpoint {
+        _verifySibling(sourceChainId, sourceManagerAddress);
+
+        bytes32 managerMessageHash = EndpointStructs.managerMessageDigest(sourceChainId, payload);
+
+        // set the attested flag for this endpoint.
+        // NOTE: Attestation is idempotent (bitwise or 1), but we revert
+        // anyway to ensure that the client does not continue to initiate calls
+        // to receive the same message through the same endpoint.
+        if (
```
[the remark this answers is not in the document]
gator-boi: `executeMsg` is now a public function, this will serve as a recovery path for when an endpoint is disabled and an attestation has already been delivered by a different endpoint.

evm/src/Manager.sol
```diff
@@ -449,12 +586,11 @@ abstract contract Manager is
 
     /// @dev Called after a message has been sufficiently verified to execute the command in the message.
     ///      This function will decode the payload as an ManagerMessage to extract the sequence, msgType, and other parameters.
-    /// TODO: we could make this public. all the security checks are done here
     function _executeMsg(
```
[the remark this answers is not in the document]
gator-boi: I pushed a fix, thanks.

djb15 (approved):

PR #155, opened 2024-02-21
feat: make the consistency level configurable.

### LINKED ISSUE #154: make consistency level configurable
set in the constructor

RahulMaganti47:
Fixes #154.

### DIFF
--- a/evm/src/WormholeEndpoint.sol
+++ b/evm/src/WormholeEndpoint.sol
@@ -15,7 +15,7 @@ abstract contract WormholeEndpoint is Endpoint, IWormholeEndpoint, IWormholeRece
 
     // TODO -- fix this after some testing
     uint256 public constant GAS_LIMIT = 500000;
-    uint8 public constant CONSISTENCY_LEVEL = 1;
+    uint8 public immutable consistencyLevel;
 
     /// @dev Prefix for all EndpointMessage payloads
     ///      This is 0x99'E''W''H'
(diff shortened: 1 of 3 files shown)


djb15:
Why do we want to make this configurable when it's going to be an immutable anyway? Is there a world where we don't want finality for the WH endpoint?

RahulMaganti47:
> Why do we want to make this configurable when it's going to be an immutable anyway? Is there a world where we don't want finality for the WH endpoint?

there may be integrators who'd like to choose their own consistency level (even if it is immutable - this allows them to make this choice upfront)

evm/test/Upgrades.t.sol
```diff
@@ -41,6 +41,7 @@ contract TestUpgrades is Test, IManagerEvents, IRateLimiterEvents {
         0xcfb12303a19cde580bb4dd771639b0d26bc68353645571a8cff516ab2ee113a0;
     WormholeSimulator guardian;
     uint256 initialBlockTimestamp;
+    uint8 constant FAST_CONSISTENCY_LEVEL = 200;
```
djb15, line 44: Why are we using this consistency level in the tests? Intrigued more than anything

evm/src/WormholeEndpoint.sol
```diff
@@ -15,7 +15,7 @@ abstract contract WormholeEndpoint is Endpoint, IWormholeEndpoint, IWormholeRece
 
     // TODO -- fix this after some testing
     uint256 public constant GAS_LIMIT = 500000;
-    uint8 public constant CONSISTENCY_LEVEL = 1;
+    uint8 public immutable consistencyLevel;
```
djb15, line 18: Can we add a docstring above this then to specify common consistency levels that integrators might want to use and why?

evm/test/Upgrades.t.sol
RahulMaganti47: this is code for instant finality. for testnet @evan-gray mentioned this would better.

PR #172, opened 2024-02-23
DRAFT: Refactor/cut code to hit contract size limit

djb15:


evm/src/Manager.sol
```diff
@@ -764,49 +711,13 @@ contract Manager is
...
-            count++;
-        }
-
-        return count;
-    }
-
     /// ============== INVARIANTS =============================================
 
     /// @dev When we add new immutables, this function should be updated
     function _checkImmutables() internal view override {
```
gator-boi, line 717: We can remove this, it's not called anywhere.

RahulMaganti47 (commented):
can we move the invariant checks to the tests? that way we still have the checks there for development

djb15:
Closing in favour of another PR

PR #190, opened 2024-02-24
foundry_toml: add back prod profile to config

RahulMaganti47:


evm/foundry.toml
```diff
@@ -8,6 +8,16 @@ src = "src"
 out = "out"
 libs = ["lib"]
 
+[profile.prod]
+solc_version = "0.8.19"
+optimizer = true
+optimizer_runs = 200
+via-ir = true
```
kcsongor, line 15: I think we should keep just this line -- the rest are inherited from the default profile

kcsongor (changes requested):

djb15 (approved):

djb15 (changes requested):
via_ir

djb15 (approved):

PR #191, opened 2024-02-24
The great renaming

kcsongor:
- [x] endpoint -> transceiver
- [x] manager -> nttmanager
- [x] sibling -> peer


evm/src/NttManager.sol
```diff
@@ -275,30 +275,33 @@ contract Manager is
...
 
-        emit MessageAttestedTo(digest, endpoint, _getEndpointInfosStorage()[endpoint].index);
+        emit MessageAttestedTo(
+            digest, transceiver, _getTransceiverInfosStorage()[transceiver].index
+        );
     }
 
     /*
-     * @dev pause the Endpoint.
+     * @dev pause the Transceiver.
```
djb15, line 291: Unrelated but this comment is incorrect

RahulMaganti47 (approved):

evm/src/WormholeTransceiver.sol
```diff
@@ -5,58 +5,58 @@ import "wormhole-solidity-sdk/WormholeRelayerSDK.sol";
...
-contract WormholeEndpoint is Endpoint, IWormholeEndpoint, IWormholeReceiver {
+contract WormholeTransceiver is Transceiver, IWormholeTransceiver, IWormholeReceiver {
     using BytesParsing for bytes;
 
     uint256 public constant GAS_LIMIT = 500000;
     uint8 public immutable consistencyLevel;
 
-    /// @dev Prefix for all EndpointMessage payloads
+    /// @dev Prefix for all TransceiverMessage payloads
     ///      This is 0x99'E''W''H'
```
djb15, line 22: More intrusive, but should we change the payload prefix to match the naming change?

djb15: Same goes for all these constants actually, the ones below aren't correct anymore (i.e. the comment does not align with the actual bytes)

kcsongor: good point, let me fix those -- tempted to keep the 'E' 'W' 'H' though. what do you think?

djb15: Yeah that one is maybe worth keeping as is

evm/src/interfaces/INttManagerStandalone.sol
djb15, line 1: I'm assuming we don't need this anymore/should rename?

solana/programs/example-native-token-transfers/src/peer.rs
djb15, line 1: Any reason why this has moved up a directory? Also the `/endpoints/accounts` folder it came from can be renamed

kcsongor: I don't think it moved up -- there is still a `transceivers/accounts/peer.rs` (which is what used to be the "endpoint sibling". this thing is the "manager sibling")

kcsongor: perhaps this file should move into an `accounts` folder

djb15: Ah yeah my bad I got those mixed up. Up to you whether you think it's worth moving

djb15 (approved):
My eyes are bleeding now but I think this looks good

PR #197, opened 2024-02-25
Enable tilt

evan-gray:


.github/workflows/tilt.yml
```diff
@@ -0,0 +1,40 @@
...
+      - uses: actions/checkout@v2
+      - name: Expand for link to Tilt dashboard (only available during build)
+        run: >
+          echo "Tilt progress dashboard: https://$DASHBOARD_URL"
+      - run: |
+          kubectl config set-context ci --namespace=$DEPLOY_NS
+          kubectl config use-context ci
+
+      - run: tilt ci -- --evm2 --namespace=$DEPLOY_NS
+        timeout-minutes: 60
```
djb15, line 36: The default timeout for `tilt ci` is 30 mins btw despite having the longer timeout on the step. Don't think we should need to increase beyond 30 though

djb15 (approved):
Glossed over the bindings but I'm on-board with enabling tilt tests

.github/workflows/tilt.yml
evan-gray: agree, this was copied from the wormhole repo. would rather have `tilt ci` timeout than GH kill the step, I guess :shrug: open to changing in a future PR

johnsaigle:
Looks good to me! I am not a reviewer on the repo though so you'll still need someone else to weigh in 😄 

My only note is that GitHub is flagging the branch as out-of-date. Might be worth a rebase-and-push in case GitHub doesn't allow you to merge because of this?

johnsaigle (approved):

evan-gray:
will follow up with another regen in #229 which also contains some test fixes

PR #198, opened 2024-02-25
normalized -> trimmed amount

kcsongor:
this name makes more sense, given that these amounts are not technically normalized (there's no normal form)


RahulMaganti47 (dismissed):

djb15 (approved):

djb15 (approved):
Lgtm

PR #206, opened 2024-02-26
Take out ntt trimmer

RahulMaganti47:


evm/src/NttManager/NttManager.sol
```diff
@@ -135,8 +137,8 @@ contract NttManager is INttManager, NttManagerState {
         if (nativeTokenTransfer.toChain != chainId) {
             revert InvalidTargetChain(nativeTokenTransfer.toChain, chainId);
         }
-
-        TrimmedAmount memory nativeTransferAmount = _nttFixDecimals(nativeTokenTransfer.amount);
+        TrimmedAmount memory nativeTransferAmount =
+            nativeTokenTransfer.amount.untrim(tokenDecimals()).trim(tokenDecimals());
```
djb15, line 141: This should operate from left to right but can we add an extra set of brackets here to be explicit?

evm/src/NttManager/NttManager.sol
```diff
@@ -214,11 +216,6 @@ contract NttManager is INttManager, NttManagerState {
         );
     }
 
-    /// @inheritdoc INttManager
```
djb15, line 217: I think we should keep this implementation around since this is called from the transceiver. We want the transceiver to fetch the immutable, not call the underlying token again

evm/src/NttManager/NttManager.sol
```diff
@@ -430,6 +431,11 @@ contract NttManager is INttManager, NttManagerState {
         }
     }
 
+    /// @inheritdoc INttManager
+    function tokenDecimals() public view override(INttManager, RateLimiter) returns (uint8) {
```
djb15, line 435: You could just put this inline in the constructor of NttManagerState to set the immutable

evm/src/NttManager/NttManager.sol
RahulMaganti47: sure

djb15 (dismissed):

djb15 (approved):

PR #207, opened 2024-02-26
cap decimal trimming by destination decimals

### LINKED ISSUE #205: Fix decimals issue if scaling down to less than 8

kcsongor:
fixes #205

### DIFF
--- a/evm/src/NttManager/NttManagerState.sol
+++ b/evm/src/NttManager/NttManagerState.sol
@@ -121,7 +121,11 @@ abstract contract NttManagerState is
         }
     }
 
-    function _getPeersStorage() internal pure returns (mapping(uint16 => bytes32) storage $) {
+    function _getPeersStorage()
+        internal
+        pure
+        returns (mapping(uint16 => NttManagerPeer) storage $)
+    {
         uint256 slot = uint256(PEERS_SLOT);
         assembly ("memory-safe") {
             $.slot := slot
@@ -157,7 +161,7 @@ abstract contract NttManagerState is
     }
 
     /// @inheritdoc INttManagerState
-    function getPeer(uint16 chainId_) public view returns (bytes32) {
+    function getPeer(uint16 chainId_) external view returns (NttManagerPeer memory) {
         return _getPeersStorage()[chainId_];
     }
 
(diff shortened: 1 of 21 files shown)


evm/src/interfaces/INttManagerEvents.sol
```diff
@@ -18,11 +18,17 @@ interface INttManagerEvents {
 
     /// @notice Emitted when the peer contract is updated.
     /// @dev Topic0
-    ///      0x51b8437a7e22240c473f4cbdb4ed3a4f4bf5a9e7b3c511d7cfe0197325735700.
+    ///      0x1456404e7f41f35c3daac941bb50bad417a66275c3040061b4287d787719599d.
     /// @param chainId_ The chain ID of the peer contract.
     /// @param oldPeerContract The old peer contract address.
     /// @param peerContract The new peer contract address.
```
djb15, line 24: Update spec

evm/src/NttManager/NttManagerState.sol
```diff
@@ -251,29 +255,35 @@ abstract contract NttManagerState is
...
+        if (decimals == 0) {
+            revert InvalidPeerDecimals();
+        }
 
-        bytes32 oldPeerContract = _getPeersStorage()[peerChainId];
+        NttManagerPeer memory oldPeer = _getPeersStorage()[peerChainId];
 
-        _getPeersStorage()[peerChainId] = peerContract;
+        _getPeersStorage()[peerChainId].peerAddress = peerContract;
+        _getPeersStorage()[peerChainId].tokenDecimals = decimals;
```
djb15, line 272: Maybe we should add a check here that if the previous token decimals were non-zero then the new token decimals should be the same?

I don't think this is a security issue since we check the manager that's encoded in the payload (specifically thinking about in-flight transfers) but just intrigued what you think

kcsongor: I think we can just keep it this way, good flag though

RahulMaganti47, line 272: can you use `setPeer` here instead?

kcsongor: this _is_ `setPeer`

djb15 (approved):

PR #219, opened 2024-02-28
solana: add more tests

kcsongor:


johnsaigle (commented):
More tests are always great! Looks likes this needs a rebase. 
After that I'll do a proper review. I'm sure adding tests will only be beneficial though so expect an approval

kcsongor:
I'll rebase this later today

johnsaigle (changes requested):
Looks like this PR has some conflicts to be resolved

kcsongor:
@johnsaigle rebased!

johnsaigle:
@kcsongor Clippy is unhappy

johnsaigle (approved):

johnsaigle (changes requested):
CI is failing because of existing `panic` calls in programs/example-native-token-transfers/tests/common/setup.rs.

https://github.com/wormhole-foundation/example-native-token-transfers/blob/cb57ed240bc4e986a144c6bc7e6b7e6fb3823cb5/solana/programs/example-native-token-transfers/tests/common/setup.rs#L420-L425

I think this is somehow related to the `rust-toolchain` change? If I change to `stable` the panic is not flagged.

Anyway, I think you can add `#[allow(clippy::panic)]` just above the failing code block and it should resolve the issue

johnsaigle (approved):

johnsaigle:
@kcsongor A couple of notes:

1. I added a commit to the branch: it adds a `bad_mint_authority` to the TestData object to resolve build failures. I'm not sure if this is the right approach so feel free to nuke that commit if it doesn't make sense to you.

2. I'm very confused about the current test failure on `test_receive()`. It's getting an `Ok(())` from `release_inbound` and so `unwrap_err()` panics. However, `Ok(())` should never happen when we call try to release an inbox item twice (as it should be in `ReleaseStatus::Released` and therefore `try_release()` should throw an error).

Interestingly, the test actually passes if `revert_on_delay` is enabled (!)
I'm not sure how this is possible. Hopefully you have some insight to figure this one out.

(I opened #336 in the process of trying to debug this. Take a look, maybe it'll help you)

PR #220, opened 2024-02-28
evm: Fix possible overflow on backfilling

djb15:
This is a prime target for upcoming fuzz testing so I'm not shoehorning a test in with this change

### DIFF
--- a/evm/src/libraries/TrimmedAmount.sol
+++ b/evm/src/libraries/TrimmedAmount.sol
@@ -105,6 +105,31 @@ library TrimmedAmountLib {
         return TrimmedAmount(a.amount + b.amount, a.decimals);
     }
 
+    function saturatingAdd(
+        TrimmedAmount memory a,
+        TrimmedAmount memory b
+    ) internal pure returns (TrimmedAmount memory) {
+        // on initialization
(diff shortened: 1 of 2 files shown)


RahulMaganti47 (approved):

evm/src/libraries/TrimmedAmount.sol
```diff
@@ -105,6 +105,31 @@ library TrimmedAmountLib {
         return TrimmedAmount(a.amount + b.amount, a.decimals);
     }
 
+    function saturatingAdd(
+        TrimmedAmount memory a,
+        TrimmedAmount memory b
+    ) internal pure returns (TrimmedAmount memory) {
+        // on initialization
+        if (isZero(a)) {
```
kcsongor, line 113: we should do something about these (separately). I feel like these checks should be somewhere on the initialisation path

kcsongor (approved):

evm/src/libraries/TrimmedAmount.sol
djb15: Created https://github.com/wormhole-foundation/example-native-token-transfers/issues/222

PR #224, opened 2024-02-28
evm: guarantee that rate limits are initialised

kcsongor:
Previously, the rate limits were not initialised, which was masked by the graceful handling of the trimmed arithmetic operations when they encountered zero values. That handling was wrong, as we should handle the zero case exactly where it's expected: the `setLimit` function (the old values will be zero on init).

Also added outbound limit initialisation to the manager initialiser, and inbound limit initialisation to peer registration. We could consider thaking the inbound limit as an argument when registering a peer instead of the current behaviour which is to set it to the max limit. That's what the solana implementation does.

### DIFF
--- a/evm/src/libraries/TrimmedAmount.sol
+++ b/evm/src/libraries/TrimmedAmount.sol
@@ -49,8 +49,7 @@ library TrimmedAmountLib {
         return a.amount < b.amount;
     }
 
-    // TODO: is this needed? let's remove it
-    function isZero(TrimmedAmount memory a) internal pure returns (bool) {
+    function isNull(TrimmedAmount memory a) internal pure returns (bool) {
         return (a.amount == 0 && a.decimals == 0);
     }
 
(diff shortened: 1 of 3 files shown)


RahulMaganti47 (approved):

evm/src/libraries/TrimmedAmount.sol
djb15, line 1: There are still the `isZero` checks on `saturatingAdd`

djb15 (commented):
I think defaulting to the max outbound limit is the lesser of two evils rather than queuing by default.

I've created https://github.com/wormhole-foundation/example-native-token-transfers/issues/227 as a follow up as I'd prefer it if we didn't have a max default in both directions.

Will approve when the last `isZero` checks are removed

evm/src/libraries/TrimmedAmount.sol
kcsongor: ah, that's a rebase artefact -- let me drop them

djb15 (approved):

RahulMaganti47 (approved):

PR #226, opened 2024-02-28
Fix incorrect bitwise comparison

mdulin2:
There's a bug in the function ``transceiverAttestedToMessage()`` for indexes greater than 0.

Original code - ``_getMessageAttestationsStorage()[digest].attestedTransceivers & uint64(1 << index) == 1;`` 

* The call at the beginning is getting a bitmap of attested transceivers. So, 7 (b111) would be 3 accepted transceivers. 
* Next, we are computing the index to compare on the bitmap with transceiver index. So, if we want check the second index then 1 << 1 would work to produce 2 or the middle bit out of the 3 in the example. 
* Finally, we're comparing this with 1, which is where the bug is at. This is ONLY true if we're checking the first index. For instance, if we're checking the second index (value of 1) then we'd do ``7 & (1 << 1) == 1`` or ``7 & 2 == 1`` which is ``2 == 1`` . So, the comparison will fail on all indexes that aren't the first one (value of 0).

So, the fix is to check if the value is greater than 0. This means that if the singular bit is set, then it's already been attested to. 

Not a huge impact, since there's no running counter (we count the bits on a given digest every time) and the replay protection is somewhere else but still worth fixing in case changes are made.

### DIFF
--- a/evm/src/NttManager/NttManagerState.sol
+++ b/evm/src/NttManager/NttManagerState.sol
@@ -170,7 +170,7 @@ abstract contract NttManagerState is
     /// @inheritdoc INttManagerState
     function transceiverAttestedToMessage(bytes32 digest, uint8 index) public view returns (bool) {
         return
-            _getMessageAttestationsStorage()[digest].attestedTransceivers & uint64(1 << index) == 1;
+            _getMessageAttestationsStorage()[digest].attestedTransceivers & uint64(1 << index) > 0;
     }
 
     /// @inheritdoc INttManagerState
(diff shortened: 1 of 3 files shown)


mdulin2:
Need to fix a few tests before this is merged.

mdulin2:
Fixed the two failing tests that were seeing the wrong error for replay protections because of the change.

Now, one of them checks the replay protection for ``execMsg()`` and the other one checks the replay check for ``attestationReceived``. Having a basic test for both checks should be good for the future if changes are made.

djb15 (approved):

kcsongor (approved):
good catch -- agree there's minimal impact currently as the entire code path is redundant, but it might be used for something else in the future

PR #228, opened 2024-02-29
sequence -> id and make it 32 bytes

kcsongor:
The key motivation behind this change is to make it possible to look up the outbox item on solana from a given id. The incrementally assigned sequence numbers don't work for this, because seeding PDAs by them would create race conditions. By putting the outbox item's account address directly into the payload, this makes it possible to perform those lookups.


evm/src/libraries/TransceiverStructs.sol
```diff
@@ -26,13 +26,13 @@ library TransceiverStructs {
 
     /// @dev Message emitted and received by the nttManager contract.
     ///      The wire format is as follows:
-    ///      - sequence - 8 bytes
+    ///      - id - 32 bytes
     ///      - sender - 32 bytes
     ///      - payloadLength - 2 bytes
     ///      - payload - `payloadLength` bytes
     struct NttManagerMessage {
         /// @notice unique sequence number
```
bingyuyap, line 34: do we still call this a unique _sequence_ number? since on solana side it's not incremental

kcsongor: good point, let me fix

evm/src/NttManager/NttManager.sol
```diff
@@ -393,7 +393,9 @@ contract NttManager is INttManager, NttManagerState {
         // construct the NttManagerMessage payload
         bytes memory encodedNttManagerPayload = TransceiverStructs.encodeNttManagerMessage(
             TransceiverStructs.NttManagerMessage(
-                seq, toWormholeFormat(sender), TransceiverStructs.encodeNativeTokenTransfer(ntt)
+                bytes32(uint256(seq)),
```
djb15, line 396: Any particular reason why you decided to cast to uint256 here rather then changing the counter to a uint256? No strong feelings either way personally

djb15 (approved):

evm/src/NttManager/NttManager.sol
kcsongor: I suppose it's in its own storage slot anyway. would it be crazy to consider packing the sequence counter into the same slot as the outbound rate limit params? they are always accessed together, so would be a nontrivial gas saving.

djb15: Nice suggestion actually, that could be a follow up PR. Should be a pretty low risk change as gas optimisations go

RahulMaganti47 (approved):

PR #229, opened 2024-02-29
ci: evm e2e with and without relaying

evan-gray:
this regens `evm_bindings` and adds @mdulin2 's e2e evm tests with and without relayers. then cleans up that script a bit. there may still be some race condition with the set up that I've yet to nail down, but should be a positive step forward.


johnsaigle (approved):
Looks good. The TypeScript code is more readable and happy to see the TODOs done and the comments addressed.
I defer to @mdulin2 ultimately since he wrote most of the code

PR #234, opened 2024-02-29
solana: Add comments and linting directives in governance program

johnsaigle:
- Add function documentation and some clarifying notes about how the governance program works
- Add linting directives for cargo dylint to ignore ownership checks in a particular case

See also: https://github.com/wormhole-foundation/example-native-token-transfers/pull/232 for context on the clippy annotations


johnsaigle:
@kcsongor I would love your input here regarding the accuracy of my comments and the security concerns about the ownership checks

solana/programs/wormhole-governance/src/instructions/governance.rs
```diff
@@ -9,7 +9,7 @@
 //! instruction is able to invoke the program's admin instructions.
 //!
 //! The instruction needs to be encoded in the VAA payload, with all the
-//! accounts. These accounts may be in any order, with two placeholder accounts:
+//! accounts. These accounts may be in any order and must include two additional placeholder accounts:
```
kcsongor, line 12: these are actually optional, so maybe 
```suggestion
//! accounts. These accounts may be in any order and may include two additional placeholder accounts:
```

kcsongor (commented):
great!

solana/programs/wormhole-governance/src/instructions/governance.rs
```diff
@@ -125,15 +124,32 @@ impl From<AccountMeta> for Acc {
...
+/// placeholder accounts with Pubkeys set to hard-coded values [OWNER] and [PAYER].
+/// These keys are overwritten. Because they are placeholders, the `owner` field is not important.
+#[allow(unknown_lints)]
+#[allow(missing_owner_check)]
 pub fn governance<'info>(ctx: Context<'_, '_, '_, 'info, Governance<'info>>) -> Result<()> {
     let vaa_data = ctx.accounts.vaa.data();
 
     let mut instruction: Instruction = vaa_data.clone().into();
 
+    // Iterate over a copy of all accounts provided in the VAA payload.
```
johnsaigle, line 141: I can remove this comment if it's too much. It helped me to reason about the code but may be redundant for someone who is more familiar with Anchor.

johnsaigle:
@kcsongor Could you re-review when you get some time? This resolves all of the false positives.

solana/programs/wormhole-governance/src/instructions/governance.rs
kcsongor: I like this comment here

kcsongor:
I like the comments, but lints themselves seem redundant. Looks like it flags false positives for things that anchor checks. How about we drop the linter and just keep the comments? The convention is moving those account security related comments above the account like
```rust
/// CHECK: <comment here>
pub account_foo: Account<...>
```

johnsaigle:
@kcsongor Did you change your mind from what we talked about in other PR? https://github.com/wormhole-foundation/example-native-token-transfers/pull/232#issuecomment-1971689527 😅 

Ultimately, I think you're right. After working with them a bunch I think these lints are meant for Solana itself more than Anchor. I can rework this PR so that contains only the function comments and the security notes formatted as `CHECK:` statements

kcsongor:
Apologies for flip-flopping on this! As you worked out all the false positives I think the picture it paints is that it's not as useful as it first seemed.

solana/programs/example-native-token-transfers/src/instructions/admin.rs
```diff
@@ -204,6 +207,8 @@ pub struct RegisterTransceiver<'info> {
     pub payer: Signer<'info>,
 
     #[account(executable)]
+    // CHECK: Missing ownership check is OK here: Transceiver is intended to be an arbitrary
```
johnsaigle, line 210: @kcsongor I'm actually unsure about this annotation. Do we want to constrain the Transceiver in some way?
This is a privileged action so it's probably OK.

johnsaigle:
I duplicated a lot of this effort in #330 but some of the comments in this PR are still useful. #330 should probably be merged first and I can rebase this PR on that one

nvsriram:
Changes merged through #635

PR #238, opened 2024-02-29
evm: Always fetch token decimals rather than caching

djb15:
Previously we cached the token decimals to save gas. However if the token linked to an NttManager was an upgradeable contract and the number of decimals of the token changed, then we'd be working with an out-of-date decimal amount and minting/unlocking orders of magnitude too many/few tokens.

This change forces us to always fetch the actual token decimals from the token contract when needed. If the decimals of the token do change, then all transfers will fail due to the decimal check on the rate limiting logic (assuming rate limiting is enabled and until the rate limiting configuration is updated). In the case where rate limiting is disabled, the impact is that dust is lost when trimming from the source chain until the peer configuration is updated.

The failure case is recoverable with a contract upgrade that explicitly changes some of the rate limiter decimal checks, but this isn't a pattern we will explicitly support (changing token decimals on the fly)

### DIFF
--- a/evm/src/NttManager/NttManager.sol
+++ b/evm/src/NttManager/NttManager.sol
@@ -155,8 +155,9 @@ contract NttManager is INttManager, NttManagerState {
         if (nativeTokenTransfer.toChain != chainId) {
             revert InvalidTargetChain(nativeTokenTransfer.toChain, chainId);
         }
-        TrimmedAmount nativeTransferAmount =
-            (nativeTokenTransfer.amount.untrim(tokenDecimals_)).trim(tokenDecimals_, tokenDecimals_);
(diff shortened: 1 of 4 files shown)


RahulMaganti47 (dismissed):

gator-boi:
I'm happy to approve this PR (the changes look good) but I think we should add a test that demonstrates outbound and inbound transfers failing when the token decimals are changed.

djb15:
> I'm happy to approve this PR (the changes look good) but I think we should add a test that demonstrates outbound and inbound transfers failing when the token decimals are changed.

Yeah good call, I'll add some tests to validate it's failing like we expect

kcsongor (approved):

RahulMaganti47 (approved):

PR #239, opened 2024-02-29
evm: add fuzz tests

### LINKED ISSUE #174: add fuzz tests

RahulMaganti47:
Fixes #174.

### DIFF
--- a/evm/test/NttManager.t.sol
+++ b/evm/test/NttManager.t.sol
@@ -72,12 +72,20 @@ contract TestNttManager is Test, IRateLimiterEvents {
 
     // === pure unit tests
 
-    function test_countSetBits() public {
-        assertEq(countSetBits(5), 2);
-        assertEq(countSetBits(0), 0);
-        assertEq(countSetBits(15), 4);
-        assertEq(countSetBits(16), 1);
-        assertEq(countSetBits(65535), 16);
+    // naive implementation of countSetBits to test against
+    function simpleCount(uint64 n) public returns (uint8) {
+        uint8 count;
+
+        while (n > 0) {
+            count += uint8(n & 1);
+            n >>= 1;
+        }
+
+        return count;
+    }
+
+    function testFuzz_countSetBits(uint64 n) public {
+        assertEq(simpleCount(n), countSetBits(n));
     }
 
     // === Deployments with rate limiter disabled
--- a/evm/test/NttManager.t.sol
+++ b/evm/test/NttManager.t.sol
@@ -73,7 +73,7 @@ contract TestNttManager is Test, IRateLimiterEvents {
     // === pure unit tests
 
     // naive implementation of countSetBits to test against
-    function simpleCount(uint64 n) public returns (uint8) {
+    function simpleCount(uint64 n) public pure returns (uint8) {
         uint8 count;
 
         while (n > 0) {
--- a/evm/test/RateLimit.t.sol
+++ b/evm/test/RateLimit.t.sol
@@ -444,6 +444,70 @@ contract TestRateLimit is Test, IRateLimiterEvents {
         nttManager.completeOutboundQueuedTransfer(0);
     }
 
+    function testFuzz_outboundRateLimitShouldQueue(uint256 limitAmt, uint256 transferAmt) public {
+        // setup
+        address user_A = address(0x123);
+        address user_B = address(0x456);
+        DummyToken token = DummyToken(nttManager.token());
+        uint8 decimals = token.decimals();
+
+        // inputs
+        uint256 totalAmt = (type(uint64).max) / (10 ** decimals);
+        // avoids the ZeroAmount() error
+        // cannot transfer more than what's available
+        vm.assume(transferAmt > 0 && transferAmt <= totalAmt);
+        // this ensures that the transfer is always queued up
+        vm.assume(limitAmt < transferAmt);
+
+        // mint
+        token.mintDummy(address(user_A), totalAmt * (10 ** decimals));
+        uint256 outboundLimit = limitAmt * (10 ** decimals);
+        nttManager.setOutboundLimit(outboundLimit);
+
+        vm.startPrank(user_A);
+
+        // initiate transfer
+        uint256 transferAmount = transferAmt * (10 ** decimals);
+        token.approve(address(nttManager), transferAmount);
+
+        // shouldQueue == true
+        uint64 qSeq = nttManager.transfer(
+            transferAmount, chainId, toWormholeFormat(user_B), true, new bytes(1)
+        );
+
+        // assert that the transfer got queued up
+        assertEq(qSeq, 0);
+        IRateLimiter.OutboundQueuedTransfer memory qt = nttManager.getOutboundQueuedTransfer(0);
+        assertEq(qt.amount.getAmount(), transferAmount.trim(decimals, decimals).getAmount());
+        assertEq(qt.recipientChain, chainId);
+        assertEq(qt.recipient, toWormholeFormat(user_B));
+        assertEq(qt.txTimestamp, initialBlockTimestamp);
+
+        // assert that the contract also locked funds from the user
+        assertEq(token.balanceOf(address(user_A)), totalAmt * (10 ** decimals) - transferAmount);
+        assertEq(token.balanceOf(address(nttManager)), transferAmount);
+
+        // elapse rate limit duration - 1
+        uint256 durationElapsedTime = initialBlockTimestamp + nttManager.rateLimitDuration();
+        vm.warp(durationElapsedTime - 1);
+
+        // assert that transfer still can't be completed
+        bytes4 stillQueuedSelector =
+            bytes4(keccak256("OutboundQueuedTransferStillQueued(uint64,uint256)"));
+        vm.expectRevert(abi.encodeWithSelector(stillQueuedSelector, 0, initialBlockTimestamp));
+        nttManager.completeOutboundQueuedTransfer(0);
+
+        // now complete transfer
+        vm.warp(durationElapsedTime);
+        uint64 seq = nttManager.completeOutboundQueuedTransfer(0);
+        assertEq(seq, 0);
+
+        // now ensure transfer was removed from queue
+        bytes4 notFoundSelector = bytes4(keccak256("OutboundQueuedTransferNotFound(uint64)"));
+        vm.expectRevert(abi.encodeWithSelector(notFoundSelector, 0));
+        nttManager.completeOutboundQueuedTransfer(0);
+    }
+
     function test_inboundRateLimit_simple() public {
         address user_B = address(0x456);
 
--- a/evm/test/RateLimit.t.sol
+++ b/evm/test/RateLimit.t.sol
@@ -738,8 +738,8 @@ contract TestRateLimit is Test, IRateLimiterEvents {
     // send tokens from user_A to user_B
     // this should consume capacity on the outbound side
     // and backfill the inbound side
-    function testFuzz_CircularFlowBackFilling(uint64 mintAmt, uint64 transferAmt) public {
-        mintAmt = uint64(bound(mintAmt, 2, type(uint64).max));
+    function testFuzz_CircularFlowBackFilling(uint64 mintAmt, uint256 transferAmt) public {
+        mintAmt = uint64(bound(mintAmt, 2, type(uint256).max));
         transferAmt = uint64(bound(transferAmt, 1, mintAmt - 1));
 
         (address user_A, address user_B, DummyToken token, uint8 decimals) = setupToken();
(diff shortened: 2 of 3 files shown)


evm/test/RateLimit.t.sol
```diff
@@ -698,4 +701,321 @@ contract TestRateLimit is Test, IRateLimiterEvents {
...
+
+    // transfer tokens from user_A to user_B
+    // this consumes capacity on the outbound side
+    // send tokens from user_B to user_A
+    // this consumes capacity on the inbound side
+    // send tokens from user_A to user_B
+    // this should consume capacity on the outbound side
+    // and backfill the inbound side
+    function testFuzz_CircularFlowBackFilling(uint64 mintAmt, uint64 transferAmt) public {
+        vm.assume(transferAmt > 0 && transferAmt < mintAmt);
```
djb15, line 737: Rather than using `assume` you should try to use `bound` when working with uints. Rather than discarding test cases it just bounds the inputs

evm/test/RateLimit.t.sol
```diff
@@ -698,4 +701,321 @@ contract TestRateLimit is Test, IRateLimiterEvents {
...
+        nttManager.setOutboundLimit(mintAmount.untrim(decimals));
+
+        // transfer 10 tokens
+        vm.startPrank(user_A);
+
+        TrimmedAmount memory transferAmount = TrimmedAmount(transferAmt, 8);
+        token.approve(address(nttManager), type(uint256).max);
+        // transfer tokens from user_A -> user_B via the nttManager
+        nttManager.transfer(
+            transferAmount.untrim(decimals), chainId, toWormholeFormat(user_B), false, new bytes(1)
```
djb15, line 752: Why did you choose to use the `TrimmedAmount` struct to hold these amounts? Doesn't that mean we'll never have an amount > `type(uint64).max` as a `transfer` input?

evm/test/TrimmedAmount.t.sol
```diff
@@ -156,4 +156,112 @@ contract TrimmingTest is Test {
...
+    //            (x <= type(uint64).max, y <= z)
+    //                    => (x.trim(x, 8).untrim(y) == x)
+    function testFuzz_trimIsLeftInverse(
+        uint256 amount,
+        uint8 fromDecimals,
+        uint8 toDecimals
+    ) public {
+        // this is guaranteed by trimming
+        uint256 amt = bound(amount, 1, type(uint64).max);
+        vm.assume(fromDecimals <= 18);
```
djb15, line 170: Could we fuzz with more decimals than this? Think there are some tokens with 24 decimals for example

evm/test/TrimmedAmount.t.sol
```diff
@@ -156,4 +156,112 @@ contract TrimmingTest is Test {
...
+        TrimmedAmount memory b
+    ) public {
+        vm.assume(a.decimals == b.decimals);
+
+        // this is guaranteed by trimming
+        a.amount = uint64(bound(a.amount, 0, type(uint64).max));
+        a.decimals = uint8(bound(a.decimals, 0, 8));
+
+        b.amount = uint64(bound(b.amount, 0, type(uint64).max));
+        b.decimals = uint8(bound(b.decimals, 0, 8));
```
djb15, line 198: Why only bounding to 8 decimals here and don't think there's any need to bound the amounts since it's already a uint64?

evm/test/RateLimit.t.sol
RahulMaganti47: i think this is fine for this test (its not really meant to test the queueing logic). can test that more explicitly in a separates test. the amount transferred should be less than the outbound limit. `setOutboundLimit` trims the `uint256` value, which if greater than u64MAX, reverts on `AmountTooLarge.` using trimmed amounts makes the inputs more explicit. to improve this slightly, we could uint256 values as inputs and then trim (and check the revert case).

evm/test/TrimmedAmount.t.sol
RahulMaganti47: yea i tested with a few more decimals. should we add an error if the decimal amount is too large (say 100?). the division becomes very numerically unstable, creating underflow issues.

evm/test/RateLimit.t.sol
```diff
@@ -639,7 +641,189 @@ contract TestRateLimit is Test, IRateLimiterEvents {
...
+
+        // TODO: also fuzz the fromDecimals?
+
+        // allow for amounts greater than uint64 to check if [`transfer`] reverts
+        // on amounts greater than u64 MAX.
+        TrimmedAmount memory transferAmount = transferAmt.trim(decimals, 8);
+
+        // check error conditions
+        if (transferAmount.amount == 0) {
+            vm.expectRevert();
```
gator-boi, line 760: Why don't we use the actual revert signature?

djb15: Also btw @RahulMaganti47 I prefer using `abiEncodeWithSelector` rather than `abiEncodeWithSignature` as then you don't have to copy the error signature string around. Calling that out as I noticed in the `setIsWormholeEvmChain` PR

RahulMaganti47: ah nice, yea thats cleaner. 

yea we should use the revert signature (it hasn't a standard that we've enforced). there are a few places, esp. in the integration tests where the revert signatures are omitted. i can address those in a subsequent PR

evm/test/RateLimit.t.sol
djb15: Yeah so my thinking around this wasn't for testing the queuing logic (I agree we don't need to test that here), but more that the trimming works properly and reverts when we expect.

Rather than branching in this fuzz test we could have another fuzz test that always reverts for an expected bounded set of input values?

evm/test/TrimmedAmount.t.sol
djb15: I can't imagine that would ever be a realistic scenario right as so many other protocols would break if a token appeared with a crazy number of decimals! So I think going up to 50 is more than enough coverage, thanks

evm/test/RateLimit.t.sol
RahulMaganti47: yea makes sense to include some more negative tests - will add that in

evm/test/TrimmedAmount.t.sol
```diff
@@ -218,4 +220,106 @@ contract TrimmingTest is Test {
...
+        uint256 amount,
+        uint8 fromDecimals,
+        uint8 toDecimals
+    ) public {
+        uint256 amt = bound(amount, 1, type(uint64).max);
+        vm.assume(fromDecimals <= 50);
+        vm.assume(toDecimals <= 50);
+
+        // NOTE: this is guaranteeed by trimming
+        vm.assume(fromDecimals <= 8 && fromDecimals <= toDecimals);
```
johnsaigle, line 237: Can you explain the guarantee here? I'm not finding the relevant code in the TrimmedAmount.sol file (and I don't see it on the Rust side either).

I think it is probably best to remove this check so that the fuzzer can actually guarantee that checks like these are performed.

johnsaigle: As written I believe the fuzzer misses this code path:

https://github.com/wormhole-foundation/example-native-token-transfers/blob/49d5ab0db6ca162e94e757e3e8aec0747ad78b9c/evm/src/libraries/TrimmedAmount.sol#L140-L142

johnsaigle:
I wrote fuzz tests for Trimmed Amount here: https://github.com/wormhole-foundation/example-native-token-transfers/pull/269/files#diff-a7a75ec27a39e2e8f3e35eb187347d1adc9eaece37a5abacd6cb2113551a88f6

It would be great if we could keep enforce the same invariants on both sides

evm/test/TrimmedAmount.t.sol
RahulMaganti47: i don't believe it does? `toDecimals` and `fromDecimals` are flipped here: https://github.com/wormhole-foundation/example-native-token-transfers/blob/92e9f3bc5e9bc26d9d5ab463aa4777a585426401/evm/test/TrimmedAmount.t.sol#L244

johnsaigle: What's the reason for flipping them? The `trim` function should take `fromDecimals` as the first parameter.

Also if they're flipped, then isn't the code missing the inverse branch?

Generally I'm not clear why the two decimal amounts are being constrained like this

RahulMaganti47: it still goes through the other code path on untrimming but can simplify this without restricting the input

RahulMaganti47:
> I wrote fuzz tests for Trimmed Amount here: https://github.com/wormhole-foundation/example-native-token-transfers/pull/269/files#diff-a7a75ec27a39e2e8f3e35eb187347d1adc9eaece37a5abacd6cb2113551a88f6
> 
> It would be great if we could keep enforce the same invariants on both sides

sure - i've updated the existing invariant tests and added in a few more

evm/test/RateLimit.t.sol
```diff
@@ -703,4 +720,369 @@ contract TestRateLimit is Test, IRateLimiterEvents {
...
+        address user_A = address(0x123);
+        address user_B = address(0x456);
+        DummyToken token = DummyToken(nttManager.token());
+        uint8 decimals = token.decimals();
+
+        // inputs
+        uint256 totalAmt = (type(uint64).max) / (10 ** decimals);
+        // avoids the ZeroAmount() error
+        // cannot transfer more than what's available
+        vm.assume(transferAmt > 0 && transferAmt <= totalAmt);
```
gator-boi, line 921: Use `bound` here?

evm/test/TrimmedAmount.t.sol
```diff
@@ -218,4 +220,153 @@ contract TrimmingTest is Test {
...
+        amount2 = bound(amount, 0, type(uint64).max);
+        vm.assume(fromDecimals <= 50);
+        vm.assume(toDecimals <= 50);
+
+        TrimmedAmount trimmedAmt = amount.trim(fromDecimals, toDecimals);
+        TrimmedAmount trimmedAmt2 = amount2.trim(fromDecimals, toDecimals);
+        uint256 untrimmedAmt = trimmedAmt.untrim(fromDecimals);
+        uint256 untrimmedAmt2 = trimmedAmt2.untrim(fromDecimals);
+
+        // trimming is the left inverse of trimming
```
gator-boi, line 319: bad comment?

gator-boi (approved):

evm/test/TrimmedAmount.t.sol
RahulMaganti47: no that's right

gator-boi (approved):

evm/test/NttManager.t.sol
```diff
@@ -72,12 +72,20 @@ contract TestNttManager is Test, IRateLimiterEvents {
...
     // === pure unit tests
 
-    function test_countSetBits() public {
-        assertEq(countSetBits(5), 2);
-        assertEq(countSetBits(0), 0);
-        assertEq(countSetBits(15), 4);
-        assertEq(countSetBits(16), 1);
-        assertEq(countSetBits(65535), 16);
+    // naive implementation of countSetBits to test against
+    function simpleCount(uint64 n) public pure returns (uint8) {
```
mdulin2, line 76: For what it's worth, I tested the 'countSetBits()' in Python using the implementation and feel pretty good about it as well.

evm/test/RateLimit.t.sol
```diff
@@ -703,4 +720,369 @@ contract TestRateLimit is Test, IRateLimiterEvents {
...
+
+        console.log("result: %s", result.length);
+        // // compare revert strings
+        bytes32 expectedRevertHash = keccak256(abi.encode(expectedRevert));
+        (bytes memory res,) = result.slice(4, result.length - 4);
+        bytes32 actualRevertHash = keccak256(abi.encodePacked(res));
+        require(expectedRevertHash == actualRevertHash, "call did not revert as expected");
+    }
+
+    // transfer tokens from user_A to user_B
```
mdulin2, line 764: Good looking test!

evm/test/RateLimit.t.sol
```diff
@@ -703,4 +720,369 @@ contract TestRateLimit is Test, IRateLimiterEvents {
...
+    function testFuzz_CircularFlowBackFilling(uint256 mintAmt, uint256 transferAmt) public {
+        mintAmt = bound(mintAmt, 1, type(uint256).max);
+        // enforces transferAmt <= mintAmt
+        transferAmt = bound(transferAmt, 0, mintAmt);
+
+        (address user_A, address user_B, DummyToken token, uint8 decimals) = setupToken();
+
+        // allow for amounts greater than uint64 to check if [`setOutboundLimit`] reverts
+        // on amounts greater than u64 MAX.
+        if (mintAmt.scale(decimals, 8) > type(uint64).max) {
```
mdulin2, line 780: Will this kill a bunch of runs of fuzzing?

mdulin2 (approved):

evm/test/RateLimit.t.sol
RahulMaganti47: this case should revert for sure anyway

evm/test/NttManager.t.sol
RahulMaganti47: tested just in case

evm/test/RateLimit.t.sol
RahulMaganti47: this case should revert in anyway

gator-boi (approved):

PR #240, opened 2024-02-29
evm: Add back rate limiter view

djb15:
This was previously removed to reduce the contract code size but being added back specifically as the max amount is useful  to know. We're getting pretty tight on the contract size again but it just about fits


gator-boi (approved):
lgtm

RahulMaganti47 (approved):

PR #241, opened 2024-03-01
solana: enforce no fees have been charged

kcsongor:
token2022 supports tokens with fees. The `transfer_checked_with_fee` function that's referenced in many places in the toke2022 documentation doesn't seem to actually exist in the sdk (or the token2022 program), and there are no helpers to perform fee calculation, so for now I've just done the defensive step of enforcing no fees by checking the transferred amount.
We could support this case once there's appropriate sdk support


johnsaigle (approved):

a5-pickle (approved):
Would be cool to handle tokens with fees in the future :smiling_face_with_tear:

PR #246, opened 2024-03-01
evm+solana: Add comments to clarify the order of struct parsing

johnsaigle:
The 'wire format' of the Transceiver messages encodes/decodes the fields of the struct in reverse order to how they are declared in the Solidity and Rust structs.
This is done safely, and we have unit tests to verify that the encoding/decoding process is consistent on both sides. However, it's not clear that this is intentional when reading through the code. This comment should help with readability and hopefully will stop reviewers from falling down rabbit holes.


johnsaigle:
Thanks @djb15. I'll rebase after #248 is merged and add documentation there where appropriate

johnsaigle:
Rebase done, should be up-to-date with `origin/main` and the changes to `TrimmedAmount`

djb15 (approved):

mdulin2 (approved):

PR #247, opened 2024-03-03
readme: add Solana message lifecycle

bingyuyap:


README.md
```diff
@@ -64,6 +64,73 @@ emit MessageAlreadyExecuted(sourceManagerAddress, digest);
...
+
+The following will be produced in the program logs:
+
+```
+Program log: Instruction: ReleaseOutbound
+```
+
+4. Receive
+
+Similar to EVM, [`receive_wormhole_message`] is the entrypoint for receiving a message on Solana. It takes a message that has been relayed from another blockchain stores it in the `transceiver_message` account.
```
nik-suri, line 104: Similar to EVM, Transceivers vary in how they receive messages, since message relaying and verification methods may differ between implementations.

The Wormhole Transceiver receives a verified Wormhole message on Solana via the [`receive_message`] entrypoint instruction. Callers can use the [`receive_wormhole_message`] Anchor library function to execute this instruction. The instruction verifies the Wormhole VAA and stores it in a `VerifiedTransceiverMessage` account.

nik-suri (commented):
Nice work! Left comments for rephrasing certain sections.

Still need to review the final couple receive steps -- `redeem` and `Mint or unlock`

README.md
```diff
@@ -64,6 +64,73 @@ emit MessageAlreadyExecuted(sourceManagerAddress, digest);
 ``` solidity
 emit TransferRedeemed(digest);
 ```
+
+
+### Solana
+
+1. Sending
+
+A client calls the [`transfer_lock`] or [`transfer_burn`] instruction depending on the mode the program was in. The client must specify the amount of the transfer, the recipient chain, the recipient address on the recipient chain, and the boolean flag `should_queue` to specify whether the if the transfer should be queued if there is a rate limit.
```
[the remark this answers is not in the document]
bingyuyap: Updated

README.md
```diff
@@ -64,6 +64,73 @@ emit MessageAlreadyExecuted(sourceManagerAddress, digest);
...
+Program log: Instruction: TransferBurn
+```
+
+The transfer is also added into an Outbox via the `insert_into_outbox` method, where the rate limit is checked against. An Outbox is a Solana Account which holds details of the outbound messages.
+
+2. Rate Limiting
+
+When the `consume_or_delay` function is invoked by the transfer instruction, the program checks if the transfer amount fits within the current capacity. If it does, the outbound capacity is reduced by the transfer amount and the transfer proceeds immediately. Inbound capacity is also refilled by the same amount. If not, the transfer is delayed by 24 hours.
+
+> When `revert_on_delay` is true, the transaction will revert if the release timestamp has not been reached. When `revert_on_delay` is false, the transaction succeeds, but the outbound release is not performed.
```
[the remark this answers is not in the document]
bingyuyap: Updated

README.md
```diff
@@ -64,6 +64,73 @@ emit MessageAlreadyExecuted(sourceManagerAddress, digest);
...
+> Using the wrong transfer instruction, i.e. [`transfer_lock`] for a program that is in "burning" mode, will result in `InvalidMode` error.
+
+Depending on the mode and instruction, the following will be produced in the program logs:
+
+```
+Program log: Instruction: TransferLock
+Program log: Instruction: TransferBurn
+```
+
+The transfer is also added into an Outbox via the `insert_into_outbox` method, where the rate limit is checked against. An Outbox is a Solana Account which holds details of the outbound messages.
```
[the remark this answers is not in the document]
bingyuyap: Updated

README.md
bingyuyap: Updated

README.md
```diff
@@ -21,49 +22,138 @@ The payload includes the trimmed amount, together with the decimals that trimmed
...
+
+```solidity
 emit TransferRedeemed(digest);
 ```
+
+### Solana
+
+1. Sending
+
+A client calls the [transfer_lock] or [transfer_burn] instruction based on whether the program is in "locking" or "burning" mode. The program mode is set during initialization. When transferring, the client must specify the amount of the transfer, the recipient chain, the recipient address on the recipient chain, and the boolean flag `should_queue` to specify whether the transfer should be queued if it hits the outbound rate limit.
```
kcsongor, line 82:
```suggestion
A client calls the [transfer_lock] or [transfer_burn] instruction based on whether the program is in "locking" or "burning" mode. The program mode is set during initialization. When transferring, the client must specify the amount of the transfer, the recipient chain, the recipient address on the recipient chain, and the boolean flag `should_queue` to specify whether the transfer should be queued if it hits the outbound rate limit. If `should_queue` is set to false, the transfer reverts instead of queuing if the rate limit were to be hit.
```

README.md
```diff
@@ -21,49 +22,138 @@ The payload includes the trimmed amount, together with the decimals that trimmed
...
+> Using the wrong transfer instruction, i.e. [`transfer_lock`] for a program that is in "burning" mode, will result in `InvalidMode` error.
+
+Depending on the mode and instruction, the following will be produced in the program logs:
+
+```
+Program log: Instruction: TransferLock
+Program log: Instruction: TransferBurn
+```
+
+Outbound transfers are always added into an Outbox via the `insert_into_outbox` method. This method checks the transfer against the configured outbound rate limit amount to determine whether the transfer should be rate limited. The Outbox is a Solana Account which holds details of the outbound transfer. If no rate limit is hit, the transfer can be released from the Outbox immediately. If a rate limit is hit, the transfer can only be released from the Outbox after the rate limit delay duration has expired.
```
kcsongor, line 93:
```suggestion
Outbound transfers are always added into an Outbox via the `insert_into_outbox` method. This method checks the transfer against the configured outbound rate limit amount to determine whether the transfer should be rate limited. An OutboxItem is a Solana Account which holds details of the outbound transfer. If no rate limit is hit, the transfer can be released from the Outbox immediately. If a rate limit is hit, the transfer can only be released from the Outbox after the rate limit delay duration has expired.
```

kcsongor (dismissed):
Very nice!

kcsongor (approved):

RahulMaganti47 (approved):

PR #248, opened 2024-03-03
evm: modify trimmed amt struct -> uint72

### LINKED ISSUE #231: use user-defined value type for trimmed amount representation

RahulMaganti47:
Fixes #231. 

This PR modifies the current TrimmedAmount struct to a bit-packed representation of a token amount and its decimals.   
```
/// @dev 64 bytes: [0 - 64] amount
/// @dev 8 bytes: [64 - 72] decimals
type TrimmedAmount is uint72; 
```  
This representation is advantageous for two reasons: (1) Less runtime overhead (since structs point to memory, calldata, or storage). Using a user-defined value type allows the trimmed amounts to be stack-allocated; (2) the ability to overload the traditional arithmetic binary operators (+/-/*, etc..) for operations on trimmed amounts.

The use of `uint72` instead of larger unsigned integer types ensures that extra words are not allocated downstream for other data types that contain `TrimmedAmounts` (e.g. `RateLimitParams`)

### DIFF
--- a/evm/src/libraries/TrimmedAmount.sol
+++ b/evm/src/libraries/TrimmedAmount.sol
@@ -2,103 +2,127 @@
 /// @dev TrimmedAmount is a utility library to handle token amounts with different decimals
 pragma solidity >=0.8.8 <0.9.0;
 
-struct TrimmedAmount {
-    uint64 amount;
-    uint8 decimals;
-}
+/// @dev TrimmedAmount is a bit-packed representation of a token amount and its decimals.
+/// @dev 64 bytes: [0 - 64] amount
+/// @dev 8 bytes: [64 - 72] decimals
+type TrimmedAmount is uint72;
+
+using {gt as >, lt as <, sub as -, add as +, eq as ==, min, unwrap} for TrimmedAmount global;
 
 function minUint8(uint8 a, uint8 b) pure returns (uint8) {
     return a < b ? a : b;
 }
 
+/// @notice Error when the decimals of two TrimmedAmounts are not equal
+/// @dev Selector. b9cdb6c2
+/// @param decimals the decimals of the first TrimmedAmount
+/// @param decimalsOther the decimals of the second TrimmedAmount
+error NumberOfDecimalsNotEqual(uint8 decimals, uint8 decimalsOther);
+
+uint8 constant TRIMMED_DECIMALS = 8;
+
+function unwrap(TrimmedAmount a) pure returns (uint72) {
+    return TrimmedAmount.unwrap(a);
+}
+
+function packTrimmedAmount(uint64 amt, uint8 decimals) pure returns (TrimmedAmount) {
+    // cast to u72 first to prevent overflow
+    uint72 amount = uint72(amt);
+    uint72 dec = uint72(decimals);
+
+    // shift the amount to the left 8 bits
+    amount <<= 8;
+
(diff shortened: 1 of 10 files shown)


evm/src/libraries/TrimmedAmount.sol
```diff
@@ -2,103 +2,121 @@
...
+function min(TrimmedAmount a, TrimmedAmount b) pure returns (TrimmedAmount) {
+    checkDecimals(a, b);
+
+    return getAmount(a) < getAmount(b) ? a : b;
+}
+
+library TrimmedAmountLib {
+    error AmountTooLarge(uint256 amount);
+
+    function unpackAmount(TrimmedAmount a) internal pure returns (uint64) {
```
kcsongor, line 93: I would just drop this function (and `unpackDecimals`) -- don't see a need to not just use `getAmount` and `getDecimals` directly. would make the diff elsewhere smaller too

evm/test/TrimmedAmount.t.sol
```diff
@@ -9,10 +9,18 @@ contract TrimmingTest is Test {
     using TrimmedAmountLib for uint256;
     using TrimmedAmountLib for TrimmedAmount;
 
+    function test_packUnpack(uint64 amount) public {
+        uint8 decimals = 8;
```
kcsongor, line 13: fuzz this too

kcsongor (changes requested):
nice!

djb15 (commented):
It would be nice to have some tests that fuzz the operator overloading. If the operator overloading isn't doing what it should be then stuff is really going to go wrong 😬 

Alternatively we don't overload the operators at all and just use the methods explicitly

RahulMaganti47:
> It would be nice to have some tests that fuzz the operator overloading. If the operator overloading isn't doing what it should be then stuff is really going to go wrong 😬
> 
> Alternatively we don't overload the operators at all and just use the methods explicitly

i can add some differential fuzz tests (i don't think operator overloading should be that big of an issue -- its just syntactic sugar)

djb15 (approved):
Thanks for these fuzz tests!

gator-boi (approved):
This is a really nice change.

evm/test/TrimmedAmount.t.sol
```diff
@@ -148,12 +159,62 @@ contract TrimmingTest is Test {
...
         // untrim to 6
         uint256 expectedRoundTrip = 5 * 10 ** targetDecimals;
 
         assertEq(expectedRoundTrip, amountRoundTrip);
     }
+
+    function testFuzz_AddOperatorOverload(TrimmedAmount a, TrimmedAmount b) public {
+        vm.assume(a.getDecimals() == b.getDecimals());
+
+        // check if the add operation reverts on an overflow.
```
kcsongor, line 174: this comment doesn't line up with the the test -- we're not checking for overlows here or any revert behaviour -- instead we're only testing non-overflowing cases

kcsongor (approved):

evm/test/TrimmedAmount.t.sol
RahulMaganti47: the line after provides more context, its checking if the sum overflows u64. if so, the test case is discarded via vm.assume; therefore, this test case checks non-overflowing cases

PR #252, opened 2024-03-04
fix: modify [`setIsWormholeEvmChain`] to take bool

### LINKED ISSUE #250: modify evm chain id registration
a registered EVM compatible chain diverges as the result of an upgrade, the existing implementation of `WormholeTransceiverState::setIsWormholeEvmChain` means that it will not be possible to update the corresponding storage. If, for whatever reason, a chain such as Optimism decides to move away from EVM (kind of the opposite of what they did in reality, going from OVM to EVM), its chain ID will now correspond to a non-EVM chain.  This function should take a boolean argument,

RahulMaganti47:
Fixes #250.

### DIFF
--- a/evm/src/interfaces/IWormholeTransceiverState.sol
+++ b/evm/src/interfaces/IWormholeTransceiverState.sol
@@ -8,7 +8,7 @@ interface IWormholeTransceiverState {
     event SetWormholePeer(uint16 chainId, bytes32 peerContract);
     event SetIsWormholeRelayingEnabled(uint16 chainId, bool isRelayingEnabled);
     event SetIsSpecialRelayingEnabled(uint16 chainId, bool isRelayingEnabled);
-    event SetIsWormholeEvmChain(uint16 chainId);
+    event SetIsWormholeEvmChain(uint16 chainId, bool isEvm);
 
     error UnexpectedAdditionalMessages();
     error InvalidVaa(string reason);
--- a/evm/src/interfaces/IWormholeTransceiverState.sol
+++ b/evm/src/interfaces/IWormholeTransceiverState.sol
@@ -50,7 +50,8 @@ interface IWormholeTransceiverState {
     /// @notice Set whether the chain is EVM compatible.
     /// @dev This function is only callable by the `owner`.
     /// @param chainId The Wormhole chain ID to set.
-    function setIsWormholeEvmChain(uint16 chainId) external;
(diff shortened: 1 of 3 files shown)


evm/src/interfaces/IWormholeTransceiverState.sol
```diff
@@ -50,7 +50,7 @@ interface IWormholeTransceiverState {
     /// @notice Set whether the chain is EVM compatible.
     /// @dev This function is only callable by the `owner`.
     /// @param chainId The Wormhole chain ID to set.
```
djb15, line 52: Can you update the docstring to include the new param pls?

RahulMaganti47: yea, fixed.

evm/test/IntegrationRelayer.t.sol
```diff
@@ -172,7 +172,7 @@ contract TestEndToEndRelayer is
         nttManagerChain2.setPeer(chainId1, bytes32(uint256(uint160(address(nttManagerChain1)))), 9);
         DummyToken token2 = DummyTokenMintAndBurn(nttManagerChain2.token());
         wormholeTransceiverChain2.setIsWormholeRelayingEnabled(chainId1, true);
-        wormholeTransceiverChain2.setIsWormholeEvmChain(chainId1);
+        wormholeTransceiverChain2.setIsWormholeEvmChain(chainId1, true);
```
gator-boi, line 175: We should add a test demonstrating that we can set the `IsWormholeEvmChain` to false after setting it to true.

evm/test/IntegrationRelayer.t.sol
```diff
@@ -201,8 +201,24 @@ contract TestEndToEndRelayer is
...
-
             bytes memory instructions = encodeTransceiverInstruction(false);
+
+            // set invalid config
+            vm.stopPrank();
+            wormholeTransceiverChain1.setIsWormholeEvmChain(chainId2, false);
+
+            // config not set correctly
+            vm.startPrank(userA);
+            vm.expectRevert();
```
djb15, line 212: Nit pick, but would be good to expect on a specific error

djb15 (approved):

gator-boi (approved):

djb15 (approved):

gator-boi (approved):

PR #256, opened 2024-03-04
fix: use native evm word size for transceiver registry state variables

### LINKED ISSUE #255: use state variables with native evm word size (instead of booleans) to store transceiver registry flags

RahulMaganti47:
Fixes #255. This PR replaces the boolean state variables for transceiver registry with uint256 integers to reduce gas consumption when setting these variables. Using booleans requires requires an extra SLOAD.


kcsongor:
This adds around 3x overhead, as what used to fit into a single word is now in 3 words. We should definitely not do this

RahulMaganti47:
> This adds around 3x overhead, as what used to fit into a single word is now in 3 words. We should definitely not do this

ah yea didn't consider the storage overhead of the struct.

PR #267, opened 2024-03-06
evm: refactor NttManager in preperation for nft support

gator-boi:
The purpose of this PR is to refactor the NttManager in preparation for adding Nft support. This refactor will reduce code duplication between the two modules.


RahulMaganti47 (commented):
why did you decide to structure the inheritance this way? seems odd to have the NttManagerState inherit the ManagerBase contract?

RahulMaganti47:
> @RahulMaganti47
> 
> I structured it this way so that both Ntt and Nnftt could have implementation specific state files (for example the peer registrations will be different) while also being able to inherit the shared state and logic in `ManagerBase`. The alternative is to shove the implementation specific state into the child contract (e.g,. NttManager).

tbh when i think about this more, latter approach of just putting the implementation specific state into the contract itself

gator-boi:
@RahulMaganti47 I have removed the NttManagerState file. Thanks for the input.

djb15:
Btw this rebase is missing the token decimals change: https://github.com/wormhole-foundation/example-native-token-transfers/pull/238

djb15 (dismissed):

RahulMaganti47 (approved):

djb15 (approved):

Issue #268, opened 2024-03-06
Add overflow checks on sender for amounts that cannot be untrimmed by receiver


johnsaigle:
Agree about adding explicit overflow checks here.

Another check we should consider adding is on the sending side, before the transfer is sent, ensuring the recipient will be able to untrim the amount to its desired decimals. The contracts know their peers' decimals, so they could perform this check before sending out. This way, the transaction would revert on the *sending* side, not the *receiving* side in case the target chain wouldn't be able to handle the amount.

_Originally posted by @kcsongor in https://github.com/wormhole-foundation/example-native-token-transfers/issues/266#issuecomment-1981002675_

djb15:
@johnsaigle what's your view on this issue? What's the range of decimals and word sizes over which this issue could manifest?

johnsaigle:
I gave some examples here: https://github.com/wormhole-foundation/example-native-token-transfers/issues/266#issue-2171615766

The dominant variable is the difference between `to_decimals` and `from_decimals`. The greater this difference, the smaller the amount must be. Conversely, larger amounts only work if the platforms have similar numbers of decimals.

Here's some example Python you can play with to get an idea. Using uint64 because it represents Solana and many other chains. `uint128` is more appropriate for Solidity though we do a lot of conversions/truncations in the codebase.

```python
>>> int(10e18).bit_length() # This number fits in uint64
64
>>> int(2*10e18).bit_length() # this number overflows uint64
65
>>> amount = 10e6
>>> to_decimals = 18
>>> from_decimals = 6
>>> to_decimals - from_decimals
12
>>> scaling_factor = 10e12 # see the `scale()` functions in EVM and Rust implementations
>>> amount * scaling_factor
1e+20
>>> int(amount * scaling_factor).bit_length() # The amount when scaled in `trimmed_amount` overflows uint64
67
```

PR #269, opened 2024-03-06
solana: Add fuzzing and overflow checks for TrimmedAmount

### LINKED ISSUE #266: solana: Integer overflows in TrimmedAmount decimal scaling
`TrimmedAmount::scale` has two potential overflows:

https://github.com/wormhole-foundation/example-native-token-transfers/blob/8936c147c3c4698aa84e016a656c388461f81a35/solana/modules/ntt-messages/src/trimmed_amount.rs#L53-L62

For clarity, I'm adding names to some of these expressions:

```rust
 } else {  // line 59 above
         let difference: u64 = (to_decimals - from_decimals).into();
         let scaling_factor: u64 = 10_u64.pow(difference);
         let product = amount * scaling_factor;
 } 
```

### Scaling Factor exponentiation overflow
(See also https://github.com/wormhole-foundation/example-native-token-transfers/issues/242 for discussion of truncation issues and how the Rust code compares to the Solidity implementation.)

`scaling_factor`, as I've named it above, will overflow when `difference > 18`. This expression is required to be u64, therefore its max value is [u64::MAX](https://doc.rust-lang.org/std/primitive.u64.html#associatedconstant.MAX) which is ~`1.8e19`.

We're dealing only with powers of 10 here. 10e18 is safe but 10e19 is not.

### Overflow occurs when scaling amount
Clearly this value can overflow for very large amounts. If `amount` is near `u64::MAX` then scaling it by even a modest scaling factor will overflow.

Interestingly, this can also occur on the other side of the expression. If `scaling_factor` is large, this limits `amount` significantly. If two peers have very different decimals, the amount must be small. 

e.g. If one chain has 24 decimals and another has 6 decimals, it will create a `scaling_factor` of `10e18` (24 - 6).  Then the only valid amounts are `0` and `1`. Multiplying by any larger (integer) value will overflow because `2 * 10e18 > 1.8e19`.

In a more plausible example, if the two chains have toDecimals == 18 and fromDecimals == 6, then the `scaling_factor` is `10e12`. Then the maximum amount is about 1800000 (`1.8e6`).

### Exponentiation

I think we should throw an error in the function linked above when `difference > 18`. The system can't handle this exponentiation and we might as well document this and throw an error rather than Panic. 

Things are more complicated if we have some sort of weird circumstance in which a chain/token happens to use a very large number of decimals. However, such a chain will be able to only transfer very small amounts. If its decimals are large enough, it will be possible to register the chain but it will not be possible to send any amounts.

Probably this isn't realistic, but perhaps we should be explicit about this in documentation somewhere; it shouldn't be possible to register a chain that has `255` decimals. Registration is an admin function so we don't need to be all that concerned. On the other hand if this is an invalid state for our program, why not enforce that?

### Scaling
As for the relationship between amount and `scaling_factor`, I'm unsure what we should do here. It's an important part of the design to be able to scale the decimals between chains, but it also seems odd to me that the `amount` that can be sent in a transfer is bounded by how dissimilar the chains are in their decimals.

Maybe this is fine? If it is, I think we can add comments, explicit reverts, and unit tests so that we handle overflows defensively rather than relying on Rust to panic (as overflow checks enabled in release). An equivalent of https://github.com/wormhole-foundation/example-native-token-transfers/pull/262 would be a good approach.

johnsaigle:
Add simple fuzz harness for TrimmedAmount::trim
Add overflow checks for issues discovered while fuzzing
Add unit tests to verify that errors are returned from overflow checks
Add new errors (in ntt-messages and in the Anchor custom errors)
Add some additional documentation for functions

This resolves #266. 

### further work, not covered here
- sender overflow checks. see #268 
- Add more fuzz tests for other functions in TrimmedAmount


solana/fuzz/src/fuzz_trimmed_amount.rs
```diff
@@ -0,0 +1,24 @@
+use honggfuzz::fuzz;
+use ntt_messages::trimmed_amount::TrimmedAmount;
+
+fn main() {
+    loop {
+        fuzz!(|data: &[u8]| {
+            // Manually discard data that doesn't fit our inputs.
+            if data.len() != 10 {
```
kcsongor, line 8: this seems inefficient, as it randomly generates inputs and discards the ones that are too long or short. It would be faster (and more readable imo) if we wrote it like
```rust
fuzz!(|data: (u64, u8, u8)| { .. }
```

I think this should be supported based on https://docs.rs/honggfuzz/latest/honggfuzz/macro.fuzz.html (as it supports anything that implements https://docs.rs/arbitrary/latest/arbitrary/trait.Arbitrary.html)

solana/modules/ntt-messages/src/errors.rs
```diff
@@ -0,0 +1,18 @@
...
+    OverflowExponent,
+    OverflowScaledAmount,
+}
+
+impl std::error::Error for ScalingError {}
+
+impl Display for ScalingError {
+    fn fmt(&self, f: &mut Formatter) -> std::fmt::Result {
+        match self {
+            ScalingError::OverflowExponent => write!(f, "File not found"),
```
kcsongor, line 14: "File not found" looks cryptic :)

solana/fuzz/src/fuzz_trimmed_amount.rs
johnsaigle: Makes sense, I'll look into it

solana/modules/ntt-messages/src/trimmed_amount.rs
```diff
@@ -40,47 +42,81 @@ impl TrimmedAmount {
...
+            return Ok(amount);
         }
         if from_decimals > to_decimals {
-            amount / 10u64.pow((from_decimals - to_decimals).into())
+            // [`u64::checked_pow`] expects a u32 argument
+            let power: u32 = (from_decimals - to_decimals).into();
+            // Exponentiation will overflow u64 when `power` is greater than 18
+            let scaling_factor: u64 = match 10u64.checked_pow(power) {
+                Some(scaling_factor) => scaling_factor,
+                None => return Err(ScalingError::OverflowExponent),
```
kcsongor, line 66:
```suggestion
            let scaling_factor: u64 = 10u64.checked_pow(power).map_err(|_| ScalingError::OverflowExponent))?;
```

solana/modules/ntt-messages/src/trimmed_amount.rs
```diff
@@ -40,47 +42,81 @@ impl TrimmedAmount {
...
             decimals: to_decimals,
-        }
+        })
     }
 
-    pub fn untrim(&self, to_decimals: u8) -> u64 {
-        Self::scale(self.amount, self.decimals, to_decimals)
+    pub fn untrim(&self, to_decimals: u8) -> Result<u64, ScalingError> {
+        let scaled_amount = Self::scale(self.amount, self.decimals, to_decimals)?;
+        Ok(scaled_amount)
```
kcsongor, line 104:
```suggestion
        Self::scale(self.amount, self.decimals, to_decimals)?
```

solana/programs/example-native-token-transfers/src/instructions/redeem.rs
```diff
@@ -96,7 +98,17 @@ pub fn redeem(ctx: Context<Redeem>, _args: RedeemArgs) -> Result<()> {
...
+    // Return an error if the resulting amount overflows.
+    // Ideally this state should never be reached: the sender should avoid sending invalid
+    // amounts when they would cause an error on the receiver.
+    let amount = match message.payload.amount.untrim(accs.mint.decimals) {
+        Ok(amount) => amount,
+        Err(ScalingError::OverflowExponent) => return Err(error!(NTTError::OverflowExponent)),
+        Err(ScalingError::OverflowScaledAmount) => {
+            return Err(error!(NTTError::OverflowScaledAMount))
+        }
+    };
```
kcsongor, line 111: instead of duplicating these error mappings everywhere, we could implement a `impl From<ScalingError> for NTTerror` in the `error.rs` file and here (and everywhere else) just do
```suggestion
    let amount = match message.payload.amount.untrim(accs.mint.decimals).map_err(|e| e.into())?;
```

solana/fuzz/src/fuzz_trimmed_amount.rs
RahulMaganti47: just curious - why not use this https://docs.rs/quickcheck/latest/quickcheck/ in conjunction with Arbitrary?

solana/modules/ntt-messages/src/trimmed_amount.rs
johnsaigle: I don't think I can use `map_err` here because `checked_pow` returns Option, not Result.

I'm going to try `ok_or` instead:
```rust
let scaling_factor: u64 = 10u64.checked_pow(power).ok_or(ScalingError::OverflowExponent)?;
```

solana/modules/ntt-messages/src/trimmed_amount.rs
kcsongor: ah right of course

solana/fuzz/src/fuzz_trimmed_amount.rs
kcsongor: honggfuzz seems to be more actively maintained (and built by google). Any reason to use quickcheck @RahulMaganti47 ?

solana/modules/ntt-messages/src/trimmed_amount.rs
kcsongor: nice, yeah `ok_or` should work! my bad

solana/modules/ntt-messages/src/trimmed_amount.rs
```diff
@@ -40,47 +42,78 @@ impl TrimmedAmount {
...
+            // Exponentiation will overflow u64 when `power` is greater than 18
+            let scaling_factor: u64 = 10u64
+                .checked_pow(power)
+                .ok_or(ScalingError::OverflowExponent)?;
+
+            // Return Result: scaled_amount or custom error on overflow
+            match amount.checked_mul(scaling_factor) {
+                Some(scaled_amount) => Ok(scaled_amount),
+                None => Err(ScalingError::OverflowScaledAmount),
+            }
```
kcsongor, line 83:
```suggestion
            amount
                .checked_mul(scaling_factor)
                .ok_or(ScalingError::OverflowScaledAmount)

```

solana/modules/ntt-messages/src/trimmed_amount.rs
```diff
@@ -40,47 +42,78 @@ impl TrimmedAmount {
...
+            // [`u64::checked_pow`] expects a u32 argument
+            let power: u32 = (to_decimals - from_decimals).into();
+
+            // Safely initialize the scaling factor, or return custom error on overflow
+            // Exponentiation will overflow u64 when `power` is greater than 18
+            let scaling_factor: u64 = 10u64
+                .checked_pow(power)
+                .ok_or(ScalingError::OverflowExponent)?;
+
+            // Return Result: scaled_amount or custom error on overflow
```
kcsongor, line 79: (apologies for being nitpicky here!) I think the code is self explanatory here, so the comments don't add that much. Same goes for the above comments

kcsongor:
Really happy with this! Left a couple minor comments

solana/fuzz/src/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
```
kcsongor, line 1: can you check in the accompanying Cargo.lock file too?

johnsaigle: Done!

a5-pickle (approved):

kcsongor (approved):

PR #273, opened 2024-03-06
evm: Add unpausing methods

djb15:


evm/src/Transceiver/Transceiver.sol
```diff
@@ -72,6 +72,11 @@ abstract contract Transceiver is
         _pause();
     }
 
+    /// @dev unpause the transceiver.
+    function _unpauseTransceiver() internal {
+        _unpause();
+    }
+
```
RahulMaganti47, line 79: should we just remove the `pause` and `unpause methods on the transceiver?

djb15: Oh fair yeah it's internal anyway...I was thinking we still want to expose this for integrators, but they can just wrap the underlying `_pause` and `_unpause`

RahulMaganti47: yea exactly

RahulMaganti47 (approved):

johnsaigle (approved):

PR #274, opened 2024-03-07
solana: add _untested_ quoter contract+sdk, introduce SDKv2

nonergodic:


solana/programs/ntt-quoter/src/state.rs
```diff
@@ -0,0 +1,46 @@
...
+    pub fee_recipient: Pubkey,
+    pub sol_price: u64, //UsdPrice (usd/sol [6 decimals])
+                        //amount of sol the user has to pay for the relayer to trigger the release of their outbox message
+                        //pub release_cost: u64, //SolAmount (lamports)
+}
+
+impl Instance {
+    pub const SEED_PREFIX: &'static [u8] = b"instance";
+    pub const BUMP: u8 = crate::INSTANCE_BUMP;
+    pub const SIGNER_SEEDS: &'static [&'static [u8]] = &[Self::SEED_PREFIX, &[Self::BUMP]];
```
kcsongor, line 17: this is not used anywhere

solana/programs/ntt-quoter/src/processor/request_relay.rs
```diff
@@ -0,0 +1,138 @@
...
+//TODO built-in u128 division likely still wastes a ton of compute units
+//     might be more efficient to use f64 or ruint crate
+pub fn request_relay(ctx: Context<RequestRelay>, args: RequestRelayArgs) -> Result<()> {
+    let accs = ctx.accounts;
+
+    require_gt!(
+        args.gas_dropoff,
+        accs.registered_chain.max_gas_dropoff,
+        NttQuoterError::ExceedsMaxGasDropoff
+    );
```
kcsongor, line 81: looks like this should be
```suggestion
    require_gte!(
        accs.registered_chain.max_gas_dropoff,
        args.gas_dropoff,
        NttQuoterError::ExceedsMaxGasDropoff
    );
```

solana/programs/ntt-quoter/src/processor/request_relay.rs
```diff
@@ -0,0 +1,138 @@
...
+
+    let total_fee_in_lamports_without_rent = {
+        let target_native_in_gwei = args.gas_dropoff
+            + if accs.registered_chain.gas_price > 0 {
+                //gas_price[wei, 18 decimals] * EVM_GAS_COST / GWEI = target_native[gwei, 9 decimals]
+                mul_div(accs.registered_chain.gas_price, EVM_GAS_COST, GWEI)
+            } else {
+                0
+            };
```
kcsongor, line 90: Special casing 0 here here seems redundant
```suggestion
            + mul_div(accs.registered_chain.gas_price, EVM_GAS_COST, GWEI)
```
as `mul_div` will correctly handle that anyway.

solana/programs/ntt-quoter/src/processor/request_relay.rs
```diff
@@ -0,0 +1,138 @@
...
+        if total_in_usd > 0 {
+            //total_fee[sol, 9 decimals] = total_usd[usd, 6 decimals] / (sol_price[usd, 6 decimals]
+            mul_div(total_in_usd, LAMPORTS_PER_SOL, accs.instance.sol_price)
+        } else {
+            0
+        }
+    };
+
+    let rent_in_lamports = sysvar::rent::Rent::get()?.minimum_balance(8 + RelayRequest::INIT_SPACE);
+    let total_fee_in_lamports = total_fee_in_lamports_without_rent + rent_in_lamports;
```
kcsongor, line 110: Currently the user is charged for the relay (`total_fee_in_lamports_without_rent`) AND rent. The relayer will claim both, so in addition to getting paid for the relay, it gets the rent lamports too. Wouldn't the right amount to charge the user be
(if we rename `total_fee_in_lamports_without_rent` to `relay_fee`)

`relay_fee.saturating_sub(rent_in_lamports)` ?

In other words, the rent amount should subsidise the relay cost, as the relayer will be able to claim it anyway.

OTOH the account in question is 8 bytes anyway, so there's a case to be made for ignoring it altogether in these calculations.

solana/programs/ntt-quoter/src/processor/request_relay.rs
```diff
@@ -0,0 +1,138 @@
...
+    pub instance: Account<'info, Instance>,
+
+    #[account(mut, address = instance.fee_recipient @ NttQuoterError::InvalidFeeRecipient)]
+    /// CHECK: leave britney alone (anchor complains about a missing check despite the address check)
+    pub fee_recipient: AccountInfo<'info>,
+
+    #[account(
+    seeds = [RegisteredChain::SEED_PREFIX, outbox_item.recipient_chain.id.to_be_bytes().as_ref()],
+    bump = registered_chain.bump,
+    constraint = registered_chain.base_price != u64::MAX @ NttQuoterError::RelayingToChainDisabled
```
kcsongor, line 29: we could improve the readability here by adding an impl function `is_enabled` to `RegisteredChain` that checks if `base_price != u64::MAX`. Then this sentinel value would be closer to the struct definition (I'm assuming you didn't want to use `Option` to save a byte allocation)

solana/programs/ntt-quoter/src/processor/request_relay.rs
```diff
@@ -0,0 +1,138 @@
...
+    owner = example_native_token_transfers::ID,
+    constraint = outbox_item.released.get(WORMHOLE_TRANSCEIVER_INDEX),
+  )]
+    pub outbox_item: Account<'info, OutboxItem>,
+
+    //TODO should this be init_if_needed?
+    //     are there any security considerations here?
+    //     init_if_needed would allow for multiple requests, even after it was already delivered
+    //     right now requesting a relay for a request is permissionless which would allow
+    //       a third party to initiate a relay without gas dropoff for the user (though they
```
kcsongor, line 44: if we do `init_if_needed` here, we should make sure that the gas dropoff is strictly higher than the existing gas dropoff request, otherwise the initial gas dropoff payment is lost.
That said, is the relayer going to close this account immediately upon processing the request? If not, then there would be a race condition where the relayer has already processed the request and someone overrides it, which would be ignored by the relayer. I guess that race condition already exists if someone recreates the account (by requesting a new relay) after it's been closed.
I'm assuming the relayer doesn't monitor new relay requests after the initial one (as my assumption is that the relayer just monitors VAAs and looks up their corresponding relay requests based on the transfer ID)

solana/programs/ntt-quoter/src/processor/request_relay.rs
```diff
@@ -0,0 +1,138 @@
...
+    accs.relay_request.requested_gas_dropoff = args.gas_dropoff;
+
+    //pay the relayer
+    if total_fee_in_lamports_without_rent > 0 {
+        system_program::transfer(
+            CpiContext::new(
+                accs.system_program.to_account_info(),
+                Transfer {
+                    from: accs.payer.to_account_info(),
+                    to: accs.fee_recipient.to_account_info(),
```
kcsongor, line 130: we could consider a simplification here: instead of transferring to the `fee_recipient`, just transfer to the `relay_request` account. Then the `fee_recipient` doesn't even need to be stored in the instance, and can just passed as an argument (well, an account) when closing the `relay_request` account. It also allows us to better see how much payment was made for an individual transfer by just looking at the account rather than logs and whatnot.

solana/ts/sdk/ntt.ts
```diff
@@ -0,0 +1,799 @@
+import { type ChainName, toChainId, coalesceChainId, type ChainId, SignedVaa } from '@certusone/wormhole-sdk'
+import { serializeLayout, toChainId as SDKv2toChainId } from "@wormhole-foundation/sdk-base";
+import { deserialize } from "@wormhole-foundation/sdk-definitions";
+//TODO these need to be properly exported by the SDK
+import { nttManagerMessageLayout, NttManagerMessage, nativeTokenTransferLayout } from "@wormhole-foundation/sdk-definitions/dist/cjs/payloads/ntt";
```
barnjamin, line 5: we're not exporting layouts for payloads directly, should we?

We've been exporting VAA specific stuff from the Protocols namespace 

https://github.com/wormhole-foundation/connect-sdk/blob/main/core/definitions/src/protocols/tokenBridge.ts#L29-L61

solana/ts/sdk/ntt.ts
```diff
@@ -0,0 +1,799 @@
...
+import { derivePostedVaaKey, getWormholeDerivedAccounts } from '@certusone/wormhole-sdk/lib/cjs/solana/wormhole'
```
barnjamin, line 6: these should also be available in the sdkv2

https://github.com/wormhole-foundation/connect-sdk/blob/5f530e25e974c51e078a34a710b765deac662082/platforms/solana/protocols/core/src/utils/accounts/postedVaa.ts#L4

https://github.com/wormhole-foundation/connect-sdk/blob/5f530e25e974c51e078a34a710b765deac662082/platforms/solana/protocols/core/src/utils/cpi.ts#L38

solana/ts/sdk/ntt.ts
```diff
@@ -0,0 +1,799 @@
+import { type ChainName, toChainId, coalesceChainId, type ChainId, SignedVaa } from '@certusone/wormhole-sdk'
```
barnjamin, line 1: should we provide an ez way to go from sdkv1 chain names to sdkv2 chain names?

solana/programs/ntt-quoter/src/processor/request_relay.rs
nonergodic: Yes, this is partly accidental, partly intentional.
Originally it was in there because it prevented a division by zero. That's not the case anymore but I figured I'd leave it in because I assumed that [u128 divisions can still be very expensive](https://github.com/solana-labs/solana/issues/19549) and so while, in the interest of time, I didn't want to bother mucking around with pulling another crate in and getting it to work with better datatypes, but at the same time I figured I can at least avoid the cost.

The check should at least be moved into `mul_div` though.

solana/ts/sdk/ntt.ts
barnjamin: these are now exported

solana/programs/ntt-quoter/src/processor/request_relay.rs
nonergodic: yes, that was actually the intention, but I screwed it up. :grimacing:

solana/programs/ntt-quoter/src/processor/request_relay.rs
nonergodic: That was the original plan, but then I figured it would actually be operationally easier to have the lion's share of the fee flow back to the fee recipient immediately without having to rely on running some off-chain process to collect them, which might be helpful during times of high through-put. Otoh it also introduces account competition around the fee recipient, which sucks so perhaps having the fully parallel, more legible solution is cleaner.

solana/programs/ntt-quoter/src/state.rs
```diff
@@ -0,0 +1,46 @@
...
+    pub assistant: Pubkey,
+    pub fee_recipient: Pubkey,
+    pub sol_price: u64, //UsdPrice (usd/sol [6 decimals])
+                        //amount of sol the user has to pay for the relayer to trigger the release of their outbox message
+                        //pub release_cost: u64, //SolAmount (lamports)
+}
+
+impl Instance {
+    pub const SEED_PREFIX: &'static [u8] = b"instance";
+    pub const BUMP: u8 = crate::INSTANCE_BUMP;
```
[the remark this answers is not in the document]
nonergodic: Yeah, it's an option to rely on the account discriminator alone, rather than also check the PDA. I didn't spend a lot of time worrying about future compatibility - I think the only truly relevant issue in this regard is making sure that the program account can be upgraded to a larger program if there ever is the need for a more powerful/scalable solution.

This is very much an quick mvp.

solana/programs/ntt-quoter/src/processor/request_relay.rs
nonergodic: [I already had the comment in here in an early version](https://github.com/nonergodic/ntt-solana-relayer/blob/6ca388a9bd5db78345a0a439f6f2fcee28b72a7a/programs/ntt-relayer/src/processor/request_relay.rs#L111) but after trying to wrap things up (too) quickly last night I was like "hey, no need to sub if I just don't add it in!" which was a complete brain-fart...

solana/programs/ntt-quoter/src/processor/request_relay.rs
nonergodic: > I'm assuming the relayer doesn't monitor new relay requests after the initial one (as my assumption is that the relayer just monitors VAAs and looks up their corresponding relay requests based on the transfer ID)

Correct. My concern was actually primarily around whether someone can prevent an account from being initialized by sending SOL to it (because then in some sense the account is already initialized, even if it has no data), but that was really just an "in the moment confusion" because in that case you could prevent any account that uses `init` and has a predictable address from being created... otoh, I do remember that there was some concern around this in the core bridge rewrite and so I left it in as a reminder to myself while also pointing out that `init_if_needed` would have unexpected consequences too (users being able to pointlessly pay multiple times).

solana/ts/sdk/ntt.ts
nonergodic: yeah, I didn't try to purge the old sdk in its entirety, only replace the redundant payloads and combine overlapping functionality between ntt and quoter.

solana/ts/sdk/ntt.ts
nonergodic: Imho no. The thing should be retired in the Bladerunner sense, not in the senior citizen sense.

solana/programs/ntt-quoter/src/lib.rs
```diff
@@ -0,0 +1,81 @@
...
+// !! update constant in TypeScript SDK when updating this value !!
+const EVM_GAS_COST: u64 = 250_000; //TODO determine true gas cost on EVM
+
+cfg_if::cfg_if! {
+  if #[cfg(feature = "solana-devnet")] {
+    declare_id!("");
+    const WORMHOLE_TRANSCEIVER_INDEX: u8 = 0;
+  } else if #[cfg(feature = "tilt-devnet")] {
+    declare_id!("NttRe1ayer111111111111111111111111111111111");
+    const INSTANCE_BUMP: u8 = 255; //TODO calculate via TypeScript SDK calcInstanceBump
```
kcsongor, line 28: we can delete this now

solana/ts/sdk/utils.ts
```diff
@@ -0,0 +1,64 @@
...
+  { name: "slot", binary: "uint", endianness: "little", size: 8 },
+  { name: "upgradeAuthority", binary: "switch", idSize: 1, idTag: "isSome", layouts: [
+      [[0, false], []],
+      [[1, true], [{name: "value", binary: "bytes", size: 32, custom: pubKeyConversion}]],
+    ],
+  }
+] as const satisfies Layout;
+
+export const U64 = {
+  MAX: new BN((2n**64n - 1n).toString()),
```
johnsaigle, line 39: CI tests seem to be failing around here. I'm not sure exactly which ES version is used by the CI pipeline.

@evan-gray might be able to shed some light here? Otherwise I'll look around and see whether I can track down the issue. It looks like we'll need to update the version used, or else find a different BigInt solution here.

```
ntt-ci-tests │ TSError: ⨯ Unable to compile TypeScript:
 ntt-ci-tests │ solana_binding/ts/sdk/utils.ts(39,16): error TS2737: BigInt literals are not available when targeting lower than ES2020.
 ntt-ci-tests │ solana_binding/ts/sdk/utils.ts(39,16): error TS2791: Exponentiation cannot be performed on 'bigint' values unless the 'target' option is set to 'es2016' or later.
 ntt-ci-tests │ solana_binding/ts/sdk/utils.ts(39,20): error TS2737: BigInt literals are not available when targeting lower than ES2020.
 ntt-ci-tests │ solana_binding/ts/sdk/utils.ts(39,26): error TS2737: BigInt literals are not available when targeting lower than ES2020.
 ```

File: solana/package.json [hunk 14]
```diff
@@ -7,6 +7,8 @@
     },
     "dependencies": {
         "@certusone/wormhole-sdk": "^0.10.10",
+        "@wormhole-foundation/sdk-base": "^0.5.0-beta.9",
```
johnsaigle, line 10: Can you ensure that these match `package-lock.json`? There seems to be a mismatch in CI and I'm seeing the same thing when running the equivalent commands locally.

johnsaigle (changes requested):
Some suggestions for fixing the CI failures

solana/ts/sdk/utils.ts
johnsaigle: Ah I think the issue is in the modifications in `tsconfig.json`. These need to be updated within CI, or else you'll need to use methods compatible with the old version and revert `tsconfig.json`

solana/package.json
nonergodic: yeah, realized that I updated both package.jsons but only ran npm install in the ci_test directory...

nonergodic:
The gods (CI pipeline) were finally merciful and have bestowed mana (a pass) upon me!

johnsaigle (approved):

kcsongor (approved):

PR #275, opened 2024-03-07
solana: require that transfer is not released in redeem

kcsongor:


solana/programs/example-native-token-transfers/src/instructions/redeem.rs
```diff
@@ -134,7 +134,7 @@ pub fn redeem(ctx: Context<Redeem>, _args: RedeemArgs) -> Result<()> {
         RateLimitResult::Delayed(release_timestamp) => release_timestamp,
     };
 
-    accs.inbox_item.release_status = ReleaseStatus::ReleaseAfter(release_timestamp);
+    accs.inbox_item.release_after(release_timestamp)?;
```
johnsaigle, line 137: Could this be moved to L114 (above the rate limit check)? Since `release_after` has the possibility to return early, we might as well avoid doing the rate limiting calculations

johnsaigle (approved):
The fix looks good. I suggested a change that might be a bit more efficient but it can be done in a subsequent PR

solana/programs/example-native-token-transfers/src/instructions/redeem.rs
kcsongor: agree it would be more efficient to perform the check first, then do the calculation, then set the value. I was thinking this way it's a little more atomic (since the setting happens in the same function as the assertion)

a5-pickle: If not done atomically via your `release_after` method, might be nice to have it as a constraint in the account context, something like:

```rust
        constraint = inbox_item.release_status == ReleaseStatus::NotApproved @ NTTError::TransferCannotBeRedeemed
```

Just so it's clear from reading the instruction's account context that it cannot be redeemed if it isn't approved yet. This works because `NotApproved` is serialized as 0 and this byte is 0 when this account is initialized.

a5-pickle (approved):
Just responded to your conversation w/ @johnsaigle about the release status check location. Like he says, can be changed in a subsequent PR.

solana/programs/example-native-token-transfers/src/instructions/redeem.rs
kcsongor: sounds good, I'll add that in a follow up PR

kcsongor:
> When https://github.com/wormhole-foundation/example-native-token-transfers/pull/219 is merged, it would be great to add a test for this change using the tools in that PR

that was my thinking too!

PR #286, opened 2024-03-08
evm: Query wormhole fee and improve quote view

djb15:
We now query the wormhole fee to ensure NTT deployments with the Wormhole transceiver don't need to make an upgrade if the fee is changed to be non-zero.

I've also updated the public view for `quoteDeliveryPrice` to make it more useful since the caller now only has to pass in the transceiver instructions, not the list of enabled transceivers as well.


gator-boi (approved):

RahulMaganti47 (approved):
lgtm

mdulin2 (approved):
Looks good to me.

PR #290, opened 2024-03-11
solana: replay protect gov messages

kcsongor:


johnsaigle (approved):

a5-pickle (approved):
Looks good. An enhancement to the expected error would be checking that the pubkey is the replay account's in a future PR.

PR #292, opened 2024-03-11
evm: Revert if manager peer is not registered

djb15:
There is no test case for this because this never gets hit...we always hit the `InvalidPeerDecimals` error first. So the test that this doesn't break anything is the fact that all the existing tests pass

### DIFF
--- a/evm/src/NttManager/NttManager.sol
+++ b/evm/src/NttManager/NttManager.sol
@@ -246,6 +246,11 @@ contract NttManager is INttManager, NttManagerState {
         uint256 numEnabledTransceivers = enabledTransceivers.length;
         mapping(address => TransceiverInfo) storage transceiverInfos = _getTransceiverInfosStorage();
         bytes32 peerAddress = _getPeersStorage()[recipientChain].peerAddress;
+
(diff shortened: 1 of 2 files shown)


RahulMaganti47:
> doesn't

not sure i understand the point of this addition -- is this a defense-in-depth check against potentially blackholeing funds as a result of the peer being set to zero?

djb15:
> not sure i understand the point of this addition -- is this a defense-in-depth check against potentially blackholeing funds as a result of the peer being set to zero? we already have a check against that in `setPeer`?

Yeah entirely defence in depth. More to prevent blackholing funds by trying to send to a chain id that doesn't have an associated peer. Right now we get saved by the decimal check of the peer when trimming the transfer amount (since the peer isn't registered so has 0 decimals by default), but imo it's nice to have this explicit backstop

RahulMaganti47:
ahhh right - I remember having this discussion. yea definitely better to expose errors explicitly

RahulMaganti47 (approved):

johnsaigle (approved):

PR #298, opened 2024-03-12
solana: Add check for no registered transceivers

johnsaigle:
Add guard on transfer instructions to prevent a transfer from being processed when no transceivers are registered. This is done by checking the `next_transceiver_id` field in the Config. This field auto-increments when a transceiver is added. If this value is 0, then the program is in a state where no transceivers have been registered.

See also https://github.com/wormhole-foundation/example-native-token-transfers/pull/289 for equivalent checks on the EVM side.

Note that, at this point, [there is no ability for Transceivers to be disabled within the Solana implementation](https://github.com/wormhole-foundation/example-native-token-transfers/issues/251). This reduces the risk explored in #289 as the only moment where no Transceivers are registered occurs just after deployment.

### DIFF
--- a/solana/programs/example-native-token-transfers/src/instructions/redeem.rs
+++ b/solana/programs/example-native-token-transfers/src/instructions/redeem.rs
@@ -24,6 +24,8 @@ pub struct Redeem<'info> {
     // NOTE: this works when the contract is paused
     #[account(
         constraint = config.threshold > 0 @ NTTError::ZeroThreshold,
+        constraint = config.next_transceiver_id != 0 @ NTTError::NoRegisteredTransceivers,
+        constraint = !config.enabled_transceivers.is_empty() @ NTTError::NoRegisteredTransceivers,
     )]
     pub config: Account<'info, Config>,
 
--- a/solana/programs/example-native-token-transfers/src/instructions/transfer.rs
+++ b/solana/programs/example-native-token-transfers/src/instructions/transfer.rs
@@ -22,6 +22,11 @@ pub struct Transfer<'info> {
     #[account(mut)]
     pub payer: Signer<'info>,
 
+    // Ensure that there exists at least one enabled transceiver
+    #[account(
+        constraint = config.next_transceiver_id != 0 @ NTTError::NoRegisteredTransceivers,
+        constraint = !config.enabled_transceivers.is_empty() @ NTTError::NoRegisteredTransceivers,
+    )]
     pub config: NotPausedConfig<'info>,
 
     #[account(
@@ -120,6 +125,7 @@ pub fn transfer_burn(ctx: Context<TransferBurn>, args: TransferArgs) -> Result<(
     );
 
     let accs = ctx.accounts;
+
     let TransferArgs {
         mut amount,
         recipient_chain,
@@ -223,6 +229,7 @@ pub fn transfer_lock(ctx: Context<TransferLock>, args: TransferArgs) -> Result<(
     );
 
     let accs = ctx.accounts;
+
     let TransferArgs {
         mut amount,
         recipient_chain,
--- a/solana/tests/example-native-token-transfer.ts
+++ b/solana/tests/example-native-token-transfer.ts
@@ -95,20 +95,28 @@ describe("example-native-token-transfers", () => {
         mode: "locking",
       });
 
-      await ntt.registerTransceiver({
+      const transceiver = await ntt.registerTransceiver({
         payer,
         owner: payer,
         transceiver: ntt.program.programId,
       });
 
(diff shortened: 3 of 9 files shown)


solana/programs/example-native-token-transfers/src/instructions/transfer.rs
```diff
@@ -120,6 +120,15 @@ pub fn transfer_burn(ctx: Context<TransferBurn>, args: TransferArgs) -> Result<(
     );
 
     let accs = ctx.accounts;
+
+    // TODO If functionality is added to disable transceivers, this check should be modified to
+    // ensure that, if all transceivers have been disabled, that an error is returned. Otherwise it
+    // may be possible to permanently lose funds.
+    // No transceivers have been registered yet
+    if accs.common.config.next_transceiver_id == 0 {
```
a5-pickle, line 128: Can this check be put in the common account context, maybe in the `NotPausedConfig`:
```rust
#[derive(Accounts)]
pub struct NotPausedConfig<'info> {
    #[account(
        constraint = !config.paused @ crate::error::NTTError::Paused,
        constraint = config.next_transceiver_id != 0 @ NTTError::NoRegisteredTransceivers
    )]
    config: Account<'info, Config>,
}
```

This would avoid having to put this check in each transfer instruction.

a5-pickle: If it is undesirable to do this check for everything that uses `NotPausedConfig` btw, adding a constraint over `common` works, too:
```rust
    #[account(
        constraint = config.next_transceiver_id != 0 @ NTTError::NoRegisteredTransceivers
    )]
    pub config: NotPausedConfig<'info>,
```

johnsaigle: Thanks for the review!
Yeah my intention is just to guard the Transfer functions so adding that constraint in the transfer context makes sense. I'll modify the code.

johnsaigle:
@a5-pickle I changed the code to add an account constraint instead of the ad-hoc check. Now that it exists, the unit tests is failing. I'm not sure if this is revealing that we had a bug in the code, or if this is an issue of mocking and that the problem exists in the way tests are set-up.

@kcsongor when you have time I'd appreciate if you could maybe checkout the branch and take a look at the tests. I tried poking around the test suite for a while but didn't make much progress. While a transceiver appears to be registered during the SDK's `setup()` function, I'm not sure that the test/mock (NotPaused)Config is updated, in particular its `enabled_transceivers` and `next_transceiver_id` fields.

solana/programs/example-native-token-transfers/src/instructions/redeem.rs
```diff
@@ -22,6 +22,10 @@ pub struct Redeem<'info> {
     pub payer: Signer<'info>,
 
     // NOTE: this works when the contract is paused
+    #[account(
+        constraint = config.next_transceiver_id != 0 @ NTTError::NoRegisteredTransceivers,
+        constraint = config.enabled_transceivers.is_empty() @ NTTError::NoRegisteredTransceivers,
```
johnsaigle, line 27: This adds parity with the EVM implementation:
https://github.com/wormhole-foundation/example-native-token-transfers/blob/0d764bf7ffbbf81b4d51d1bfd5e51929e7fb0460/evm/src/NttManager/ManagerBase.sol#L260

johnsaigle, line 27: `redeem()` will fail if there aren't enough votes/attestations on a message and it's not possible to have attestations without transceivers, so we might as well fail early and with a more accurate error message.

solana/tests/example-native-token-transfer.ts
```diff
@@ -101,7 +110,8 @@ describe('example-native-token-transfers', () => {
         amount,
         recipientChain: 'ethereum',
         recipientAddress: Array.from(user.publicKey.toBuffer()), // TODO: dummy
-        shouldQueue: false
+        shouldQueue: false,
+        config: await ntt.getConfig()
```
kcsongor, line 114: is this necessary? the function internally will make this exact function call anyway

solana/tests/example-native-token-transfer.ts
```diff
@@ -180,6 +190,7 @@ describe('example-native-token-transfers', () => {
       const released = await ntt.redeem({
         payer,
         vaa: vaaBuf,
+        config: await ntt.getConfig()
```
kcsongor, line 193: same as above

kcsongor (changes requested):
looks good overall, left a couple of minor comments

a5-pickle (approved):

nvsriram:
Changes merged through #514

PR #299, opened 2024-03-12
solana: Add build, test, and lint Makefile targets to top-level dir

johnsaigle:
Adds Makefile targets for common Solana tasks in the top-level directory. This allows the test and build pipeline to be run from the main directory for both the EVM and Solana implementations.

This also helps to capture the Solana commands that are configured in the [.github/workflows/solana.yml](https://github.com/wormhole-foundation/example-native-token-transfers/blob/main/.github/workflows/solana.yml) file. In case this codebase is merged or adapted into the Wormhole monorepo, the GitHub workflow may be removed for modified. Therefore it is beneficial to track the expected test configurations here.

See also: https://github.com/wormhole-foundation/example-native-token-transfers/pull/285. Part of the motivation for adding these Makefile targets is to configure additional clippy rules that can be run locally in addition to CI runs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,14 @@ build-evm-prod: clean-evm
 gen-evm-bindings: build-evm-prod
 	npm ci && cd evm/ts && npm run generate
 
+.PHONY: build-solana
+build-solana:
+	cd solana; BPF_OUT_DIR="$(pwd)/target/deploy" cargo build-sbf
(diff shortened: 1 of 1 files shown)


nvsriram:
Changes merged through https://github.com/wormhole-foundation/example-native-token-transfers/pull/502

PR #301, opened 2024-03-12
evm: Use OZ v5 directly

djb15:
The `PUSH0` instructions we're seeing in the opcode output is actually the metadata literal `0x5f` being interpreted as a `PUSH0`. The metadata is appended to the end of the actual contract bytecode but will not be executed as a PUSH0.

This change updates the repo to use OZ v5 libraries directly rather than importing and bumping down the compiler version.


RahulMaganti47 (approved):

djb15:
I suggest we close this for now as it's quite an intrusive change this late in the day. We can always consider doing this in the future

PR #304, opened 2024-03-13
solana: introduce "session authority" to remove atomic approval requirement

kcsongor:


solana/programs/example-native-token-transfers/src/instructions/transfer.rs
```diff
@@ -71,6 +62,23 @@ pub struct TransferArgs {
...
 
+impl TransferArgs {
+    pub fn keccak256(&self) -> solana_program::keccak::Hash {
+        let TransferArgs {
+            amount,
+            recipient_chain,
+            recipient_address,
+            should_queue,
+        } = self;
+        solana_program::keccak::hashv(&[
```
johnsaigle, line 73: A couple points here when calculating the hash:

- If a user sends a transaction with the same amount, recipient chain, address, they will get the same hash. So let's say they send 10 tokens to themselves twice. It will generate the same authority. Is there any problem here? I suppose it's still gated by the SPL Token program but I'm curious if you can think of a reason why this might cause a strange edge-case

- Would there any benefit to calculating separate hashes for lock vs. burn transfers?

a5-pickle (approved):
Nice fix

solana/programs/example-native-token-transfers/src/instructions/transfer.rs
kcsongor: interesting edge case! after sending the first one, the approval will be reset back to 0 (since the approval amount will have been used up), and the next one will just proceed normally.
The way to think about a session authority is that it's a token authority that can send a specified amount of tokens to a specified recipient. If someone wants to do this multiple times, it's perfectly fine to reuse the same authority.

On the second point -- I considered this, but I think it's fine not to. It's an implementation detail anyway, any a program support exactly one of those paths, so it's not possible to send it to the wrong one

johnsaigle (approved):

PR #316, opened 2024-03-18
evm: Adds Transceiver Payload V1

solanoepalacio:
Adds transceiver payload to notify specialized relayer about messages to be relayed.
The payload is composed by:
- 1 byte prefix representing the version
- 1 byte representing a boolean that indicates to the offchain specialized relayer whether or not the VAA should be automatically redeemed


evm/src/Transceiver/WormholeTransceiver/WormholeTransceiver.sol
```diff
@@ -250,4 +246,38 @@ contract WormholeTransceiver is
...
+        bytes memory
+    ) {
+        // Transceiver payload is prefixed with 2 bytes
+        // representing the version of the payload.
+        // The rest of the bytes are the -actual- payload data.
+        // In payload v1, the payload data is a boolean
+        // representing whether the message should be picked
+        // up by the special relayer or not.
+        bytes memory transceiverPayload = abi.encodePacked(
+            uint16(1),
```
mdulin2, line 266: Is setting this to 1 technically correct?

mdulin2: I thought it'd be the `TranseiverInstruction` from the previous call? 

I don't think the Wormhole Transceiver actually processes this field atm though. It's more for the way in than the way out.

nonergodic: It's only in the wormhole transceiver payload to make life for the specialized relayer easier: Namely, so it can look at the VAA alone to figure out if it is on the hook for relaying it or not. It has no relevance beyond that.

nonergodic:
[sdkv2 layout changes PR](https://github.com/wormhole-foundation/wormhole-sdk-ts/pull/419) to reflect changes made here

evm/src/Transceiver/WormholeTransceiver/WormholeTransceiver.sol
solanoepalacio: If we use `TransceiverInstruction` we are leaving it up to the client to pass the right payload, this would be worst in terms of trust for the offchain relayer (because we'd now have to trust that the client offchain code, which we can't audit as easily as inspecting a verified contract)

evm/src/Transceiver/WormholeTransceiver/WormholeTransceiver.sol
```diff
@@ -271,4 +273,32 @@ contract WormholeTransceiver is
         checkFork(wormholeTransceiver_evmChainId);
         return getWormholePeer(vm.emitterChainId) == vm.emitterAddress;
     }
+
+    function _encodePayload(
```
djb15, line 277: Can we call this `_encodeTransceiverPayload` or something more descriptive instead please?

evm/src/Transceiver/WormholeTransceiver/WormholeTransceiver.sol
```diff
@@ -271,4 +273,32 @@ contract WormholeTransceiver is
...
+        bytes memory nttManagerMessage,
+        bool isSpecialRelayer
+    ) internal pure returns (TransceiverStructs.TransceiverMessage memory, bytes memory) {
+        // Transceiver payload is prefixed with 2 bytes
+        // representing the version of the payload.
+        // The rest of the bytes are the -actual- payload data.
+        // In payload v1, the payload data is a boolean
+        // representing whether the message should be picked
+        // up by the special relayer or not.
+        bytes memory transceiverPayload = abi.encodePacked(uint16(1), isSpecialRelayer);
```
djb15, line 289: Why is the version a uint16? I assume a uint8 should be sufficient?

djb15 (commented):
This doesn't seem very efficient as we're encoding everything twice when sending via the special relayer? Should we instead change the encoding methods to add on a transceiver payload, or instead do the building and encoding inside the if blocks

iturricf:
> This doesn't seem very efficient as we're encoding everything twice when sending via the special relayer? Should we instead change the encoding methods to add on a transceiver payload, or instead do the building and encoding inside the if blocks

Good call, the building and encoding was inside the blocks originally and then we changed that back. I will put it inside the blocks again.

evm/src/Transceiver/WormholeTransceiver/WormholeTransceiver.sol
iturricf: done

evm/src/Transceiver/WormholeTransceiver/WormholeTransceiver.sol
iturricf: you're right, there's no need for it to be uint16, uint8 is enough, I'm changing this

iturricf:
A forge format issue has been introduced to `main`, I'll fix it here

djb15 (commented):
With this pattern of doing the encoding inside each if-else block, does it make more sense to get rid of the extra method all together? We can just call `buildAndEncodeTransceiverMessage` directly, and in the special relayer block we just add the packed bytes as the last argument? We can keep some of that docstring to explain the transceiver payload structure.

Also is it worth diffing these changes into a fresh PR, this one is getting pretty messy with the commit history?

iturricf:
> With this pattern of doing the encoding inside each if-else block, does it make more sense to get rid of the extra method all together? We can just call `buildAndEncodeTransceiverMessage` directly, and in the special relayer block we just add the packed bytes as the last argument? We can keep some of that docstring to explain the transceiver payload structure.
> 

alright, I can make it so it would call `buildAndEncodeTransceiverMessage` directly, and the a reduced docstring for the special relayer case.

> Also is it worth diffing these changes into a fresh PR, this one is getting pretty messy with the commit history?

About this, do you think using Squash and merge would be enough?

djb15:
> alright, I can make it so it would call buildAndEncodeTransceiverMessage directly, and the a reduced docstring for the special relayer case.

Sounds good, thanks!

> About this, do you think using Squash and merge would be enough?

True, yeah provided we squash it's ok, especially if you update the commit description to not list all the commits being squashed

djb15 (dismissed):
Lgtm, just needs a rebase

djb15 (approved):

iturricf (approved):

PR #318, opened 2024-03-19
fix layout/parsing of bpf_loader_upgradeable ProgramData

nonergodic:


nonergodic:
Decided to add the fix to the deployment-scripts branch instead.

PR #319, opened 2024-03-19
solana: support for multiple ntt tokens with the same quoter

nonergodic:
* assumes evm gas cost is the same across all chains
* very egalitarian - does not allow for subsidizing costs

### DIFF
--- a/solana/programs/ntt-quoter/src/processor/admin.rs
+++ b/solana/programs/ntt-quoter/src/processor/admin.rs
@@ -154,10 +154,7 @@ pub struct RegisterNtt<'info> {
     system_program: Program<'info, System>,
 }
 
-pub fn register_ntt(
-    ctx: Context<RegisterNtt>,
-    args: RegisterNttArgs
-) -> Result<()> {
+pub fn register_ntt(ctx: Context<RegisterNtt>, args: RegisterNttArgs) -> Result<()> {
     ctx.accounts.registered_ntt.bump = ctx.bumps.registered_ntt;
     ctx.accounts.registered_ntt.wormhole_transceiver_index = args.wormhole_transceiver_index;
     ctx.accounts.registered_ntt.gas_cost = args.gas_cost;
--- a/solana/programs/ntt-quoter/src/processor/admin.rs
+++ b/solana/programs/ntt-quoter/src/processor/admin.rs
@@ -1,6 +1,6 @@
 use crate::{
     error::NttQuoterError,
-    state::{Instance, RegisteredChain},
+    state::{Instance, RegisteredChain, RegisteredNtt},
 };
 use anchor_lang::prelude::*;
 use wormhole_solana_utils::cpi::bpf_loader_upgradeable as bpf;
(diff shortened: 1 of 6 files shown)


solana/programs/ntt-quoter/src/processor/admin.rs
```diff
@@ -121,3 +121,73 @@ pub fn register_chain(ctx: Context<RegisterChain>, _args: RegisterChainArgs) ->
...
+        bump,
+    )]
+    pub registered_ntt: Account<'info, RegisteredNtt>,
+
+    system_program: Program<'info, System>,
+}
+
+pub fn register_ntt(ctx: Context<RegisterNtt>, args: RegisterNttArgs) -> Result<()> {
+    ctx.accounts.registered_ntt.bump = ctx.bumps.registered_ntt;
+    ctx.accounts.registered_ntt.wormhole_transceiver_index = args.wormhole_transceiver_index;
```
johnsaigle, line 159: If this function is called twice, will it overwrite an existing index?
Maybe we could add a constraint for `wormhole_transceiver_index` to make sure that it is a new value. There's probably not too much risk but it would be good to prevent accidentally updating an existing entry.

I also think it could be good to constrain this to a value between 0-127 as the max number of transceivers is 128. (https://github.com/wormhole-foundation/example-native-token-transfers/issues/296)

solana/programs/ntt-quoter/src/processor/admin.rs
```diff
@@ -121,3 +121,73 @@ pub fn register_chain(ctx: Context<RegisterChain>, _args: RegisterChainArgs) ->
...
+        seeds = [RegisteredNtt::SEED_PREFIX, args.ntt_program_id.key().as_ref()],
+        bump = registered_ntt.bump,
+    )]
+    pub registered_ntt: Account<'info, RegisteredNtt>,
+
+    system_program: Program<'info, System>,
+}
+
+pub fn deregister_ntt(_ctx: Context<DeregisterNtt>, _args: DeregisterNttArgs) -> Result<()> {
+    Ok(())
```
johnsaigle, line 192: It could be nice to add a `msg!()` statement here to be able to check that the account has been closed

solana/programs/ntt-quoter/src/processor/request_relay.rs
```diff
@@ -29,10 +28,18 @@ pub struct RequestRelay<'info> {
...
+            RegisteredNtt::SEED_PREFIX,
+            (&outbox_item as &dyn AsRef<AccountInfo<'info>>).as_ref().owner.to_bytes().as_ref()
+        ],
+        bump = registered_ntt.bump,
+    )]
+    pub registered_ntt: Account<'info, RegisteredNtt>,
+
     //TODO eventually drop the released constraint and instead implement release by relayer
     #[account(
-        owner = example_native_token_transfers::ID,
```
johnsaigle, line 34: Is there any other ownership check or permission that we should be checking here?
My initial thinking is no because this is a relayer so it does not need any sort of authorization. But I still want to raise it in case we are missing something

johnsaigle (commented):
Some minor questions and suggestions. Overall looks like a good change

solana/programs/ntt-quoter/src/processor/admin.rs
nonergodic: Anchor will log the the instruction that's being executed anyway, no?

solana/programs/ntt-quoter/src/processor/admin.rs
nonergodic: No such thing as calling it twice, because the `registered_ntt` account must be uninitalized (it's marked `init` in the context). If you want to change something, you'll have to deregister and re-register.

Regarding the max value of 128: If you register the wrong value, things will be broken anyway. I'd rather not duplicate what is effectively an implementation detail of the ntt program in the quoter, especially for something so marginal. Realistically, any value above 4 is almost certainly wrong.

solana/programs/ntt-quoter/src/processor/request_relay.rs
nonergodic: My main concern when writing this was actually that Anchor might check the owner, even though I don't intend for it to do so:
https://docs.rs/anchor-lang/latest/anchor_lang/accounts/account/struct.Account.html

In fact, when I implemented it the way it is right now, I missed that `OutboxItem` actually has the Anchor `#[account]` attribute and thought it was just a normal struct with some derives... (without the account attribute, it wouldn't implement the owner trait and so we'd be in the clear)

So this is actually buggy because it will introduce this unwanted owner check, which is what might have made you subconsciously feel uneasy when looking at this!

solana/programs/ntt-quoter/src/processor/admin.rs
johnsaigle: You're right, thanks I got tunnel vision on the possibility of overwriting the value and overlooked `init`.

Fair enough. I'm generally in favour of handling things explicitly but it's not worth blocking this PR

solana/programs/ntt-quoter/src/processor/request_relay.rs
nonergodic: The original implementation was already bullshit / a red herring because `Account` checks that the owner is `T::owner()`. For some reason, I thought it would check that the owner is my current program without the `owner = ...` but that's false.

johnsaigle: Ah OK, I assumed that this was a temporary hard-coding similar to the WORMHOLE_TRANSCEIVER_INDEX.

So basically this should be removed (like you have it in this PR) and Anchor should take care of it like normal. Is that your understanding too?

nonergodic: What I need is something that deserializes the account (and checks the discriminator) but without checking the account owner (because I don't know or care which ntt program it is). I don't think Anchor provides this functionality (because it is a bit of a weird Frankenstein) so I'm currently looking into what the best way is to accomplish this.

solana/programs/ntt-quoter/src/processor/admin.rs
```diff
@@ -121,3 +121,71 @@ pub fn register_chain(ctx: Context<RegisterChain>, _args: RegisterChainArgs) ->
...
+        seeds = [RegisteredNtt::SEED_PREFIX, args.ntt_program_id.key().as_ref()],
+        bump,
+    )]
+    pub registered_ntt: Account<'info, RegisteredNtt>,
+
+    system_program: Program<'info, System>,
+}
+
+pub fn register_ntt(ctx: Context<RegisterNtt>, args: RegisterNttArgs) -> Result<()> {
+    ctx.accounts.registered_ntt.bump = ctx.bumps.registered_ntt;
```
kcsongor, line 156: nit: `set_inner(...)` would make this struct more obviously fully initialised

kcsongor (approved):

johnsaigle:
This can be merged. It's failing due to #340, I think the solution is just to re-run the failed forge test periodically until it succeeds.

PR #322, opened 2024-03-19
Initial configuration verifier implementation

mdulin2:
The goal with this code is to find bugs in deployed NTT tokens. Right now, it has the following capabilities:

* Checks decimals of manager on Solana and EVM. Probably the most important thing to get right for the security of this
* Checks the peers to ensure they line up as expected
* Quick sanity check on outbound/inbound rate limiting. I think this is bugged out from not considering decimals but it's a good start.
* Lock is only a single chain and everything else is a burn
* Threshold maxed out to the amount of transceivers
* Consistency level check for wormhole transceivers on EVM

Going forward, I want to make this more *portable* within a Docker container.

### DIFF
--- a/config_verifier/src/index.ts
+++ b/config_verifier/src/index.ts
@@ -1,0 +1,389 @@
+
+import fs from "fs";
+import { NttManager__factory } from "../../ci_tests/evm_binding/factories/NttManager__factory"
+import { WormholeTransceiver__factory } from "../../ci_tests/evm_binding/factories/WormholeTransceiver__factory"
+
+import { BigNumber, ethers, providers } from "ethers";
+import { keccak256 } from "@certusone/wormhole-sdk";
+import {NTT} from "../../solana/ts/sdk/ntt";
+import {web3 } from "@coral-xyz/anchor";
+import { BN, translateError, type IdlAccounts, Program } from '@coral-xyz/anchor'
+import { PublicKey, PublicKeyInitData } from "@solana/web3.js";
+import * as base58 from "bs58";
+
+import {
+    Layout,
+    CustomConversion,
+    encoding,
+    ChainId,
+  } from "@wormhole-foundation/sdk-base";
+import { type ExampleNativeTokenTransfers as RawExampleNativeTokenTransfers } from '../../solana/target/types/example_native_token_transfers'
+import * as spl from "@solana/spl-token";
+
+type OmitGenerics<T> = {
+    [P in keyof T]: T[P] extends Record<"generics", any>
+    ? never
+    : T[P] extends object
+    ? OmitGenerics<T[P]>
+    : T[P];
+  };
+  
+export type ExampleNativeTokenTransfers = OmitGenerics<RawExampleNativeTokenTransfers>
+export type Config = IdlAccounts<ExampleNativeTokenTransfers>['config']
+
+async function getConfig(config?: Config): Promise<Config> {
+    return config ?? await this.program.account.config.fetch(this.configAccountAddress())
+}
+
+async function calcLocation(slot, index ){
+    return BigNumber.from(keccak256(slot)).add(index);
+}
+
+type Seed = Uint8Array | string;
+function derivePda(
+    seeds: Seed | readonly Seed[],
+    programId: PublicKeyInitData
+  ) {
+    const toBytes = (s: string | Uint8Array) => typeof s === "string" ? encoding.bytes.encode(s) : s;
+    return PublicKey.findProgramAddressSync(    
+      Array.isArray(seeds) ? seeds.map(toBytes) : [toBytes(seeds as Seed)],
+      new PublicKey(programId),
+    )[0];
+    }
+
+function getChainIds(chains) {
+    let arr = [];
+    for (const index in chains){
+        const chain = chains[index]
+        arr.push(chain['chainId']);
+    }
+    return arr;
+}
+
+async function configureChains(){
+    const chainFile = fs.readFileSync(`./src/config.json`);
+    const chains = JSON.parse(chainFile.toString());
+    const chainIdList = getChainIds(chains['chains']);
+
+    var configurationData = {};
+    for (const chain of chains['chains']){
+        console.log(`Getting configuration for chain - ${chain['description']}(${chain['chainId']})`)
+        var rpc = chain['rpc'];
+        var managerAddress = chain["managerAddress"];
+        if(chain['networkType'] == 'evm'){
+            var data = await configureEvm(rpc, managerAddress);
+            data['type'] = chain['networkType'];
+            data['chainId'] = chain['chainId'];
+            configurationData[chain['chainId']] = data; 
+        }
+        else if(chain['networkType'] == 'solana'){
+            var data = await configureSolana(rpc, managerAddress, chainIdList, chain['chainId']);
+            data['chainId'] = chain['chainId'];
+            data['type'] = chain['networkType'];
+            configurationData[chain['chainId']] = data; 
+            //process.exit()
+        }
+        else {
+            console.log("Not supported!");
+        }
+    }
+
+    return configurationData;
+}
+
+async function configureSolana(rpc, managerAddress, chainIdList, chainId){
+    const managerData = {}; 
+
+    const SOL_CONNECTION = new web3.Connection(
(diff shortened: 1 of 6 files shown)


config_verifier/src/index.ts
```diff
@@ -0,0 +1,389 @@
...
+                    console.log(`ERROR: Invalid mode for chain ${target_id}`)
+                }
+
+                // Threshold check to ensure its the maximum that it can be. 
+                if(target['threshold'] < target['count']){
+                    console.log(`WARNING: NTT Manager has lower threshold than registered transceivers. ChainID - ${target_id}`); 
+                }
+
+                // Check if the rate limit is turned off or maxed out
+                if(target['outboundRateLimit'].eq(0) || target['outboundRateLimit'].eq(BigNumber.from(2).pow(256).sub(1))){
```
johnsaigle, line 327: I think you can bring the RateLimit checks down by quite a lot:

1. Solana uses u64 to store this value so this check will never trigger even if capacity is at its maximum value 
https://github.com/wormhole-foundation/example-native-token-transfers/blob/0d764bf7ffbbf81b4d51d1bfd5e51929e7fb0460/solana/programs/example-native-token-transfers/src/queue/rate_limit.rs#L8

2. I think it's safe to say that both of these cases would count as "disabled". If the rate limit is set to be the maximum possible integer value, there is no rate limit

3. I wonder if we would add another check that's in the range of 2^32 or even smaller? To me that would still be very high for e.g. a 24h period

mdulin2: Yeah! Good ideas. I'm unsure if the `rateLimit` has values that are trimmed or not. I'll confirm this. 

The 'delay' check is good! I'll add that in.

config_verifier/src/index.ts
```diff
@@ -0,0 +1,520 @@
...
+    // console.log(accounts);
+    // //console.log(accounts.length)
+    // console.log(await manager.program.account.registeredTransceiver.fetch(accounts[0].pubkey));
+    const managerConfig = await manager.getConfig();
+
+    managerData['mode'] = (managerConfig['mode'].locking == undefined ? 1 : 0); // Convert to an integer for later use
+    managerData['token'] = managerConfig['tokenProgram']
+    managerData['threshold'] = managerConfig['threshold'].valueOf() // The 'threshold' value
+    managerData['count'] = managerConfig.enabledTransceivers; // Amount of active transceivers
+    managerData['chainId'] = 1;
```
johnsaigle, line 147: It would be helpful to add a note here that this is the "wormhole chain ID" for Solana.

config_verifier/src/index.ts
```diff
@@ -0,0 +1,520 @@
...
+    managerData['duration'] = BigNumber.from(60 * 60 * 24); // Harcoded in Solana https://github.com/wormhole-foundation/example-native-token-transfers/blob/e05b5276b59cd18d674575eb39ad7d402ca7ebfa/solana/programs/example-native-token-transfers/src/queue/rate_limit.rs#L40
+
+    /*
+    This information is much different than the EVM.
+        - Manager is the same as the transceiver on Solana. Only support for a single transceiver. So, I feel safe setting the 'managerAddress'
+        - The 'emitter account' is the actual emitter that we need to use. I am getting the address of this account and using this.
+        - Consistency level is hardcoded on the Solana side to be 'finalized'. So, just hardcoding that here
+    */
+    managerData['transceivers'] = [{ 
+        address: transceiver_address_buffer, isWormhole: true, consistencyLevel: 1, nttManager: managerAddress
```
johnsaigle, line 174: So `consistencyLevel: 1` means `finalized`?

config_verifier/src/index.ts
```diff
@@ -0,0 +1,520 @@
...
+
+            var consistencyLevel = await transceiver.consistencyLevel();
+            var nttManager = await transceiver.nttManager();
+
+            // Only going to operate on a transceiver that we know. 
+            transceiverList.push({address: transceiverAddress, isWormhole: isWormhole, transceiver: transceiver, consistencyLevel: consistencyLevel, nttManager: nttManager});
+
+            break;
+        }
+        catch{} // defaults to not adding it
```
johnsaigle, line 242: Any chance you could log something here? Having a silent `try-catch` might hide some important errors

config_verifier/src/index.ts
```diff
@@ -0,0 +1,520 @@
...
+                if(target['transceivers'][0]['consistencyLevel'] == 200 || target['transceivers'][0]['consistencyLevel'] == 201){
+                     console.log(`WARNING: Non-final consistency level for transceiver: ${target['transceivers'][0]['address']}-${target_id}`);
+                }
+
+                // Check the transeiver ownership. Important for sending and receiving messages
+                if(target['transceivers'][0]['nttManager'] != target['managerAddress']){
+                    console.log(`ERROR: Transceiver inconsistent NTTManager Ownership: ${target['transceivers'][0]['address']}-${target_id}`); 
+                }
+
+                continue;
```
johnsaigle, line 286: Could you add a note here as to why the 'chain specific code' below is being skipped in this case?

config_verifier/src/index.ts
```diff
@@ -0,0 +1,520 @@
...
+                inboundRateLimitParams = (await target['manager'].program.account.inboxRateLimit.fetch(target['manager'].inboxRateLimitAccountAddress(src['chainId'])))['rateLimit'];
+            }
+            
+            // Rate limit is set to 0. Indicates that it's disabled.
+            if(inboundRateLimitParams['limit'].eq(0)){
+                console.log(`WARNING: Inbound rate limit disabled for sender chain ${target_id} and target chain ${src_id}`); 
+                continue;
+            }
+
+            // Don't feel this is necessary or proper. Depends on the use case.
```
johnsaigle, line 459: IMO this is a nice check, especially since you flag it as a warning rather than an error.

config_verifier/src/index.ts
```diff
@@ -0,0 +1,520 @@
...
+            // if(inboundRateLimitParams['limit'].eq(BigNumber.from(2).pow(64).sub(1))){
+            //     console.log(`WARNING: Inbound rate limit very high for ${target_id} and target chain ${src_id}`); 
+            // }
+        }
+    } 
+
+    // Checking if the mint/burn setup is correct
+    if(Object.keys(chainData).length -1 != burn || lock != 1){
```
johnsaigle, line 467: Isn't it possible to do only a mint-and-burn model without a locking chain? I thought this was part of the architecture at least at a certain point of the design discussions.

It might be good to add a check for this. If you don't want to complicate this PR then maybe we could add it as an Issue

config_verifier/src/index.ts
```diff
@@ -0,0 +1,520 @@
...
+                try{
+                    targetPeerAddress = await target['manager'].program.account.transceiverPeer.fetch(target['manager'].transceiverPeerAccountAddress(src['chainId']))
+                    targetPeerAddress = BigNumber.from("0x" + bufferToHex( targetPeerAddress['address']));
+                }
+                catch(err){
+                    //console.log("Error in checkWormholeTransceiver Solana:", err);
+                    targetPeerAddress = BigNumber.from(0x000000000000000000000000000000000000000000000); // Can't find the addresss, since it doesn't exist
+                }
+            }else{
+                console.log("Invaild chain type in checkTransceivers");
```
johnsaigle, line 305:
```suggestion
                console.log("Invalid chain type in checkTransceivers");
```

config_verifier/src/index.ts
```diff
@@ -0,0 +1,520 @@
...
+    managerData['chainId'] = 1;
+    
+    // https://spl.solana.com/token#example-creating-your-own-fungible-token
+    const mintInfo = await spl.getMint(
+        SOL_CONNECTION,
+        managerConfig['mint'],
+        'finalized',
+      );
+        
+    managerData['token'] = "0x" + bufferToHex(managerConfig['mint'].toBuffer());
```
johnsaigle, line 156: I think it's worth a comment here that the `0x` prefix is being added for convenience to allow comparing Solana/Eth with the same functions. 
It looks a little strange to see Solana addresses with a `0x` in front since these aren't used in the ecosystem (and in fact program addresses are not hex-encoded but use base58).

PR #326, opened 2024-03-20
evm: add support for refund address for standard relaying

### LINKED ISSUE #291: evm: Wire up refunds on standard relaying
The current implementation does not consider refunding any additional gas paid by the user for "dummy attestations" (attestations that are done without execution). This change should refund the additional gas paid by the user back to the recipient

RahulMaganti47:
Fixes #291.

### DIFF
--- a/evm/src/NttManager/ManagerBase.sol
+++ b/evm/src/NttManager/ManagerBase.sol
@@ -185,11 +185,13 @@ abstract contract ManagerBase is
             revert PeerNotRegistered(recipientChain);
         }
 
+        // push onto the stack again to avoid stack too deep error
+        bytes32 refundRecipient = refundAddress;
+
         // call into transceiver contracts to send the message
         for (uint256 i = 0; i < numEnabledTransceivers; i++) {
             address transceiverAddr = enabledTransceivers[i];
 
-            bytes32 refundRecipient = refundAddress;
             // send it to the recipient nttManager based on the chain
             ITransceiver(transceiverAddr).sendMessage{value: priceQuotes[i]}(
                 recipientChain,
--- a/evm/src/NttManager/ManagerBase.sol
+++ b/evm/src/NttManager/ManagerBase.sol
@@ -171,6 +171,7 @@ abstract contract ManagerBase is
 
     function _sendMessageToTransceivers(
         uint16 recipientChain,
+        bytes32 refundAddress,
         bytes32 peerAddress,
         uint256[] memory priceQuotes,
         TransceiverStructs.TransceiverInstruction[] memory transceiverInstructions,
@@ -187,12 +188,15 @@ abstract contract ManagerBase is
         // call into transceiver contracts to send the message
         for (uint256 i = 0; i < numEnabledTransceivers; i++) {
             address transceiverAddr = enabledTransceivers[i];
+
+            bytes32 refundRecipient = refundAddress;
             // send it to the recipient nttManager based on the chain
             ITransceiver(transceiverAddr).sendMessage{value: priceQuotes[i]}(
                 recipientChain,
                 transceiverInstructions[transceiverInfos[transceiverAddr].index],
                 nttManagerMessage,
-                peerAddress
+                peerAddress,
+                refundRecipient
             );
         }
     }
--- a/evm/src/NttManager/NttManager.sol
+++ b/evm/src/NttManager/NttManager.sol
@@ -30,6 +30,7 @@ import {ManagerBase} from "./ManagerBase.sol";
 ///  - the amount
 ///  - the recipient chain
 ///  - the recipient address
+///  - the refund address: the address to refund the user for "dummy" attestations
 ///  - (optional) a flag to indicate whether the transfer should be queued
 ///    if the rate limit is exceeded
 contract NttManager is INttManager, RateLimiter, ManagerBase {
--- a/evm/src/NttManager/NttManager.sol
+++ b/evm/src/NttManager/NttManager.sol
@@ -30,7 +30,9 @@ import {ManagerBase} from "./ManagerBase.sol";
 ///  - the amount
 ///  - the recipient chain
 ///  - the recipient address
-///  - the refund address: the address to refund the user for "dummy" attestations
+///  - the refund address: the address to which refunds are issued for any unused gas
+///    for attestations on a given transfer. If the gas limit is configured
+///    to be too high, users will be refunded the difference.
 ///  - (optional) a flag to indicate whether the transfer should be queued
 ///    if the rate limit is exceeded
 contract NttManager is INttManager, RateLimiter, ManagerBase {
--- a/evm/src/NttManager/NttManager.sol
+++ b/evm/src/NttManager/NttManager.sol
@@ -150,7 +150,8 @@ contract NttManager is INttManager, RateLimiter, ManagerBase {
         uint16 recipientChain,
         bytes32 recipient
     ) external payable nonReentrant whenNotPaused returns (uint64) {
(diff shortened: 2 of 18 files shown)


evm/src/Transceiver/WormholeTransceiver/WormholeTransceiver.sol
```diff
@@ -194,14 +195,24 @@ contract WormholeTransceiver is
             parseWormholeTransceiverInstruction(instruction.payload);
 
         if (!weIns.shouldSkipRelayerSend && _shouldRelayViaStandardRelaying(recipientChain)) {
+            // NOTE: standard relaying supports refunds. The amount to be refunded will be sent
+            // to a refundAddress specified by the client.
```
djb15, line 199: Worth being explicit in this comment that this is to the address on the destination chain. Also worth adding to the documentation I think

evm/src/Transceiver/WormholeTransceiver/WormholeTransceiver.sol
```diff
@@ -194,14 +195,24 @@ contract WormholeTransceiver is
...
+                fromWormholeFormat(getWormholePeer(destinationChain)),
                 encodedTransceiverPayload,
                 0,
-                gasLimit
+                gasLimit,
+                destinationChain,
+                fromWormholeFormat(refundRecipient)
             );
 
+            // TODO: emit the refund address
```
djb15, line 215: Address this todo pls

evm/test/IntegrationManual.t.sol
djb15, line 1: What is this new test file primarily for?

RahulMaganti47: the integration test file was getting unwieldy imo - this separates standard relaying tests from tests that deliver messages manually

mdulin2:
I wasn't sure how this works but was worried about the case of the refund failing. It seems like we handle that case quite well in the relayer already so shouldn't be a problem for NTT - https://github.com/wormhole-foundation/wormhole/blob/aa22a2b950fbbd10221c25a7e19e82e7fd688ed8/ethereum/contracts/relayer/wormholeRelayer/WormholeRelayerDelivery.sol#L413

evm/src/NttManager/ManagerBase.sol
```diff
@@ -187,12 +188,15 @@ abstract contract ManagerBase is
         // call into transceiver contracts to send the message
         for (uint256 i = 0; i < numEnabledTransceivers; i++) {
             address transceiverAddr = enabledTransceivers[i];
+
+            bytes32 refundRecipient = refundAddress;
```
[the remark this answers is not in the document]
RahulMaganti47: this is a stack depth issue and re-assignment should be fine here, although i agree a bit ugly. imo block scoping doesn't really make sense here since the `refundAddress` is passed down from the entrypoint (e.g. the lifetime of the variable is really that long).

RahulMaganti47:
> I wasn't sure how this works but was worried about the case of the refund failing. It seems like we handle that case quite well in the relayer already so shouldn't be a problem for NTT - https://github.com/wormhole-foundation/wormhole/blob/aa22a2b950fbbd10221c25a7e19e82e7fd688ed8/ethereum/contracts/relayer/wormholeRelayer/WormholeRelayerDelivery.sol#L413

yea - the standard relayer should support all the necessary functionality - we should make a note of this so that similar logic is implemented in the specialized relayer

evm/src/NttManager/NttManager.sol
```diff
@@ -30,6 +30,7 @@ import {ManagerBase} from "./ManagerBase.sol";
 ///  - the amount
 ///  - the recipient chain
 ///  - the recipient address
+///  - the refund address: the address to refund the user for "dummy" attestations
```
gator-boi, line 33: What is a "dummy" attestation?

gator-boi: Also, folks will be refunded for attestations that succeed, but the gas limit is configured to be too high. It would be nice to mention that.

evm/src/NttManager/NttManager.sol
```diff
@@ -317,6 +321,7 @@ contract NttManager is INttManager, RateLimiter, ManagerBase {
         uint256 amount,
         uint16 recipientChain,
         bytes32 recipient,
+        bytes32 refundAddress,
```
gator-boi, line 324: We should check that the `refundAddress` is not bytes32(0).

evm/src/Transceiver/WormholeTransceiver/WormholeTransceiver.sol
```diff
@@ -196,15 +197,24 @@ contract WormholeTransceiver is
...
                 encodedTransceiverPayload,
                 0,
-                gasLimit
+                gasLimit,
+                destinationChain,
+                fromWormholeFormat(refundRecipient)
             );
 
-            emit RelayingInfo(uint8(RelayingType.Standard), deliveryPayment);
+            emit RelayingInfo(uint8(RelayingType.Standard), refundAddress, deliveryPayment);
```
gator-boi, line 217: I don't think it's necessary to emit the `refundAddress` in this event since it's already included in the delivery VAA and not very relevant for the client.

evm/test/libraries/IntegrationHelpers.sol
```diff
@@ -0,0 +1,197 @@
...
+        uint16 peerChainId,
+        uint8 peerDecimals,
+        uint64 inboundLimit
+    ) internal {
+        sourceManager.setPeer(
+            peerChainId, toWormholeFormat(address(peerManagerAddr)), peerDecimals, inboundLimit
+        );
+    }
+
+    function _enableSR(WormholeTransceiver[2] memory transceivers, uint16 chainId) internal {
```
gator-boi, line 83: Should future proof this by naming it `enableEvmSR`.

RahulMaganti47: this method passes in an arbitrary chainId - more future proof with the current name imo

evm/src/Transceiver/WormholeTransceiver/WormholeTransceiver.sol
RahulMaganti47: imo it's nice to have (e.g. client can ensure that they passed in the right refund address). there's also minimal/no overhead for including this.

gator-boi: We can keep it, but what recourse can the client take if they do pass in the wrong refund address?

evm/src/NttManager/NttManager.sol
```diff
@@ -30,6 +30,9 @@ import {ManagerBase} from "./ManagerBase.sol";
 ///  - the amount
 ///  - the recipient chain
 ///  - the recipient address
+///  - the refund address: the address to which refunds are issued for any unused gas
+///    for attestations on a given transfer. Currently, the gas limit is configured
```
gator-boi, line 34: I would say, "If the gas limit is configured too high" since the integrator decides what gas limit to pass to the constructor.

evm/src/Transceiver/WormholeTransceiver/WormholeTransceiver.sol
RahulMaganti47: it's informational - not all event logs have to be actionable at the moment. (e.g. on the next transfer it may be useful information so they don't pass in the same invalid or wrong address again).

gator-boi (dismissed):

djb15 (approved):

djb15 (approved):

PR #330, opened 2024-03-21
solana: Add anchor linting to CI

johnsaigle:
Add a bash script to manually run anchor linting commands as it does not occur automatically in anchor-cli 0.29.0
Add Makefile targets that point to the new script
Add CI step to run the new make command

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,6 +34,11 @@ fix-format:
 test-evm:
 	cd evm && forge test -vvv
 
+.PHONY: anchor-lint
+anchor-lint:
+	cd solana && make anchor-lint
+
+
 # Verify that the contracts do not include PUSH0 opcodes
 test-push0:
 	forge build --extra-output evm.bytecode.opcodes
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,7 @@ test-evm:
 
 .PHONY: anchor-lint
 anchor-lint:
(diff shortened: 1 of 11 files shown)


Makefile
```diff
@@ -34,6 +34,11 @@ fix-format:
 test-evm:
 	cd evm && forge test -vvv
 
+.PHONY: anchor-lint
+anchor-lint:
+	cd solana && make anchor-lint
```
kcsongor, line 39: nit: 
```suggestion
	make -C solana anchor-lint
```
is a little more idiomatic. As an aside, I'm not sure this top-level makefile is all that useful. I think it would be better to just have `evm/Makefile` and `solana/Makefile`

kcsongor (approved):

Makefile
johnsaigle: cool, I'll update make the idiomatic update in a second PR if that's OK with you, just so we can get the actual logic bits merged.

I think that's my preference too, but while we have a top-level Makefile I figure it should probably actually do something for Solana. The main goal anyway is just to create a target that captures the actual GitHub CI commands so that they can be run locally in a convenient way.

nvsriram:
Changes merged through https://github.com/wormhole-foundation/example-native-token-transfers/pull/507

PR #344, opened 2024-03-26
SDK: Add typescript sdk definition and implementations for NTT protocol

barnjamin:
Add Ntt TS SDK
---------------
Definition of interface and layouts in `/sdk/definitons`
Evm and Solana implementations in `/sdk/evm|solana`

CI Updates
-----------
Modifies e2e CI tests to run in the context of the SDK.
Remove the `/ci_tests` directory (move `Dockerfile` and `ci.yaml` to `/sdk` and update paths)

Solana
------
Make `make anchor-test` run in the context of the SDK and add note in README about testing

It should be possible to run `make anchor-test` in the Solana directory of a fresh clone and it "just works"

TODO:
----
- [x] clean up e2e test, with nonce manager we should be able to run more concurrently
- [x] add admin method calls (init, set params) to Evm implementation
- [x] add publishing scripts
- [ ] add some build pipeline to the SDK so that running build on un-modified source is instant
- [ ] add Evm only tests like the anchor tests + scripts to actually run them


solana/Anchor.toml
```diff
@@ -19,6 +19,7 @@ cluster = "Localnet"
 wallet = "keys/test.json"
 
 [scripts]
+# test = "yarn --cwd ../sdk/solana test"
```
barnjamin, line 22: in here so you can uncomment and run the tests, removed in follow up

sdk/definitions/__tests__/layout.test.ts
```diff
@@ -0,0 +1,46 @@
...
+  wormholeTransceiverMessageLayout,
+} from "../src/nttLayout.js";
+
+const payloads = {
+  transceiver: {
+    info: ["transceiver_info_1.txt"],
+    message: ["transceiver_message_1.txt"],
+    registration: ["transceiver_registration_1.txt"],
+  },
+};
```
barnjamin, line 16: only message is parsed, others require layout definition

solana/Makefile
```diff
@@ -12,9 +12,15 @@ target/idl/example_native_token_transfers.json:
 _anchor-build:
 	@anchor build --arch sbf
 
-anchor-test: node_modules build target/idl/example_native_token_transfers.json
+anchor-test: node_modules build target/idl/example_native_token_transfers.json sdk
```
barnjamin, line 15: running `make anchor-test` is not efficient, too much gets built

I think i can put some is-built flags in to speed it up but happy to take suggestions

PR #348, opened 2024-03-26
evm: Add view for enabled transcevier infos

djb15:
The purpose of this view is to make it easier for off-chain processes to know which ids correspond to which transceivers. This is important because transceiver instructions need to be indexed per transceiver

### DIFF
--- a/evm/src/NttManager/TransceiverRegistry.sol
+++ b/evm/src/NttManager/TransceiverRegistry.sol
@@ -230,9 +230,22 @@ abstract contract TransceiverRegistry {
         return _getTransceiverBitmapStorage().bitmap;
     }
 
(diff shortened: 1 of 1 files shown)


johnsaigle (approved):

johnsaigle:
Do we need the ABI label?

mdulin2:
No tests? Would be good to ensure that this functions works correctly.

djb15:
> No tests? Would be good to ensure that this functions works correctly.

This is only a convenience view and isn't called internally. It compiles as well so I'm personally not that fussed about adding tests, but I can if you feel strongly about it I can add a test

RahulMaganti47 (approved):

PR #349, opened 2024-03-27
SDK: Update CI to test the SDK 

barnjamin:
Modifies CI tests to run in the context of the SDK.

Remove the `/ci_tests` directory but cp `Dockerfile` and `ci.yaml` and update paths

Make `make anchor-test` run in the context of the SDK and add note in README about testing

Remove codegen files from git tracking but provide scripts to generate them


sdk/evm/src/ntt.ts
```diff
@@ -56,6 +56,7 @@ export class EvmNttWormholeTranceiver<N extends Network, C extends EvmChains>
     const tx = await this.transceiver.receiveMessage.populateTransaction(
       serialize(attestation)
     );
+    console.log(tx);
```
barnjamin, line 59: oops

PR #360, opened 2024-03-27
evm: Fix quoting with transceiver instructions

djb15:


johnsaigle (approved):

mdulin2:
How does the quoting logic loop work with the standard transfer call where the instructions are manually set be a 0 length? Since we're not providing anything for the quoting, I'm guessing this would fail but haven't tested.

https://github.com/wormhole-foundation/example-native-token-transfers/blob/e05b5276b59cd18d674575eb39ad7d402ca7ebfa/evm/src/NttManager/NttManager.sol#L151

kcsongor (approved):

djb15:
> How does the quoting logic loop work with the standard transfer call where the instructions are manually set be a 0 length? Since we're not providing anything for the quoting, I'm guessing this would fail but haven't tested.
> 
> https://github.com/wormhole-foundation/example-native-token-transfers/blob/e05b5276b59cd18d674575eb39ad7d402ca7ebfa/evm/src/NttManager/NttManager.sol#L151

We have quite a few tests using that entrypoint now so I think the parsing is ok. Also double checked and we allocate an array of empty instructions that are then passed to the transceivers. For the WH transceiver an empty instruction is handled and defaults to `shouldSkipRelayerSend = false`.

PR #362, opened 2024-03-27
evm: Validate transfers early before queuing

djb15:


evm/src/NttManager/NttManager.sol
```diff
@@ -386,6 +391,28 @@ contract NttManager is INttManager, RateLimiter, ManagerBase {
         );
     }
 
+    function _validateTransfer(
+        uint16 recipientChain,
+        bytes memory transceiverInstructions
+    ) internal view {
+        if (_getPeersStorage()[recipientChain].peerAddress == bytes32(0)) {
```
johnsaigle, line 398: Another sanity check against https://github.com/wormhole-foundation/example-native-token-transfers/pull/364 (Wormhole Chain ID is canonical). This will access the correct spot, right?

Looks good otherwise

johnsaigle (approved):

mdulin2:
Is this worth the gas cost, considering we have the ability to cancel now?

djb15:
> Is this worth the gas cost, considering we have the ability to cancel now?

Yeah that's the tradeoff. Imo we shouldn't rely on the cancel flow for an obviously incorrect set of instructions/peer that could be validated earlier. Pretty bad UX if you only find out 24 hours after queuing your funds

kcsongor (approved):
the checks make sense logically. worth the extra gas?

PR #363, opened 2024-03-27
evm: Allow users to cancel outbound queued transfers

djb15:


johnsaigle (dismissed):
Should this have the ABI label?

mdulin2 (approved):
Good change. This adds fairly little risk with simple code but allows bricked queued messages to become unstuck.

kcsongor (dismissed):

PR #364, opened 2024-03-27
evm: Make Wormhole chainId the canonical chainId

djb15:
This probably needs some more discussion since it does tie us to Wormhole supported chains. However it does reduce the risk of a misconfig where an admin specified chain id isn't the same as the wormhole chain id for that chain.

We would also need to update the deployment scripts to support this change

### DIFF
--- a/evm/src/NttManager/ManagerBase.sol
+++ b/evm/src/NttManager/ManagerBase.sol
@@ -3,6 +3,7 @@ pragma solidity >=0.8.8 <0.9.0;
 
 import "wormhole-solidity-sdk/Utils.sol";
 import "wormhole-solidity-sdk/libraries/BytesParsing.sol";
+import "wormhole-solidity-sdk/interfaces/IWormhole.sol";
 
 import "../libraries/external/OwnableUpgradeable.sol";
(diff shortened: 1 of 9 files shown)


johnsaigle (changes requested):
Needs some more updates to the test suite in order to pass CI.

Do we need the ABI label for this PR?

mdulin2:
I agree, probably worth more discussion. 

From the Wormhole side, it makes sense to do this. I also have this check in my configuration checker script, for what's it's worth.

RahulMaganti47:
> This probably needs some more discussion since it does tie us to Wormhole supported chains. However it does reduce the risk of a misconfig where an admin specified chain id isn't the same as the wormhole chain id for that chain.
> 
> We would also need to update the deployment scripts to support this change

imo not sure the wormhole chain id should be set as the default. this framework is meant to be more general anyway and integrators who choose to use the non-WH supported chains will have to configure it manually anyway. defaulting to it in the contracts makes this opinionated, which i don't think this is meant to be.

djb15:
Closing this as the consensus is that we shouldn't force this coupling to Wormhole

PR #365, opened 2024-03-27
evm: Prevent registering peers on the same chain

djb15:


johnsaigle (commented):
How does this interact with https://github.com/wormhole-foundation/example-native-token-transfers/pull/364/files? Will the checks here properly validate against the Wormhole ID? Is it clear enough to users that they should be supplying a Wormhole ID here?

djb15:
> How does this interact with #364? Will the checks here properly validate against the Wormhole ID? Is it clear enough to users that they should be supplying a Wormhole ID here?

This doesn't actually have any dependencies with that PR, since the other PR also saves to the `chainId` immutable. And there is no easy way to validate the peer chainId is a valid chainId (this doesn't change whether or not we use the WH chainId as the canonical one)

mdulin2 (approved):
Seems like a nice thing to prevent. If I recall, Spearbit said that the incoming would fail anyway but don't want to lock things up.

evm/src/interfaces/INttManager.sol
```diff
@@ -123,6 +123,10 @@ interface INttManager is IManagerBase {
     /// @param sender The original sender that initiated the transfer that was queued.
     error CancellerNotSender(address canceller, address sender);
 
+    /// @notice Peer cannot be on the same chain
+    /// @dev Selector 0x20371f2a.
+    error InvalidPeerSameChainId();
```
RahulMaganti47, line 128: can you include the `peerChainId` or the `chainId` in the error?

mdulin2 (approved):

RahulMaganti47 (approved):

PR #372, opened 2024-04-01
evm: remove default addresses from sample.env

gator-boi:


RahulMaganti47 (approved):

evm/env/.env.sample
```diff
@@ -25,18 +25,18 @@ export RELEASE_OUTBOUND_LIMIT=0
...
 ### Wormhole Chain ID
-export RELEASE_WORMHOLE_CHAIN_ID=10002
+export RELEASE_WORMHOLE_CHAIN_ID=
 
 ### Wormhole Core Bridge Address
-export RELEASE_CORE_BRIDGE_ADDRESS=0x4a8bc80Ed5a4067f1CCf107057b8270E0cC11A78
+export RELEASE_CORE_BRIDGE_ADDRESS=
 
 ### Wormhole Finality
 export RELEASE_CONSISTENCY_LEVEL=201
```
djb15, line 34: Should we remove this default too? Seems dangerous to leave this at 201

djb15 (commented):
What about also removing the `RELEASE_OUTBOUND_LIMIT` default? Especially since we currently don't have the tooling to easily update rate limits?

evm/env/.env.sample
djb15: Ah wait, is 200 instant, 201 is finalized?

gator-boi:
> What about also removing the `RELEASE_OUTBOUND_LIMIT` default? Especially since we currently don't have the tooling to easily update rate limits?

I added this change to the latest commit. Thanks.

RahulMaganti47 (approved):

PR #375, opened 2024-04-02
evm: Explicitly check staticcalls

djb15:


mdulin2:
Can you check the result on this one as well - https://github.com/wormhole-foundation/example-native-token-transfers/blob/fb39daa2a6b6f5bc9289c0bc48f376880e26853c/evm/test/libraries/NttManagerHelpers.sol#L20

mdulin2 (dismissed):
Add the check in the other location and I think it's good to go.

RahulMaganti47 (dismissed):
can you add a negative test for this? otherwise, lgtm

RahulMaganti47:
> Can you check the result on this one as well -
> 
> https://github.com/wormhole-foundation/example-native-token-transfers/blob/fb39daa2a6b6f5bc9289c0bc48f376880e26853c/evm/test/libraries/NttManagerHelpers.sol#L20

might be nice to add but probably not necessary since this is in the tests (unless that error is being tested specifically)?

RahulMaganti47 (approved):

mdulin2:
> > Can you check the result on this one as well -
> > https://github.com/wormhole-foundation/example-native-token-transfers/blob/fb39daa2a6b6f5bc9289c0bc48f376880e26853c/evm/test/libraries/NttManagerHelpers.sol#L20
> 
> might be nice to add but probably not necessary since this is in the tests (unless that error is being tested specifically)?

Ahh, you're right. I didn't see this was in the tests - I just saw library and figured it was under the hood.

mdulin2 (approved):

PR #378, opened 2024-04-02
evm: Check forks on all entrypoints

djb15:
There was a very unlikely scenario whereby a source chain could fork, wormhole supports this forked chain, and the admin of the forked NTT manager updates the chain id and supports this new peer. Any queued transfers could then be played twice, once from the original chain, and once from the forked chain. You'd double your money on the destination chain.

This change adds a fork check on all the key entry and exit points:
- When queuing a transfer
- When sending both a queued and non-queued transfer
- When releasing funds on execution of a message
- When trying to cancel a queued outbound transfer


mdulin2 (dismissed):
All of the input and outputs have checks. Good work!

mdulin2 (approved):
Thanks for adding the negative test for this!

PR #381, opened 2024-04-02
evm: Check threshold invariants when adding/removing transceivers

djb15:
Note that this change prevents an admin from disabling all the transceivers from an NTT manager...there has to be at least 1 enabled transceiver at all times when `numRegisteredTransceivers > 0`.


mdulin2 (approved):

RahulMaganti47 (commented):
curious, what is the gas overhead for adding these invariant checks?

RahulMaganti47:
> > curious, what is the gas overhead for adding these invariant checks?
> 
> I haven't checked, but does it matter? It's an admin method that shouldn't be called very often so better to run the check than gas optimise imo

yea fair

RahulMaganti47 (approved):

PR #391, opened 2024-04-09
solana: upgrade to solana version 1.18.10

kcsongor:
This PR upgrades the solana sdk version from 1.17.2 to 1.18.10.
The main reason for the upgrade is in preparation for supporting token22 tokens with transfer hooks, as the new version includes several bug fixes in the on-chain CPI helper functions.

The upgrade is smooth for the most part, two minor changes were necessary:
1. Heap allocated the mint account in `initialize`. This was necessary because I was getting a stack overflow when testing a token22 mint with extensions (and the mint account's size is dynamically allocated depending on which extensions are used). As an aside, we should consider heap-allocating the mint account in other instructions too for the same reason.
2. The `ProcessInstructionWithContext` type no longer resolved in the test setup module. The argument it was referenced in wasn't used anyway, so I just removed it from the relevant function.

I will follow up with a separate PR that adds the relevant token22 modifications.

### DIFF
--- a/.github/workflows/solana.yml
+++ b/.github/workflows/solana.yml
@@ -61,8 +61,8 @@ jobs:
             ~/.cargo/bin
             ~/.cargo/registry
             ~/.cargo/git/db
-            target
-          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('Cargo.lock') }}
+            solana/target
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('solana/Cargo.lock') }}
           restore-keys: |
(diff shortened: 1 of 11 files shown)


johnsaigle (approved):

johnsaigle (changes requested):
Looks like the tests are failing. Does this interact with the `rust-toolchain` file? Maybe that needs to be updated as the error is relating to incompatible `rustc` versions.

kcsongor:
ah I believe this has to do with the solana version that's installed in CI -- I'll push a fix in a bit

johnsaigle (approved):

a5-pickle (approved):

PR #310, opened 2024-03-14
solana: Add error handling on mul_div function

johnsaigle:

PR #392, opened 2024-04-09
evm: update submodule SDK

johnsaigle:
resolves https://github.com/wormhole-foundation/example-native-token-transfers/issues/340. We thought this had been fixed before but we forgot to update the submodule. As such, the changes were made upstream but not in this repo.

~~This should help #310 to pass CI.~~ It made it through anyway!

I don't have much experience working with submodules so I would appreciate a careful review here (or suggestions).


kcsongor (approved):
This looks good (also from a submodule perspective)

nik-suri (approved):
Diff lgtm

PR #396, opened 2024-04-09
evm/src/interfaces: update topic hashes

panoel:


nik-suri (approved):
verified the topic hashes match

djb15 (approved):
Verified topic hashes

PR #400, opened 2024-04-10
solana: versioning

kcsongor:


solana/programs/example-native-token-transfers/src/lib.rs
```diff
@@ -62,6 +62,8 @@ pub const TOKEN_AUTHORITY_SEED: &[u8] = b"token_authority";
 /// user, atomically).
 pub const SESSION_AUTHORITY_SEED: &[u8] = b"session_authority";
 
+pub const VERSION: &str = "0.1.0";
```
nik-suri, line 65: This should be set to `1.0.0`

johnsaigle (commented):
LGTM, will merge once the string is updated to 1.0.0 as Nikhil mentioned.

kcsongor:
Would it be better to merge as is and update the version number in #401? Otherwise the version number would deviate from the crate versions. Not a big deal either way but seems more appropriate than to do it that way

johnsaigle (approved):

nik-suri (approved):

PR #402, opened 2024-04-11
Unilateral pause implementation

mdulin2:
From the Spearbit audit, we noticed that some edits would brick in-flight calls. So, we created a mechanism, alongside proper administration, that should solve this problem. 

There is now a way to pause all inbound or all outbound txs. This allows for *outbound* txs to be paused during a potentially bricking update for in-flight Txs but allow incoming calls. After a certain amount of time that we're confident all Txs have made it, we can change the settings. 

Currently, the inboundPause and outboundPause are on all functions where pause is on. 

The functions that must be paused in order to update, it's more complicated. I identified that `setPeer`, `removeTransceiver` and `setThreshold` all have the capability to break in flights txs. So, these are the only ones that must be outbound paused in order to allow for updates. There is currently no strict time enforcement on this - it's the job of the manager to know this. 

Open questions
* What needs to be changed in the deployment and maintenance scripts? 
* Are there more functions that should be restricted? I thought that the changing of a transceiver peer on the Wormhole transceiver was another case where this could be useful. 
* Should this feature be done *globally* or *per chain*? Right now, it's setup per chain.

### DIFF
--- a/evm/src/NttManager/ManagerBase.sol
+++ b/evm/src/NttManager/ManagerBase.sol
@@ -15,6 +15,7 @@ import "../interfaces/ITransceiver.sol";
 import "../interfaces/IManagerBase.sol";
 
 import "./TransceiverRegistry.sol";
+import "forge-std/console.sol"; // TODO - remove this - MAX
 
 abstract contract ManagerBase is
     IManagerBase,
@@ -42,6 +43,27 @@ abstract contract ManagerBase is
         deployer = msg.sender;
     }
 
+    modifier inboundNotPaused(){
+        if(_getUnilateralPauseStorage().inbound){
+            revert InboundPaused();
+        }
+        _;
+    }
+
+    modifier outboundNotPaused(){
+        if(_getUnilateralPauseStorage().outbound){
+            revert OutboundPaused();
+        }
+        _;
+    }
+
+    modifier outboundWhenPaused(){
+        if(_getUnilateralPauseStorage().outbound == false){
+            revert NotPausedForUpdate();
+        }
+        _;
+    }
+
     function _migrate() internal virtual override {
         _checkThresholdInvariants();
         _checkTransceiversInvariants();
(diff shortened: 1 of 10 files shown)


evm/src/NttManager/ManagerBase.sol
```diff
@@ -42,6 +43,27 @@ abstract contract ManagerBase is
...
+
+    modifier outboundNotPaused() {
+        if (_getUnilateralPauseStorage().outbound) {
+            revert OutboundPaused();
+        }
+        _;
+    }
+
+    modifier outboundWhenPaused() {
+        if (_getUnilateralPauseStorage().outbound == false) {
```
djb15, line 61: Nit pick but prefer `!` vs `== false`

evm/src/NttManager/NttManager.sol
```diff
@@ -241,7 +243,12 @@ contract NttManager is INttManager, RateLimiter, ManagerBase {
     }
 
     /// @inheritdoc INttManager
-    function completeInboundQueuedTransfer(bytes32 digest) external nonReentrant whenNotPaused {
+    function completeInboundQueuedTransfer(bytes32 digest)
+        external
+        nonReentrant
+        whenNotPaused
+        inboundNotPaused
```
djb15, line 250: I think we can safely complete here without pausing on inbound transfers? Or do you think it's better to be aggressive on the pausing here?

evm/src/NttManager/ManagerBase.sol
```diff
@@ -372,7 +419,7 @@ abstract contract ManagerBase is
     }
 
     /// @inheritdoc IManagerBase
-    function setThreshold(uint8 threshold) external onlyOwner {
+    function setThreshold(uint8 threshold) external onlyOwner outboundWhenPaused {
```
djb15, line 422: Isn't the threshold of the local chain more concerned around receiving inbound messages? For example a deployment may want to change the threshold on one chain only, and they would only have to pause the outbound transfer of that chain, not every other chain that sends to it?

djb15 (commented):
I think generally this makes sense and does what we discussed. Having read the code, unless I'm misunderstanding, I still think this relies quite a lot on documentation for operators on how to coordinate these pausing transactions across chain, do you agree?

If this approach needs documentation and specific instructions for operators too, do you think this sufficiently solves a problem for the added complexity?

PR #404, opened 2024-04-12
solana: support tokens with transfer hooks

kcsongor:
This PR adds support for token2022 tokens with transfer hooks. For more details on transfer hooks, see https://solana.com/developers/guides/token-extensions/transfer-hook.

At a high level, a transfer hook is a program that is called (by the token program) each time a transfer (of a token of a specific mint) is made. The key difficulty is that the transfer hook program may require additional accounts, which need to be propagated through the instruction.

This PR modifies the on-chain code to call the appropriate helper function that propagates the necessary accounts, and also the off-chain code (ts sdk) that derives these account addresses based on the on-chain schema standard that token2022 introduces for transfer hooks. The rust tests (`cargo test-sbf`) still test against the legacy spl token, while in this PR I modified the typescript tests to test against a token with a transfer hook. This way, both paths are tested appropriately.

There is a dummy transfer hook in the test that bumps a counter on each transfer, and has a dummy account seeded by the owner field of the sender token account, which is just used to test that the account information is properly propagated.

Adding this was the motivation for #391, because the on-chain helper code is simply completely unusable (broken) in earlier sdk versions.

# ABI changes
## Breaking changes
This change is breaking, so it warrants a major ABI version upgrade (bumped the version to 2.0.0). This is because we now have to perform an extra transfer when burning and minting tokens in order to trigger any transfer hooks. This extra transfer in turn requires some additional accounts -- refer to the documentation in the `transfer_burn` and `release_inbound_mint` functions for more details.

## New features
Also added an extra set of instructions to programmatically create/manage address lookup tables. These were necessary to add extra transaction size headroom for transfer hooks to be triggered. Without a lookup table, a transfer hook could use up to 2 custom accounts, but now there is space for 15+. See the documentation in `lut.rs` for more details.


johnsaigle (changes requested):
No opposition to the changes so far. Looks like this needs a bit of work to get past CI

kcsongor:
Hm, interesting. `anchor test` passes for me locally, I can't reproduce the CI failure here (which seems to be deterministic) :(

johnsaigle:
@kcsongor I'm getting the same issue as CI when I run `make anchor-test` locally.

Maybe check your local CLI tool versions? Here are mine:
```
avm 0.29.0
solana-cli 1.18.10 (src:50b3f337; feat:3469865029, client:SolanaLabs)
```

kcsongor:
```
avm 0.29.0
solana-cli 1.18.10 (src:50b3f337; feat:3469865029, client:SolanaLabs)
anchor-cli 0.30.0
```
for me

johnsaigle:
I have `anchor-cli 0.29.0`. Maybe you could try downgrading?

kcsongor:
Downgraded to 2.90, it still works. But I managed to reproduce this on my x86 ubuntu machine (it's working on arm64/macOS). What arch/OS are you running on?

johnsaigle:
I haven't done an in-depth review yet but I've assigned myself and will return to it later today or this week.

Quick note: Are the changes to the Solidity files meant to be here or are they the result of the rebase? I'm unclear how they're related to the changes for the Solana code.

solana/programs/example-native-token-transfers/src/instructions/release_inbound.rs
```diff
@@ -119,8 +120,8 @@ pub struct ReleaseInboundUnlock<'info> {
 /// Setting this flag to `false` is useful when bundling this instruction
 /// together with [`crate::instructions::redeem`], so that the unlocking
 /// is attempted optimistically.
-pub fn release_inbound_unlock(
-    ctx: Context<ReleaseInboundUnlock>,
+pub fn release_inbound_unlock<'info>(
+    ctx: Context<'_, '_, '_, 'info, ReleaseInboundUnlock<'info>>,
```
johnsaigle, line 124: Is it worth documenting these lifetimes? (These are lifetimes, right?)
I'm sure it's fine but I'm not sure why they are required

solana/programs/example-native-token-transfers/src/instructions/release_inbound.rs
```diff
@@ -138,22 +139,22 @@ pub fn release_inbound_unlock(
...
-                    mint: ctx.accounts.common.mint.to_account_info(),
-                },
+        Mode::Locking => {
+            onchain::invoke_transfer_checked(
+                &ctx.accounts.common.token_program.key(),
+                ctx.accounts.custody.to_account_info(),
+                ctx.accounts.common.mint.to_account_info(),
+                ctx.accounts.common.recipient.to_account_info(),
+                ctx.accounts.common.token_authority.clone(),
+                ctx.remaining_accounts,
```
johnsaigle, line 149: So I see this is how the API is intended to be used - but, what accounts are we intending to pass along here? Multi-sigs?

`remaining_accounts` is considered to be semi-dangerous because it jettisons Anchor's usual checks on these accounts. If we use it we should be clear about why and why it's safe.

See:
https://docs.rs/anchor-lang/latest/anchor_lang/context/struct.Context.html#structfield.remaining_accounts
https://www.zellic.io/blog/the-vulnerabilities-youll-write-with-anchor/#misuse-of-anchors-ctxremaining_accounts

solana/ts/sdk/ntt.ts
```diff
@@ -891,15 +958,98 @@ export class NTT {
...
     }
   }
 }
 
 function exhaustive<A>(_: never): A {
   throw new Error('Impossible')
 }
+
+/**
+ * TODO: this is copied from @solana/spl-token, because the most recent released
```
johnsaigle, line 975: Might be good to also capture this in a GH Issue so it doesn't get lost

johnsaigle (commented):
A few questions about the API. Will approve after we discuss.

solana/programs/example-native-token-transfers/src/instructions/release_inbound.rs
kcsongor: `Context` has 4 lifetime parameters, and `ReleaseInboundUnlock` has 1. When you don't explicitly write out the lifetimes (like in the original code), the compiler will introduce fresh lifetime variables for them, so behind the scene the original code is translated to
```
pub fn release_inbound_unlock<'a, 'b, 'c, 'info, 'e>(
     ctx: Context<'a, 'b, 'c, 'info, ReleaseInboundUnlock<'e>>,
```
(here I used `'info` by convention, the compiler would not use that name internally)

below, when we call `invoke_transfer_checked`, it has the following signature
https://github.com/solana-labs/solana-program-library/blob/9375e9642be4050ea32a4d2112c838d068f1a370/token/program-2022/src/onchain.rs#L20-L30
```rust
pub fn invoke_transfer_checked<'a>(
    token_program_id: &Pubkey,
    source_info: AccountInfo<'a>,
    mint_info: AccountInfo<'a>,
    destination_info: AccountInfo<'a>,
    authority_info: AccountInfo<'a>,
    additional_accounts: &[AccountInfo<'a>],
    amount: u64,
    decimals: u8,
    seeds: &[&[&[u8]]],
) -> ProgramResult {
```
notice that it expects all of the inputs to have the same lifetime (`'a`). Also notice that it has some unannotated `&`s like the slice of `additional_accounts`. These get the same treatment as before, i.e. the compiler introduces unique lifetime variables for them that are distinct from anything else.

`Context` in anchor is defined as https://github.com/coral-xyz/anchor/blob/852fcc77beb6302474a11e0f8e6f1e688021be36/lang/src/context.rs#L24-L37
```rust
pub struct Context<'a, 'b, 'c, 'info, T: Bumps> {
    /// Currently executing program id.
    pub program_id: &'a Pubkey,
    /// Deserialized accounts.
    pub accounts: &'b mut T,
    /// Remaining accounts given but not deserialized or validated.
    /// Be very careful when using this directly.
    pub remaining_accounts: &'c [AccountInfo<'info>],
    /// Bump seeds found during constraint validation. This is provided as a
    /// convenience so that handlers don't have to recalculate bump seeds or
    /// pass them in as arguments.
    /// Type is the bumps struct generated by #[derive(Accounts)]
    pub bumps: T::Bumps,
}
```
now if we substitute `ReleaseInboundUnlock<'e>` for `T` we can see that the lifetime of `ctx.accounts` mentions `'b` and `'e` while `ctx.remaining_accounts` mentions `'c` and `'info`. The `'b'` and `'c'` parts are fine, because these are ignored in `invoke_transfer_checked` above (unnamed lifetimes), but `e` and `'info` must be the same. So I made that change by introducing an explicit lifetime `'info` and syntactically restricting those two to be the same.
The code doesn't typecheck without this, not sure if we need to add documentation into the code, as it doesn't really have any impact on operational semantics.

solana/programs/example-native-token-transfers/src/instructions/release_inbound.rs
kcsongor: `remaining_accounts` is dangerous when used directly because no account checks have been made by our program. In this case, these accounts are just passed on directly to the token2022 program, which in turn will pass them to the transfer hook, which is responsible for validating all the usual things about it (in practice the transfer hook will be an anchor program).
So there's no real security impact, as we're not using those accounts directly.

The accounts that are actually passed in are computed by the client here https://github.com/wormhole-foundation/example-native-token-transfers/blob/29fb437c3dbae0e5f018731e655f54f73660a1b5/solana/ts/sdk/ntt.ts#L438-L445 by consulting the transfer hook program's `extra-accounts-meta` PDA. The `invoke_transfer_checked` function then does some verificaiton on these accounts, and ultimately passes them to the token2022 program.

solana/ts/sdk/ntt.ts
kcsongor: opened an issue https://github.com/wormhole-foundation/example-native-token-transfers/issues/417 thanks!

johnsaigle (approved):

kcsongor:
Added a new commit which makes some non-trivial changes. tldr we now perform a transfer before burning on the way out, and a transfer after minting on the way in. This is done in order to trigger the transfer hooks of the underlying token, which minting and burning don't trigger.

kev1n-peters:
Should we bump `pub const VERSION: &str = "1.0.0";` in this PR since there's an ABI change?

kcsongor:
> Should we bump `pub const VERSION: &str = "1.0.0";` in this PR since there's an ABI change?

yes! I'll do that

nonergodic:
I'm still wrapping my head around things, but what sticks out to me so far is that this has implications for the specialized relayer into Solana. It now has to fetch and parse the extra-account-metas pda so it can pass these accounts in as "remaining accounts". Besides that, I'm wondering if there are any security implications that make the relayer exploitable in some way. It definitely adds an additional, unknown burden in terms of the compute budget.

solana/programs/example-native-token-transfers/src/instructions/release_inbound.rs
```diff
@@ -40,6 +41,13 @@ pub struct ReleaseInbound<'info> {
     pub mint: InterfaceAccount<'info, token_interface::Mint>,
 
     pub token_program: Interface<'info, token_interface::TokenInterface>,
+
+    /// CHECK: the token program checks if this indeed the right authority for the mint
+    #[account(
+        mut,
+        address = config.custody
+    )]
+    pub custody: InterfaceAccount<'info, token_interface::TokenAccount>,
```
nonergodic, line 50: Given this change and seeing how custody is now used both in locking and in mint/burn mode, the [comment in initialize](https://github.com/wormhole-foundation/example-native-token-transfers/blob/bf49cabe9f539fbf284925ab81d1e28722b75050/solana/programs/example-native-token-transfers/src/instructions/initialize.rs#L71-L72) about it not being used in the latter is now outdated.

nonergodic (approved):
Why go through custody when burning/minting? Presumably to give the hooks a chance to trigger, but if that's the intended behavior, then burn/mint should also support hooks and it would be an oversight in the hook implementation, no?

Besides that, everything looks good to me besides the one outdated comment. Excellent documentation/comments!

solana/programs/example-native-token-transfers/src/instructions/release_inbound.rs
kcsongor: good catch, updated the comment

kcsongor:
> Why go through custody when burning/minting? Presumably to give the hooks a chance to trigger, but if that's the intended behavior, then burn/mint should also support hooks and it would be an oversight in the hook implementation, no?

That's right, added comments in the code and updated the PR description to explain this. Indeed, arguably this is an oversight in the Token2022 program (the fact that transfer hooks are only triggered on transfers, but not on mints and burns).

nonergodic (dismissed):

johnsaigle (changes requested):
Still trying to work in some time to give this the thorough review that it deserves. I'm putting the red X here so that it doesn't get merged before I have a change to review, hope that's OK 🙏🏻

solana/programs/example-native-token-transfers/src/instructions/release_inbound.rs
```diff
@@ -135,25 +173,19 @@ pub fn release_inbound_unlock(
         }
     }
 
-    assert!(inbox_item.release_status == ReleaseStatus::Released);
```
johnsaigle, line 138: Note to self/other reviews: 
Deleting this is OK because `try_release()` ensures that a return value of `true` implies `ReleaseStatus::Released` and vice-versa. 
The `if !released` check above is therefore logically equivalent to this `assert` and so keeping it here is redundant.

solana/programs/example-native-token-transfers/src/instructions/release_inbound.rs
```diff
@@ -135,25 +173,19 @@ pub fn release_inbound_unlock(
...
-    match ctx.accounts.common.config.mode {
-        Mode::Burning => Err(NTTError::InvalidMode.into()),
```
johnsaigle, line 140: Note to self/other reviews: 
Deleting this is OK because it's guaranteed by the new constraint on the composite account `ReleaseInboundUnlock`

solana/programs/example-native-token-transfers/src/instructions/release_inbound.rs
```diff
@@ -78,38 +89,65 @@ pub fn release_inbound_mint(
...
+                mint: ctx.accounts.common.mint.to_account_info(),
+                to: ctx.accounts.common.custody.to_account_info(),
+                authority: ctx.accounts.common.token_authority.clone(),
+            },
+            &[&[
+                crate::TOKEN_AUTHORITY_SEED,
+                &[ctx.bumps.common.token_authority],
+            ]],
         ),
-        Mode::Locking => Err(NTTError::InvalidMode.into()),
```
johnsaigle, line 97: Note to self/other reviews: 
Deleting this is OK because it's guaranteed by the new constraint on the composite account `ReleaseInboundMint`

solana/programs/example-native-token-transfers/src/instructions/release_inbound.rs
```diff
@@ -78,38 +89,65 @@ pub fn release_inbound_mint(
     }
 
     assert!(inbox_item.release_status == ReleaseStatus::Released);
```
johnsaigle, line 91: @kcsongor I think you can delete this `assert` as well given the behaviour of `try_release`. It looks a bit funny to have it in one path/mode but not the other. I think we should do both or neither.

It can be done in a separate cleanup PR though. We don't need to derail this PR.

solana/programs/example-native-token-transfers/src/instructions/transfer.rs
```diff
@@ -51,6 +66,15 @@ pub struct Transfer<'info> {
     #[account(mut)]
     pub outbox_rate_limit: Account<'info, OutboxRateLimit>,
 
+    #[account(
+        mut,
+        address = config.custody
+    )]
+    /// Tokens are always transferred to the custody account first regardless of
+    /// the mode.
+    /// For an explanation, see the note in [`transfer_burn`].
```
johnsaigle, line 75: or `transfer_lock`, I suppose

solana/ts/sdk/ntt.ts
```diff
@@ -349,22 +474,58 @@ export class NTT {
...
+      await addExtraAccountMetasForExecute(
+        this.program.provider.connection,
+        transferIx,
+        transferHook.programId,
+        source,
+        mint,
+        destination,
+        owner,
+        // TODO(csongor): compute the amount that's passed into transfer.
+        // Leaving this 0 is fine unless the transfer hook accounts addresses
```
johnsaigle, line 520: It would be nice to have a test to check the error condition when token with transfer fee is used too. (the `before != after` condition in both transfer paths)

johnsaigle (commented):
Added some notes, mostly to confirm that removing some of the checks should be safe. I like the Anchor constraint changes for the Mode checks and the comment additions quite a lot.

solana/programs/example-native-token-transfers/src/instructions/release_inbound.rs
kcsongor: yeah good catch. I think maybe doing it in both cases is fine, even though it's a redundant check (that's why it's an assert with no custom error, as it should never be hit)

solana/programs/example-native-token-transfers/src/instructions/transfer.rs
kcsongor: this is phrased a little awkwardly -- my intention here was to say something like
"For an explanation of why the custody account is needed even in burning mode, see the note in [`transfer_burn`]" as that's the less intuitive case (and indeed that's the only place where there is a note to this effect)

solana/programs/example-native-token-transfers/src/instructions/luts.rs
```diff
@@ -0,0 +1,172 @@
...
+        &ix,
+        &[
+            ctx.accounts.lut_address.to_account_info(),
+            ctx.accounts.authority.to_account_info(),
+            ctx.accounts.payer.to_account_info(),
+            ctx.accounts.system_program.to_account_info(),
+        ],
+    )?;
+
+    let entries_infos = ctx.accounts.entries.to_account_infos();
```
johnsaigle, line 148: Would there be any benefit (maybe CU savings?) to failing early on this line? The program could check whether entries_info.len() + 1  exceeds 256 (max LUT size) instead of doing the CPI. We know that the CPI will fail in this case so we may as well not call itIf so, it could even be moved above the first invoke.

The only downside that comes to mind is wiring the LUT max size constant into the program, but maybe this could be imported from the Solana LUT crate. I imagine it's not likely to change either, but not sure

johnsaigle (dismissed):
Looks good on my side from a security perspective. Csongor and I discussed some ideas on Slack and I'm comfortable with the state of the PR.

johnsaigle (approved):

nonergodic (approved):
no additional changes since my last review from what I can tell

the only thing that stuck out to me when taking a second glance is that the ntt quoter version also got bumped to version 2.0 even though for the ntt quoter itself nothing actually changed, which might be slightly confusing ("What's the difference between ntt-quoter 1.x and 2.0?" "No difference.") though ofc not bumping it is maybe also bad ("Is this the right version of the quoter for that version of NTT?").

PR #413, opened 2024-04-18
SDK: Add Routes and Examples

barnjamin:
Fixes the `sed` hack for watched accountant emitters that I'd previously updated incorrectly

Adds Relayer testing back to CI (for EVM only)

Adds `examples` directory with a couple examples of manual transfer init/redemption

Adds `route` package with `ManualRoute` and `AutomaticRoute` implementation

Adds Solana quoter class (copied with edits from https://github.com/wormhole-foundation/wormhole-connect/blob/development/wormhole-connect/src/routes/ntt/chains/solana/nttQuoter.ts)

Add method `getIsApproved`, `getIsExecuted` and `quoteDeliveryPrice` to NTT interface and implement them for each platform


sdk/__tests__/utils.ts
```diff
@@ -431,9 +470,7 @@ async function deployEvm(ctx: Ctx): Promise<Ctx> {
 
   // Setup the initial calls, like transceivers for the manager
   console.log("Set transceiver for manager");
-  await tryAndWaitThrice(() =>
-    transceiver.getAddress().then((addr) => manager.setTransceiver(addr))
-  );
+  await tryAndWaitThrice(() => manager.setTransceiver(transceiverProxyAddress));
```
barnjamin, line 473: This has been changed form the transceiver address to the proxy address. 

I think this is correct but I'm not positive

sdk/__tests__/utils.ts
```diff
@@ -253,6 +269,38 @@ export async function transferWithChecks(
   );
 }
 
+async function waitForRelay(
+  msgId: WormholeMessageId,
+  dst: Ctx,
+  retryTime: number = 2000
+) {
+  console.log("Sleeping for 1 min to allow signing of VAA");
+  await new Promise((resolve) => setTimeout(resolve, 60 * 1000));
```
barnjamin, line 278: Would like to find a way to set the consistency level for the relayer to instant so this does not need to wait for a full minute before checking

sdk/route/__tests__/route.test.ts
```diff
@@ -0,0 +1,239 @@
...
+  });
+
+  let vp: routes.ValidationResult<typeof op>;
+  it("Should validate a transfer request", async function () {
+    vp = await found.validate({ amount: "1.0", options: op });
+    expect(vp.valid).toBeTruthy();
+    expect(vp.params.amount).toEqual("1.0");
+  });
+
+  let qr: Awaited<ReturnType<typeof found.quote>>;
```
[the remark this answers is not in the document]
barnjamin: they're used again in the next step so wanted them to be outside the single `it` scope

sdk/route/src/automatic.ts
```diff
@@ -0,0 +1,254 @@
...
+    return chain.supportsProtocol("Ntt");
+  }
+
+  getDefaultOptions(): Op {
+    return NttRoute.AutomaticOptions;
+  }
+
+  async isAvailable(): Promise<boolean> {
+    // TODO: check that both src/dst are available for relayed NTT transfers
+    return true;
```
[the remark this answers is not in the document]
barnjamin: Need to actually implement this oops

sdk/evm/src/ntt.ts
```diff
@@ -128,6 +128,16 @@ export class EvmNtt<N extends Network, C extends EvmChains>
     ];
   }
 
+  async isRelayingAvailable(destination: Chain): Promise<boolean> {
+    const enabled = await Promise.all(
+      this.xcvrs.map(async (x) => {
+        return await x.isWormholeRelayingEnabled(destination);
```
kev1n-peters, line 134: We should also check if special relaying is enabled

artursapek (approved):

kev1n-peters (approved):

PR #418, opened 2024-04-18
solana: add recovery flow

kcsongor:


solana/programs/example-native-token-transfers/Cargo.toml
```diff
@@ -9,15 +9,17 @@ crate-type = ["cdylib", "lib"]
 name = "example_native_token_transfers"
 
 [features]
-default = ["mainnet"]
+default = ["owner-recovery", "mainnet"]
```
kcsongor, line 12: should remove this by default (mainnet too)

tonyjin:
We've had some discussions about whether or not this should be included, and are currently thinking that this functionality could be provided to integrators directly instead of existing in the main reference repo.

PR #426, opened 2024-04-23
solana: fix governance action byte and add serde test from guardian

kcsongor:
This PR fixes the width of the governance action (from the inferred usize to u8). Also, extracted the body serialiser code for governance packets, as that will be needed to generate actual governance proposals.

### DIFF
--- a/solana/programs/wormhole-governance/src/instructions/governance.rs
+++ b/solana/programs/wormhole-governance/src/instructions/governance.rs
@@ -109,6 +109,62 @@ impl GovernanceMessage {
         0x50, 0x75, 0x72, 0x70, 0x6F, 0x73, 0x65, 0x47, 0x6F, 0x76, 0x65, 0x72, 0x6E, 0x61, 0x6E,
         0x63, 0x65,
     ];
+
+    pub fn read_body<R>(reader: &mut R, governance_program_id: Pubkey) -> io::Result<Self>
+    where
+        R: io::Read,
+    {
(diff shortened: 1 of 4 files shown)


nik-suri (dismissed):

nik-suri (approved):

solana/programs/wormhole-governance/src/instructions/governance.rs
```diff
@@ -109,6 +109,62 @@ impl GovernanceMessage {
...
         0x63, 0x65,
     ];
+
+    pub fn read_body<R>(reader: &mut R, governance_program_id: Pubkey) -> io::Result<Self>
+    where
+        R: io::Read,
+    {
+        let program_id: Pubkey = Pubkey::new_from_array(Readable::read(reader)?);
+        let accounts_len: u16 = Readable::read(reader)?;
+        let mut accounts = Vec::with_capacity(accounts_len as usize);
```
johnsaigle, line 119: I'm wondering if there is a case here where the account length could be longer than the number of actual accounts. In this case the reader could crash if it tries to read beyond the actual bounds of the message.

This may be completely unlikely - I'm not sure if it would be possible for someone to spoof this message of if it could reach this code without being well-formed. Also, a panic might be fine here - I don't think there is a safe way to proceed in this scenario.

kcsongor: this code will panic in that case, but we assume the guardians only sign sensible inputs in general

johnsaigle (approved):

PR #430, opened 2024-04-24
Root workspace

barnjamin:


barnjamin:
taking this into https://github.com/wormhole-foundation/example-native-token-transfers/pull/433

PR #440, opened 2024-04-30
evm: make TrimmedAmount fuzz tests more efficient (fixes #431)

### LINKED ISSUE #431: Replace `vm.assume` with `bound` in fuzz tests
See test failure: https://github.com/wormhole-foundation/example-native-token-transfers/actions/runs/8836029057/job/24261632577?pr=399

From a contributor:
> It means the test threw out too many fuzz inputs for a particular test parameter. We should look into replacing vm.assume with bound if we can.

kcsongor:
The code that tests operations on trimmed amounts previously made a
`vm.assume` call to enforce that the decimals of both amounts are the
same.

Since the decimal is an 8 bit value, the probability that
two (uniformly) randomly generated decimals are the same is 1/256,
meaning that the vast majority of test cases were discarded.

Now, we instead set the decimals of the first number to the decimals of
the second number. This does not change* the characteristics of the test
cases, as the amounts are sampled independently from the decimals.

*NOTE: this is only true under the assumption that the TrimmedAmount
type (which is backed by a uint72) is actually sampled from a uniform
distribution. This might not be the case depending on the fuzzer
implementation, as some fuzzers prefer certain edge cases. Nevertheless,
for most intents and purposes, it should be close enough without
significantly weakening the test cases.

### DIFF
--- a/evm/src/libraries/TrimmedAmount.sol
+++ b/evm/src/libraries/TrimmedAmount.sol
@@ -106,6 +106,14 @@ library TrimmedAmountLib {
         return uint8(TrimmedAmount.unwrap(a) & 0xFF);
     }
 
+    /// @dev Set the decimals of the TrimmedAmount.
+    ///      This function should only be used for testing purposes, as it
+    ///      should not be necessary to change the decimals of a TrimmedAmount
+    ///      under normal circumstances.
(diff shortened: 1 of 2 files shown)


djb15 (approved):
Neat!

RahulMaganti47 (approved):
nice!

PR #442, opened 2024-05-03
SDK: Solana TS deduplication

barnjamin:
Convert NTT class to namespace and treat it as a library of utility functions to get instructions or transactions created

Delete duplicated logic for creating instructions or transactions


kev1n-peters (approved):

solana/ts/lib/ntt.ts
```diff
@@ -129,115 +167,136 @@ export class NTT {
...
+    program: Program<NttBindings.NativeTokenTransfer<IdlVersion>>,
+    config: NttBindings.Config<IdlVersion>,
+    args: {
+      payer: PublicKey;
+      wormholeId: PublicKey;
+    },
+    pdas?: Pdas
+  ) {
+    // if the program is at version 1.0.0, we don't need to initialize the LUT
+    if (program.idl.version === "1.0.0") return;
```
kcsongor, line 270: we should introduce a more sophisticated way of checking versions than direct string matching. Here, I think we should check that the major version is at least 2 -- the rationale is that if there is a security fix that necessitates a patch bump, or an external deployment that adds some tag to the version (like `1.0.0+foo`) then it would be handled gracefully

solana/ts/lib/ntt.ts
```diff
@@ -259,12 +318,10 @@ export class NTT {
...
 
-    var existingLut: web3.AddressLookupTableAccount | null = null;
+    let existingLut: web3.AddressLookupTableAccount | null = null;
     try {
-      existingLut = await this.getAddressLookupTable(false);
-    } catch {
-      // swallow errors here, it just means that lut doesn't exist
-    }
+      existingLut = await getAddressLookupTable(program, pdas);
+    } catch {}
```
kcsongor, line 324: this catchall situation makes me a little uneasy -- maybe we should just change `getAddressLookupTable` to return `web3.AddressLookupTableAccount | null` (so genuine exceptions are propataged)

solana/ts/lib/utils.ts
```diff
@@ -159,3 +97,15 @@ export const quoterAddresses = (programId: PublicKeyInitData) => {
     registeredNttAccount,
   };
 };
+
+//   // The `translateError` function expects this format, but the idl gives us a
+//   // different one, so we preprocess the idl and store the expected format.
+//   // NOTE: I'm sure there's a function within anchor that does this, but I
+//   // couldn't find it.
+//   private processErrors(): Map<number, string> {
```
kcsongor, line 105: should we delete this?

kcsongor (approved):
looks great, just left a couple minor comments

solana/ts/lib/utils.ts
barnjamin: ah yes, not sure if you wrote the note initially but I did find the correct method to use for processing the errors 

https://github.com/wormhole-foundation/example-native-token-transfers/blob/sdk/solana-dedupe/solana/ts/lib/ntt.ts#L199-L206

solana/ts/lib/ntt.ts
barnjamin: should we allow non-numeric values like `0+foo` for the patch version? instead of just bumping the patch version

PR #446, opened 2024-05-22
evm: allow calling pause() with two Guardian signatures

nik-suri:


evm/src/wormhole/Governance.sol
```diff
@@ -7,6 +7,9 @@ import "wormhole-solidity-sdk/libraries/BytesParsing.sol";
 contract Governance {
     using BytesParsing for bytes;
 
+    // Only 2 Guardian signatures are required for quorum to call the pause function on governed contracts.
+    uint PAUSER_QUORUM = 2;
```
kcsongor, line 11: let's make this a constant?

evm/src/wormhole/Governance.sol
```diff
@@ -115,11 +120,23 @@ contract Governance {
...
+        IWormhole.VM memory vm = wormhole.parseVM(encodedVM);
+        GeneralPurposeGovernanceMessage memory message =
+            parseGeneralPurposeGovernanceMessage(vm.payload);
 
-        if (!valid) {
-            revert(reason);
+        bytes memory pauseSig = abi.encodeWithSignature("pause()");
+        if (keccak256(message.callData) == keccak256(pauseSig)) {
+            // If we're calling the pause() function, only require 2 Guardian signatures
+            (bool valid, string memory reason) = _verifyVMForPause(vm, true);
```
kcsongor, line 130:
```suggestion
            (bool valid, string memory reason) = _verifyVMForPause(vm, false);
```
don't need to re-validate the hash, as it came out of `parseVM` directly -- in fact we could delete that branch from `verifyVMForPause` entirely

evm/test/wormhole/Governance.t.sol
```diff
@@ -215,4 +220,25 @@ contract GovernanceTest is Test {
 
         assert(myContract.governanceStuffCalled());
     }
-}
+
+    function test_pause_single_guardian() public {
+        // build VAA to call pause
+        // should go through smoothly (requiredQuorum == 1)
```
kcsongor, line 226: looks like these tests are still missing -- maybe put the PR back into draft?

kcsongor (changes requested):

PR #451, opened 2024-06-07
add solana signer interface and implementation for local private key

barnjamin:
An integrator with no access to a ledger may not be able to deploy the Solana contracts

This makes the `getSigner` function allow _either_ the ledger derivation path _or_ an environment variable `SOLANA_PRIVATE_KEY` to be used to create a signer

### DIFF
--- a/solana/ts/scripts/env.ts
+++ b/solana/ts/scripts/env.ts
@@ -1,12 +1,10 @@
 import fs from "fs";
-import { Connection, Commitment } from "@solana/web3.js";
-import { ChainId } from "@certusone/wormhole-sdk";
+import { Connection, Commitment, Keypair, Transaction } from "@solana/web3.js";
+import { ChainId, sign } from "@certusone/wormhole-sdk";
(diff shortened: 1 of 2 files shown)


ben-ambient (commented):
We should also update the 'helpers.ts' file, with the 'ledgerSignAndSend' function

solana/ts/scripts/env.ts
```diff
@@ -1,27 +1,65 @@
...
+  async signMessage(message: Buffer): Promise<Buffer> {
+    return Buffer.from(
+      nacl.sign.detached(new Uint8Array(message), this.keypair.secretKey)
+    );
+  }
+
+  async signTransaction(transaction: Buffer): Promise<Buffer> {
+    const tx = Transaction.from(transaction);
+    tx.sign(this.keypair);
+    return tx.serialize();
```
merlin-ambient, line 40: Replace this with:

`return this.signMessage(transaction)`

ben-ambient:
update function names in 'helpers.ts' :

ledgerSignAndSend -> signAndSend
addLedgerSignature -> addSignature

PR #466, opened 2024-06-28
sdk: add missing admin functionality and queries

kcsongor:
This PR adds some missing functionality to the sdk, mostly related to admin functions


sdk/evm/src/ntt.ts
```diff
@@ -50,17 +51,48 @@ export class EvmNttWormholeTranceiver<N extends Network, C extends EvmChains>
...
     yield this.manager.createUnsignedTx(tx, "WormholeTransceiver.registerPeer");
   }
+
+  async getPeer<C extends Chain>(chain: C): Promise<ChainAddress<C> | null> {
+    const peer = await this.transceiver.getWormholePeer(toChainId(chain));
+    const peerAddress = Buffer.from(peer.substring(2), 'hex');
+    const zeroAddress = Buffer.alloc(32);
+    if (peerAddress.equals(zeroAddress)) {
+      return null;
+    }
```
barnjamin, line 76: We should prefer Uint8Array over Buffer

```suggestion
    const peer = await this.transceiver.getWormholePeer(toChainId(chain));
    const peerAddress = encoding.hex.decode(peer);
    const zeroAddress = new Uint8Array(32);
    if (encoding.bytes.equals(zeroAddress, peer)) {
      return null;
    }
```

An alternate approach might be to just `BigInt(peer) === 0n` since BigInt will parse hex

```
> Buffer.alloc(32).toString("hex")
'0000000000000000000000000000000000000000000000000000000000000000'
> BigInt(Buffer.alloc(32).toString("hex"))
0n
```

sdk/evm/src/ntt.ts
```diff
@@ -117,14 +159,62 @@ export class EvmNtt<N extends Network, C extends EvmChains>
...
+    yield this.createUnsignedTx(tx, "Ntt.pause");
+  }
+
+  async *unpause() {
+    const tx = await this.manager.unpause.populateTransaction()
+    yield this.createUnsignedTx(tx, "Ntt.unpause");
+  }
+
+  async getOwner(): Promise<AccountAddress<C>> {
+    return new EvmAddress(await this.manager.owner()) as AccountAddress<C>;
```
barnjamin, line 201: does this complain without `as`?

sdk/evm/src/ntt.ts
```diff
@@ -117,14 +159,62 @@ export class EvmNtt<N extends Network, C extends EvmChains>
...
+  }
+
+  async *setOwner(owner: AccountAddress<C>) {
+    const canonicalOwner = canonicalAddress({chain: this.chain, address: owner});
```
barnjamin, line 205: The parameter here could be `AnyEvmAddress` to allow for string to be passed as well

and the `canonicalOwner` could be set as `new EvmAddress(owner).toString()`

barnjamin (commented):
partial review but looks v good

sdk/evm/src/ntt.ts
barnjamin: also works with 0x prefix
```
> BigInt("0x"+Buffer.alloc(32).toString("hex"))
0n
```

sdk/evm/src/ntt.ts
kcsongor: oo, that's neat. wasn't aware of that type

solana/ts/sdk/ntt.ts
```diff
@@ -86,6 +136,71 @@ export class SolanaNtt<N extends Network, C extends SolanaChains>
...
+    return config.mode.locking != null ? "locking" : "burning";
+  }
+
+  async isPaused(): Promise<boolean> {
+    const config = await this.getConfig();
+    return config.paused;
+  }
+
+  async *pause(payer: AccountAddress<C>) {
+    const sender = new SolanaAddress(payer).unwrap();
```
barnjamin, line 157: Same as above re parameter, changing from `AccountAddress<C>` to `AnySolanaAddress` should allow string or PublicKey to be passed, same "unwrap" method can be used to get just a PublicKey out

https://github.com/wormhole-foundation/wormhole-sdk-ts/blob/main/platforms/solana/src/types.ts#L21

sdk/evm/src/ntt.ts
kcsongor:
```
 ntt.ts   201   5 error    2322   Type 'EvmAddress' is not assignable to type 'AccountAddress<C>'.
   Type 'EvmAddress' is not assignable to type 'UniversalAddress'.
     Types of property 'address' are incompatible.
       Type 'string' is not assignable to type 'Uint8Array'. (lsp)
```

maybe the assertion is unsound?

solana/ts/sdk/ntt.ts
```diff
@@ -156,11 +283,18 @@ export class SolanaNtt<N extends Network, C extends SolanaChains>
     contracts: Contracts & { ntt: Ntt.Contracts },
     sender?: AccountAddress<SolanaChains>
   ): Promise<IdlVersion> {
-    return NTT.getVersion(
-      connection,
-      new PublicKey(contracts.ntt.manager!),
-      sender ? new SolanaAddress(sender).unwrap() : undefined
-    );
+    // TODO: what? the try catch doesn't seem to work. it's not catching the error
```
barnjamin, line 286: might have to add `await` to it?

kcsongor: AH! of course! ty

sdk/evm/src/ntt.ts
barnjamin: feels weird but I guess leave it

image

sdk/evm/src/ntt.ts
```diff
@@ -50,17 +51,48 @@ export class EvmNttWormholeTranceiver<N extends Network, C extends EvmChains>
...
+      chain: chain,
+      address: toUniversal(chain, peerAddress),
+    };
+  }
+
+  async isEvmChain(chain: Chain): Promise<boolean> {
+    return await this.transceiver.isWormholeEvmChain(toChainId(chain));
+  }
+
+  async *setIsEvmChain(chain: Chain, isEvm: boolean) {
```
solanoepalacio, line 88: nice :black_heart:

solanoepalacio (approved):

nik-suri (approved):

PR #467, opened 2024-06-28
cli tool for managing NTT deployments

kcsongor:


cli/src/evmsigner.ts
```diff
@@ -0,0 +1,205 @@
...
+            maxPriorityFeePerGas,
+            gasLimit,
+          };
+
+    // TODO: DIFF ENDS HERE
+
+    for (const txn of tx) {
+      const { transaction, description } = txn;
+      if (this.opts?.debug)
+        console.log(`Signing: ${description} for ${this.address()} (tx: ${transaction.to}})`);
```
solanoepalacio, line 150:
```suggestion
        console.log(`Signing: ${description} for ${this.address()} (tx: ${transaction.to})`);
```

cli/src/evmsigner.ts
```diff
@@ -0,0 +1,205 @@
...
+            // Hardcode type
+            type: 0,
+          }
+        : {
+            gasPrice,
+            maxFeePerGas,
+            maxPriorityFeePerGas,
+            gasLimit,
+          };
+
```
solanoepalacio, line 144: mostly feedback:
Generally as we deploy to mainy evms we find that these kind of special cases for gas price happen in many chains, so this is something that will probably be modified often... I think it makes sense to move this into a separate function so that anyone using the cli can modify it quickly without fear of breaking something else

solanoepalacio (approved):

evgeniko:
Another PR (https://github.com/wormhole-foundation/example-native-token-transfers/pull/468) updates the current README in the foundations repo to point to the CLI branch, if users would like to install the NTT CLI directly. We need to keep in mind, that after this PR is merged and the _cli_ branch is deleted, we need to update the README to point to the foundation install script.

solanoepalacio (approved):

cli/src/evmsigner.ts
kcsongor: thanks for the feedback, we can iterate on this in the future as needed 🙏

cli/test/sepolia-bsc.sh
```diff
@@ -0,0 +1,90 @@
...
+
+cleanup() {
+  kill $pid1 $pid2
+  rm -rf $dir
+}
+
+trap "cleanup" INT TERM EXIT
+
+# devnet private key
+export ETH_PRIVATE_KEY=0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80
```
evgeniko, line 34: is this a test key?

evgeniko (approved):
lgtm

cli/test/sepolia-bsc.sh
kcsongor: yes, this is one of the pre-funded wallets anvil gives you (originally comes from
hardhat)

PR #472, opened 2024-07-09
SDK: Fix the relay fee amount in Solana NTT

barnjamin:
The fee returned from `quoteDeliveryPrice` is in lamports but the `createRequestRelayInstruction` expects it to be in Sol


nvsriram (approved):
LGTM!

kcsongor (approved):

johnsaigle (approved):

PR #474, opened 2024-07-11
Typescript SDK changes and bug fixes needed for Connect

kev1n-peters:
- Added a WormholeTransferStandardRelayer payload type needed to parse the VAA for an automatic transfer between EVM chains
- EVM: getIsExecuted should check if the transfer is inbound queued to determine if the transfer is completed
- Solana: getIsExecuted should check if the inboxItem is released
- Solana: getIsApproved should not swallow exception
- Solana: getInboundQueuedTransfer should return null if the transfer is not queued
- Bumped SDK dependencies needed for Connect

### DIFF
--- a/evm/ts/src/ntt.ts
+++ b/evm/ts/src/ntt.ts
@@ -98,7 +98,7 @@ export class EvmNtt<N extends Network, C extends EvmChains>
     readonly chain: C,
     readonly provider: Provider,
     readonly contracts: Contracts & { ntt?: Ntt.Contracts },
-    readonly version: string = "default"
+    readonly version: string = "1.0.0"
   ) {
     if (!contracts.ntt) throw new Error("No Ntt Contracts provided");
 
@@ -366,9 +366,10 @@ export class EvmNtt<N extends Network, C extends EvmChains>
     transceiverMessage: Ntt.Message,
     payer?: AccountAddress<C>
   ) {
-    const tx = await this.manager.completeInboundQueuedTransfer(
-      Ntt.messageDigest(fromChain, transceiverMessage)
-    );
+    const tx =
+      await this.manager.completeInboundQueuedTransfer.populateTransaction(
+        Ntt.messageDigest(fromChain, transceiverMessage)
+      );
     yield this.createUnsignedTx(tx, "Ntt.completeInboundQueuedTransfer");
   }
 
--- a/evm/ts/src/ntt.ts
+++ b/evm/ts/src/ntt.ts
@@ -9,8 +9,6 @@ import {
   ChainAddress,
   ChainsConfig,
   Contracts,
-  TokenAddress,
-  VAA,
   serialize,
   universalAddress,
 } from "@wormhole-foundation/sdk-definitions";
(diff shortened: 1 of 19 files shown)


kev1n-peters:
Should probably release SDK `0.8.0` and change the dependencies in here to `^0.8`

evm/ts/src/ntt.ts
```diff
@@ -143,19 +141,30 @@ export class EvmNtt<N extends Network, C extends EvmChains>
...
+        : attestation.payload.payload;
+    const isExecuted = await this.manager.isMessageExecuted(
+      Ntt.messageDigest(attestation.emitterChain, payload["nttManagerPayload"])
     );
+    if (!isExecuted) return false;
+    // Also check that the transfer is not queued for it to be considered complete
+    const queued = await this.getInboundQueuedTransfer(
+      attestation.emitterChain,
+      payload["nttManagerPayload"]
+    );
```
barnjamin, line 157: Should we have a separate `isQueued` fn on the interface?

kev1n-peters: Maybe it would make the code cleaner, I can add that

sdk/route/src/types.ts
```diff
@@ -153,4 +167,15 @@ export namespace NttRoute {
...
   }
+
+  // returns true if the amount is greater than 95% of the capacity
+  // useful for warning about the possibility of a transfer being queued
+  export function isCapacityThresholdExceeded(
+    amount: bigint,
+    capacity: bigint
+  ): boolean {
+    const threshold = (capacity * 95n) / 100n;
+    console.log(amount, threshold, amount > threshold);
```
barnjamin, line 178:
```suggestion
```

solana/ts/sdk/ntt.ts
```diff
@@ -543,16 +545,26 @@ export class SolanaNtt<N extends Network, C extends SolanaChains>
...
-    if (!this.getIsApproved(attestation)) return false;
-
-    const { emitterChain } = attestation as WormholeNttTransceiver.VAA;
-    const inboundQueued = await this.getInboundQueuedTransfer(
-      emitterChain,
-      attestation
-    );
+  async getRateLimitDuration(): Promise<bigint> {
+    // The rate limit duration is hardcoded to 24 hours on Solana
+    return BigInt(24 * 60 * 60);
```
barnjamin, line 550: always?

barnjamin (dismissed):

solana/ts/sdk/ntt.ts
kev1n-peters: This is hard-coded in the contract and there's no method we can call like we do to get the version

barnjamin (dismissed):

barnjamin (approved):

sdk/route/src/manual.ts
```diff
@@ -125,24 +130,55 @@ export class NttManualRoute<N extends Network>
...
+      const dstAmount = amount.parse(
+        params.amount,
+        request.destination.decimals
+      );
+      if (
+        NttRoute.isCapacityThresholdExceeded(amount.units(dstAmount), capacity)
+      ) {
+        result.warnings = [
+          {
+            type: "DestinationCapacityWarning",
```
artursapek, line 165: extremely minor comment, this could be a shared const. but im fine with it the way it is

artursapek (approved):

sdk/route/src/manual.ts
kev1n-peters: agreed, will revisit when we add more warnings. don't want to make changes now to dismiss approvals

PR #486, opened 2024-07-25
Ntt push solana payer

evgeniko:


cli/src/index.ts
```diff
@@ -158,7 +158,12 @@ const options = {
         describe: "Skip contract verification",
         type: "boolean",
         default: false,
-    }
+    },
+    payerPath: {
+        describe: "Path to the payer json file",
```
kcsongor, line 163:
```suggestion
        describe: "Path to the payer json file (Solana)",
```

cli/src/index.ts
```diff
@@ -158,7 +158,12 @@ const options = {
...
+        default: false,
```
kcsongor, line 164: this is a little weird here, because we specify the type to be a string, then provide a boolean as a default value, which is not type correct. instead, just don't provide a default value, in that case the type of `argv["payer"]` will be `string | undefined`

cli/src/index.ts
```diff
@@ -582,14 +588,15 @@ yargs(hideBin(process.argv))
             const network = deployments.network as Network;
             const deps: Partial<{ [C in Chain]: Deployment<Chain> }> = await pullDeployments(deployments, network, verbose);
             const signerType = argv["signer-type"] as SignerType;
+            const payerPath = argv["payer"] === false ? undefined : argv["payer"];
```
kcsongor, line 591: this is redundant if we remove the `default: false` above we can then just have
```suggestion
            const payerPath = argv["payer"];
```
instead (or even inline the use of `payerPath`)

kcsongor (approved):
nice one

PR #487, opened 2024-07-26
sdk: Added resume transfer methods to routes

kev1n-peters:


kev1n-peters:
The resume method may be useful for when a transfer is inbound queued (note: we're not supporting outbound queued at the moment, should be done in a future PR) and needs to be completed. Allows UIs such as Connect to "pick up"/"resume" where a transfer left off.

sdk/route/src/manual.ts
```diff
@@ -145,6 +148,7 @@ export class NttManualRoute<N extends Network>
         token: request.destination.id,
         amount: amount.parse(params.amount, request.destination.decimals),
       },
+      eta: finality.estimateFinalityTime(request.fromChain.chain),
```
artursapek, line 151: I just realized that if we only add `eta` to quotes, then when resuming a manual route the UI will have no way of counting down time or showing a progress bar (because the receipt doesn't have any ETA in it). this probably isn't a showstopper, just calling it out.

sdk/route/src/automatic.ts
```diff
@@ -235,6 +239,87 @@ export class NttAutomaticRoute<N extends Network>
...
+    return {
+      from: vaa.emitterChain,
+      to: recipientChain,
+      state: TransferState.Attested,
+      originTxs: [tx],
+      attestation: {
+        id: msgId,
+        attestation: vaa,
+      },
+      params: {
```
artursapek, line 298: out of scope for this change but wdyt about requiring all `Receipt` types to include the `Vp` in the future? NTT is the only route that does this today, I believe.

artursapek (approved):
read through it a couple times, seems good to me.

sdk/route/src/manual.ts
kev1n-peters: Right now the `eta` is just an estimate for how long it takes for the attestation to become available (essentially the source chain's finality time)

sdk/route/src/automatic.ts
kev1n-peters: I don't see why we couldn't, I suppose the more information on the receipt the better

evgeniko (approved):
lgtm

PR #490, opened 2024-07-31
cli: query token decimals outside of forge script

kcsongor:
This works around and edge case when the token contract is compiled against a different EVM version than what the forge script is running in (london currently).


evgeniko (approved):
lgtm

PR #491, opened 2024-07-31
solana: Update README prerequisites and testing instructions

nvsriram:
This PR adds:
* Updates to `Prerequisites` to match `Anchor.toml` and instructions to verify toolchain
* Update to `Testing` make command to match `Makefile` target
* `Troubleshooting` section


evgeniko (approved):

solana/README.md
```diff
@@ -103,5 +106,42 @@ In order to run them, the Solana programs must be built and their IDL made avail
...
+
+  ```sh
+  tsx scripts/regenerateIdl.ts $$jsonfile > $$tsfile; \
+  ```
+
+  to:
+
+  ```sh
+  npx tsx scripts/regenerateIdl.ts $$jsonfile > $$tsfile; \
+  ```
```
nik-suri, line 135: Should we just update the Makefile to use `npx tsx` by default?

nik-suri (approved):
Looks good to me! Left one small comment

solana/README.md
nvsriram: We could update it but it might need to be tested out in different machines. Just `tsx` didn't work for me and so had to update it to `npx tsx` to make it work. There could be cases where `npx` is not globally installed either. 
So I think we can just have it in the troubleshooting section for now and if others run into this, we could accordingly update the Makefile.

solana/README.md
```diff
@@ -103,5 +106,42 @@ In order to run them, the Solana programs must be built and their IDL made avail
...
\ No newline at end of file
+make test
+```
+
+### Troubleshooting
+
+<details>
+<summary><code>make: *** No rule to make target `test'.  Stop.</code></summary>
+
+- Ensure `Makefile` has target `test`
```
kcsongor, line 117: not sure if this should be called out here, the makefile does have a test target. How would someone hit this issue?

solana/README.md
```diff
@@ -103,5 +106,42 @@ In order to run them, the Solana programs must be built and their IDL made avail
...
+
+<details>
+<summary><code>tsx: command not found</code></summary>
+
+- Screenshot:
+  <img src="images/tsx-command-not-found.png" alt="tsx command not found screenshot">
+- Update `Makefile` ([line #29](https://github.com/wormhole-foundation/example-native-token-transfers/blob/main/solana/Makefile#L29)) from:
+
+  ```sh
+  tsx scripts/regenerateIdl.ts $$jsonfile > $$tsfile; \
```
kcsongor, line 128: wouldn't it make sense to just put into the prerequisites section how to install this binary?
encouraging local modifications sounds a bit dubious

solana/README.md
```diff
@@ -103,5 +106,42 @@ In order to run them, the Solana programs must be built and their IDL made avail
...
+
+  </details>
+
+  <details>
+  <summary><code>Lifecycle script `build:esm` failed with error</code></summary>
+
+  - Screenshot:
+    <img src="images/lifecycle-script.png" alt="lifecycle script screenshot">
+  - This occurs due to Typescript files failing compilation.
+  - [`patch-idl` script](https://github.com/wormhole-foundation/example-native-token-transfers/blob/main/solana/scripts/patch-idl) requires [`jq`](https://jqlang.github.io/jq/) to be installed. Install `jq` and retry.
```
kcsongor, line 145: again, maybe the prerequisites section would be a better place for this.

also we should make the error explicitly warn about this

solana/README.md
nvsriram: I think I hit the issue because the README was not up-to-date with the changes that were being made. Currently, this issue would not be hit.

solana/README.md
nvsriram: Yeah, my idea for the troubleshooting section was that it was specific to testing but I do see the merit in moving it to the prerequisites section as well.

PR #492, opened 2024-07-31
Make the ntt code public

real-felix:
Make some NTT code public so that I can use it in the token-bridge-transfer. I need the quotes to make the user pay before a transfer.


real-felix:
We don't need that anymore because we'll have our dedicated quoter for TBRv3.

PR #502, opened 2024-08-21
Makefile: Add Solana Makefile targets to top-level dir

nvsriram:
This PR is based on #299.

* Makefile targets and EVM CI targets are renamed to sync with latest target names
* Lint command is updated to sync with latest [.github/workflows/solana.yml](https://github.com/wormhole-foundation/example-native-token-transfers/blob/main/.github/workflows/solana.yml) file command
* Refactored top-level Makefile EVM targets to EVM Makefile
* General targets for `build`, `test`, and `lint` added that run EVM and Solana versions of the same.


Makefile
```diff
@@ -30,11 +36,32 @@ check-format:
...
+
+.PHONY: test-anchor
+test-anchor:
+	cd solana; make test
+
+.PHONY: test-solana
+test-solana: build-solana test-solana-unit build-anchor test-anchor
+
+
+.PHONY: lint
```
nonergodic, line 63:
```suggestion
.PHONY: lint-solana
```

Makefile
```diff
@@ -30,11 +36,32 @@ check-format:
...
 	cd evm && forge fmt
 
-.PHONY: test
+.PHONY: test-evm
 test-evm:
 	cd evm && forge test -vvv
 
+
 # Verify that the contracts do not include PUSH0 opcodes
 test-push0:
```
nonergodic, line 45:
```suggestion
.PHONY: test-push0
test-push0:
```

nonergodic (changes requested):
Generally speaking, this does not look like a particularly sensible Makefile to me. Nothing here depends on anything, it's really just a set of commands. There isn't even general targets to build/test/lint everything.

nonergodic (dismissed):
Eventually this should probably be split into 2 Makefiles - one for each subdir (evm, solana), but good enough for now.

kcsongor:
I want to echo @nonergodic's point here: I think we should be moving these targets into the subdirectory makefiles, instead of out of them. It looks like the makefile was moved to the top level in this PR https://github.com/wormhole-foundation/example-native-token-transfers/pull/147/files but I don't really see the reason why.

I think we should just move them back into the relevant subdirectories instead, and if someone wants to run these targets from the top-level, just do `make -C evm build` etc. It's a lot cleaner than ad-hoc namespacing the build targets.

.github/workflows/evm.yml
```diff
@@ -29,43 +29,43 @@ jobs:
 
       - name: Run Forge build
         run: |
-          make test-push0
+          make -C evm push0-test
```
kcsongor, line 32: minor nit: it's possible to specify the working directory for all the commands here, like so:

https://github.com/wormholelabs-xyz/example-native-token-transfers/blob/8712337bb9ec20e6cf32ab70767789af1d0025f6/.github/workflows/solana.yml#L10-L12

kcsongor (approved):

nonergodic (approved):

PR #505, opened 2024-08-29
evm: forge fmt

kcsongor:
This partially undoes #501, which itself updated formatting after an upstream foundry change. That upstream change has been reverted https://github.com/foundry-rs/foundry/pull/8762, so we do the same here.


nvsriram (approved):

nik-suri (approved):
lgtm, changes are purely cosmetic

PR #514, opened 2024-09-10
solana: Add no registered transceivers check

nvsriram:
This PR is based on #298.

### DIFF
--- a/solana/programs/example-native-token-transfers/src/instructions/redeem.rs
+++ b/solana/programs/example-native-token-transfers/src/instructions/redeem.rs
@@ -24,6 +24,8 @@ pub struct Redeem<'info> {
     // NOTE: this works when the contract is paused
     #[account(
         constraint = config.threshold > 0 @ NTTError::ZeroThreshold,
+        constraint = config.next_transceiver_id != 0 @ NTTError::NoRegisteredTransceivers,
+        constraint = !config.enabled_transceivers.is_empty() @ NTTError::NoRegisteredTransceivers,
     )]
     pub config: Account<'info, Config>,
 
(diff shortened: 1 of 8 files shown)


nonergodic (approved):

solana/programs/example-native-token-transfers/src/instructions/redeem.rs
```diff
@@ -24,6 +24,8 @@ pub struct Redeem<'info> {
     // NOTE: this works when the contract is paused
     #[account(
         constraint = config.threshold > 0 @ NTTError::ZeroThreshold,
+        constraint = config.next_transceiver_id != 0 @ NTTError::NoRegisteredTransceivers,
```
kcsongor, line 27: these seem redudnant because we already have a constraint on `transceiver` (line 51) that it must be enabled. So if that constraint passes, it means that it's enabled, so we can't have 0 enabled transceivers

solana/programs/example-native-token-transfers/src/instructions/transfer.rs
```diff
@@ -37,6 +37,11 @@ pub struct Transfer<'info> {
     #[account(mut)]
     pub payer: Signer<'info>,
 
+    // Ensure that there exists at least one enabled transceiver
+    #[account(
+        constraint = config.next_transceiver_id != 0 @ NTTError::NoRegisteredTransceivers,
```
kcsongor, line 42: just the second constraint seems sufficient to me

solana/programs/example-native-token-transfers/src/instructions/redeem.rs
nonergodic: I had the same impression, but assumed it was intentional... "to be safe". I disagree with this "to be safe" philosophy, but I assumed that this is what's going on here because of other instances in the past.

kcsongor (approved):

nonergodic (approved):

PR #516, opened 2024-09-12
solana: add instruction to transfer ownership in one step

kcsongor:
When reviewing, it's best to look at the differences between the new instruction and `transfer_ownership`. The changes:
1. immediately set the new owner
2. use `set_upgrade_authority` instead of `set_upgrade_authority_checked`

Reused the same account struct between the two. Technically the `upgrade_lock` account is redundant on this 1-step transfer code path, but it's a small price to pay for not having to duplicate the entire account struct.

### DIFF
--- a/solana/programs/example-native-token-transfers/src/instructions/admin.rs
+++ b/solana/programs/example-native-token-transfers/src/instructions/admin.rs
@@ -24,7 +24,7 @@ use crate::{
 /// instruction.
 ///
 /// Alternatively, the ownership can be transferred in a single step by calling the
-/// [`transfer_ownership_one_step`] instruction. This can be dangerous because if the new owner
+/// [`transfer_ownership_one_step_unchecked`] instruction. This can be dangerous because if the new owner
 /// cannot actually sign transactions (due to setting the wrong address), the program will be
 /// permanently locked. If the intention is to transfer ownership to a program using this instruction,
 /// take extra care to ensure that the owner is a PDA, not the program address itself.
@@ -79,7 +79,7 @@ pub fn transfer_ownership(ctx: Context<TransferOwnership>) -> Result<()> {
     )
 }
 
-pub fn transfer_ownership_one_step(ctx: Context<TransferOwnership>) -> Result<()> {
+pub fn transfer_ownership_one_step_unchecked(ctx: Context<TransferOwnership>) -> Result<()> {
     ctx.accounts.config.pending_owner = None;
     ctx.accounts.config.owner = ctx.accounts.new_owner.key();
 
--- a/solana/programs/example-native-token-transfers/src/instructions/admin.rs
+++ b/solana/programs/example-native-token-transfers/src/instructions/admin.rs
@@ -15,13 +15,19 @@ use crate::{
 
 // * Transfer ownership
 
-/// Transferring the ownership is a 2-step process. The first step is to set the
+/// For safety reasons, transferring ownership is a 2-step process. The first step is to set the
 /// new owner, and the second step is for the new owner to claim the ownership.
 /// This is to prevent a situation where the ownership is transferred to an
 /// address that is not able to claim the ownership (by mistake).
 ///
 /// The transfer can be cancelled by the existing owner invoking the [`claim_ownership`]
 /// instruction.
+///
+/// Alternatively, the ownership can be transferred in a single step by calling the
+/// [`transfer_ownership_one_step`] instruction. This can be dangerous because if the new owner
+/// cannot actually sign transactions (due to setting the wrong address), the program will be
+/// permanently locked. If the intention is to transfer ownership to a program using this instruction,
+/// take extra care to ensure that the owner is a PDA, not the program address itself.
 #[derive(Accounts)]
 pub struct TransferOwnership<'info> {
     #[account(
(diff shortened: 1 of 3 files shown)


solana/programs/example-native-token-transfers/src/instructions/admin.rs
```diff
@@ -73,6 +79,28 @@ pub fn transfer_ownership(ctx: Context<TransferOwnership>) -> Result<()> {
     )
 }
 
+pub fn transfer_ownership_one_step(ctx: Context<TransferOwnership>) -> Result<()> {
```
johnsaigle, line 82: Any thoughts on renaming this to something like `transfer_ownership_unchecked` or `transfer_ownership_one_step_unchecked` (or `dangerous` instead of unchecked etc.)?

It might be nice to surface the `_unchecked` naming from the BPF loader call into our API so that users know exactly what they're up to and are encouraged to read the documentation thoroughly.

solana/programs/example-native-token-transfers/src/instructions/admin.rs
```diff
@@ -15,13 +15,19 @@ use crate::{
...
 /// new owner, and the second step is for the new owner to claim the ownership.
 /// This is to prevent a situation where the ownership is transferred to an
 /// address that is not able to claim the ownership (by mistake).
 ///
 /// The transfer can be cancelled by the existing owner invoking the [`claim_ownership`]
 /// instruction.
+///
+/// Alternatively, the ownership can be transferred in a single step by calling the
+/// [`transfer_ownership_one_step`] instruction. This can be dangerous because if the new owner
+/// cannot actually sign transactions (due to setting the wrong address), the program will be
```
johnsaigle, line 28: Very very minor point on this but what if we add a constraint such that `new_owner` cannot be `upgrade_lock`?

* In the normal two-step case, this is invalid but it can be (re)claimed by the current `owner`.

* In the one-step case, this bricks the owner field though not in a particularly special way as any invalid address will permanently lock the owner field, as you noted here.

That said, might as well eliminate one tiny footgun since it's wrong in both cases?

solana/programs/example-native-token-transfers/tests/governance.rs
```diff
@@ -131,6 +131,50 @@ async fn test_governance() {
...
+        set_paused(&test_data.ntt, SetPaused { owner: OWNER }, true),
+        None,
+        None,
+        None,
+    )
+    .await
+    .unwrap();
+
+    let config_account: Config = ctx.get_account_data_anchor(test_data.ntt.config()).await;
+    assert!(config_account.paused);
```
johnsaigle, line 175: Would you mind adding an `assert!()` that the contract is _not_ paused at the beginning of step 1? I assume it isn't by default but it makes it a bit more clear and guaranteed that the governance call does change the value.

solana/programs/example-native-token-transfers/tests/governance.rs
```diff
@@ -131,6 +131,50 @@ async fn test_governance() {
...
+    .submit_with_signers(&[&test_data.program_owner], &mut ctx)
+    .await
+    .unwrap();
+
+    // step 2. set paused
+    wrap_governance(
+        &mut ctx,
+        &test_data.governance,
+        &test_data.ntt.wormhole,
+        set_paused(&test_data.ntt, SetPaused { owner: OWNER }, true),
```
johnsaigle, line 166: Should `OWNER` be `governance_pda` here? The test works on my machine so I guess not, but I'm a bit confused reading the code. (I'll dig around test suite more to figure out what's going on but this line looks a little odd at first glance given that we've just changed the owner)

solana/programs/example-native-token-transfers/src/instructions/admin.rs
```diff
@@ -15,13 +15,19 @@ use crate::{
...
+/// permanently locked. If the intention is to transfer ownership to a program using this instruction,
+/// take extra care to ensure that the owner is a PDA, not the program address itself.
```
johnsaigle, line 30: Is there anything we can do to help the user here? Could we check that the account has data for example and use that as a heuristic about whether that series of bytes should _really_ be the permanent owner?

Perhaps one approach would be adding extra endpoints here, like `transfer_owner_to_account` vs. `transfer_owner_to_program_derived_account`? I'm asking in case there are any checks we can do to help the user not brick the owner here. If they use one route we could help them determine whether the address is really a PDA or else a regular user account and so on.

Arguably any of these would be a restriction especially if the point of this entrypoint is to do this cross-chain basically not worry about errors.

johnsaigle (commented):
The change is good, should cover the case for doing a (risky) one-step ownership update. The documentation and unit test are a nice addition to help give context for the new entrypoint.

I left some _very_ minor feedback that doesn't need to be blocking. They are suggestions that might add some belt-and-suspenders-style safety so that users make a really deliberate choice when taking this path of no return.

solana/programs/example-native-token-transfers/src/instructions/admin.rs
kcsongor: funnily enough I originally named it exactly `transfer_ownership_unchecked` but wasn't satisfied. I like `transfer_ownership_one_step_unchecked`. It's a mouthful, but that's a feature in this case

solana/programs/example-native-token-transfers/src/instructions/admin.rs
kcsongor: I agree it makes no sense to directly set the owner to `upgrade_lock`, but wouldn't it make the code a little confusing to read? In that I think that would be a fairly arbitrary check, or at least I don't see the risk of anyone accidentally supplying that particular wrong address vs any other one

solana/programs/example-native-token-transfers/tests/governance.rs
kcsongor: good shout

solana/programs/example-native-token-transfers/src/instructions/admin.rs
kcsongor: yeah, unfortunately we can't really tell if it's a sensible account or not. The only way to tell is if the new account demonstrates that it can sign a transaction, which is why the existing flow is 2-step. This is the "I know what I'm doing" use case, so everyone should use it at their own peril

solana/programs/example-native-token-transfers/src/instructions/admin.rs
johnsaigle: I think if I was reading for the first time it's helpful to know if something is invalid especially if it has something like `Error::UpgradeLockCantBeOwner`

Clearly a subjective matter though :) So let's not do it

solana/programs/example-native-token-transfers/tests/governance.rs
johnsaigle: Ah OK so OWNER isn't constant it's like a placeholder for whatever the actual owner value is at a given time? Interesting

solana/programs/example-native-token-transfers/src/instructions/admin.rs
johnsaigle: Fair point, I suppose it does sort of put pressure on someone to take the normal, safe path too if we provide absolutely no help on the dangerous route

solana/programs/example-native-token-transfers/tests/governance.rs
kcsongor: yeah, it's replaced here
https://github.com/wormhole-foundation/example-native-token-transfers/blob/4f165a3d8c743e543fbc38e958185aad06f22d91/solana/programs/wormhole-governance/src/instructions/governance.rs#L464-L470

johnsaigle (approved):
Previous conversations resolved and addressed.

mdulin2 (approved):

PR #517, opened 2024-09-12
evm: per chain transceivers

bruce-riley:
This PR implements the `NttManagerWithPerChainTransceivers` contract which inherits from `NttManagerNoRateLimiting` and allows configuring different transceivers and thresholds for each chain. You can configure a different set of send and receive transceivers for each chain. If you don't specifically enable any transceivers for a chain, then all transceivers will be used.

This new contract has to inherit from the no-rate-limiting due to the contract size limit.


bruce-riley:
Contract sizes before this feature:

```
| NttManager                            |   23,615 |        961 |
| NttManagerNoRateLimiting              |   18,169 |      6,407 |
```

Contract sizes as of commit e442b2d62b244db7f342d8fcac4f64a3621142d0:

```
| NttManager                                  |   23,853 |        723 |
| NttManagerNoRateLimiting                    |   19,381 |      5,195 |
```

evm/src/NttManager/ManagerBase.sol
```diff
@@ -482,7 +527,12 @@ abstract contract ManagerBase is
     }
 
     function _checkThresholdInvariants() internal view {
-        uint8 threshold = _getThresholdStorage().num;
+        _checkThresholdInvariants(_getThresholdStorage().num);
+    }
+
+    function _checkThresholdInvariants(
+        uint8 threshold
+    ) internal pure {
```
djb15, line 535: I'm surprised the compiler doesn't complain at you for making this pure given we read state?

evm/src/NttManager/NttManagerWithPerChainTransceivers.sol
```diff
@@ -0,0 +1,165 @@
...
+    /// @inheritdoc IManagerBase
+    function enableRecvTransceiverForChain(
+        address transceiver,
+        uint16 forChainId
+    ) external override(ManagerBase, IManagerBase) onlyOwner {
+        _enableTransceiverForChain(transceiver, forChainId, RECV_TRANSCEIVER_BITMAP_SLOT);
+        emit RecvTransceiverEnabledForChain(transceiver, forChainId);
+    }
+
+    function _enableTransceiverForChain(
```
djb15, line 50: I think we should also add the inverse `_disableTransceiverForChain` with associated entrypoints?

evm/src/NttManager/NttManagerWithPerChainTransceivers.sol
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    function _getPerChainTransceiverBitmap(
+        uint16 forChainId,
+        bytes32 tag
+    ) private view returns (uint64 bitmap) {
+        bitmap = _getPerChainTransceiverBitmapStorage(tag)[forChainId].bitmap;
+        if (bitmap == 0) {
+            bitmap = type(uint64).max;
+        }
```
djb15, line 92: Doesn't this default to all registered transceivers being enabled, even if they were disabled? Should this instead load from `_getTransceiverBitmapStorage`?

evm/src/NttManager/NttManagerWithPerChainTransceivers.sol
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    function _getMessageAttestations(
+        bytes32 digest
+    ) internal view override returns (uint64) {
+        AttestationInfo memory attInfo = _getMessageAttestationsStorage()[digest];
+        uint64 enabledTransceiverBitmap = _getEnabledTransceiversBitmap();
+        uint64 enabledTransceiversForChain =
+            _getEnabledRecvTransceiversForChain(attInfo.sourceChainId);
+        return attInfo.attestedTransceivers & enabledTransceiverBitmap & enabledTransceiversForChain;
```
djb15, line 163: Nice, at a glance I think this may guard against my previous comment. Is defaulting the bitmap to `type(uint64).max` above instead of loading the enabled transceivers deliberate to save gas?

djb15 (commented):
Would like to have a longer look at this next week, the bitmap and threshold stuff was a source of some pain in the original audit.

One general comment is I feel like it becomes a lot easier to misconfig with these changes. For example, in a 2 transceiver, threshold of 2 global setup, if only 1 transceiver is enabled for receiving from a given chain, but the deployer forgets to set the `perChainThreshold` for that chain, then messages won't get delivered. Maybe in `enableRecvTransceiverForChain` we need to check that `perChainThreshold` is <= set bits in the joint `enabledTransceiverBitmap & enabledTransceiversForChain` bitmap?

evm/src/NttManager/ManagerBase.sol
bruce-riley: That is curious. I changed it to view.

evm/src/NttManager/NttManagerWithPerChainTransceivers.sol
bruce-riley: Is this really necessary (worth the storage)? Is an integrator ever likely to disable something?

bruce-riley:
> Would like to have a longer look at this next week, the bitmap and threshold stuff was a source of some pain in the original audit.
> 
> One general comment is I feel like it becomes a lot easier to misconfig with these changes. For example, in a 2 transceiver, threshold of 2 global setup, if only 1 transceiver is enabled for receiving from a given chain, but the deployer forgets to set the `perChainThreshold` for that chain, then messages won't get delivered. Maybe in `enableRecvTransceiverForChain` we need to check that `perChainThreshold` is <= set bits in the joint `enabledTransceiverBitmap & enabledTransceiversForChain` bitmap?

That would require them to set the per-chain threshold before enabling anything, right? Is that acceptable?

evm/src/NttManager/NttManagerWithPerChainTransceivers.sol
bruce-riley: Yeah, returning max does avoid a storage read, and makes this work. I can either add a comment on `_getPerChainTransceiverBitmap` to explain that or add the extra storage read in `_getPerChainTransceiverBitmap`. How strongly do you feel about it?

djb15: I'm ok with adding some inline docs to explain it, think it's a neat way to save some gas

djb15:
> > Would like to have a longer look at this next week, the bitmap and threshold stuff was a source of some pain in the original audit.
> > One general comment is I feel like it becomes a lot easier to misconfig with these changes. For example, in a 2 transceiver, threshold of 2 global setup, if only 1 transceiver is enabled for receiving from a given chain, but the deployer forgets to set the `perChainThreshold` for that chain, then messages won't get delivered. Maybe in `enableRecvTransceiverForChain` we need to check that `perChainThreshold` is <= set bits in the joint `enabledTransceiverBitmap & enabledTransceiversForChain` bitmap?
> 
> That would require them to set the per-chain threshold before enabling anything, right? Is that acceptable?

I was thinking we could write `perChainThreshold` to 1 when we add the first per chain transceiver? I think we do something similar in the existing transceiver registry

evm/src/NttManager/NttManagerWithPerChainTransceivers.sol
bruce-riley: As discussed below, this saves gas by avoid an extra storage slot read, and it works because the value gets combined with the bitmap of enabled transceivers. I added a comment to reflect that.

evm/test/PerChainTransceivers.t.sol
```diff
@@ -0,0 +1,789 @@
...
+        require(nttManagerChain1.getThreshold() != 0, "Threshold is zero with active transceivers");
+
+        // Actually set it
+        nttManagerChain1.setThreshold(1);
+        nttManagerChain2.setThreshold(1);
+        nttManagerChain3.setThreshold(1);
+
+        // On chain 1, set the threshold to chain 3 to be different.
+        nttManagerChain1.setPerChainThreshold(chainId3, 2);
+        nttManagerChain3.setPerChainThreshold(chainId1, 2);
```
banescusebi, line 283: Does this mean that the per-chain threshold should be set to the same value in both directions? Or is it just a coincidence that the threshold is set to 2 in both c1 -> c3 and c3 -> c1?

banescusebi (commented):
I will continue looking a bit into the test files, but wanted to submit the comments so far.

evm/src/NttManager/NttManagerWithPerChainTransceivers.sol
```diff
@@ -0,0 +1,183 @@
+// SPDX-License-Identifier: Apache 2
+pragma solidity >=0.8.8 <0.9.0;
+
+import "./NttManagerNoRateLimiting.sol";
+
+/// @title  NttManagerWithPerChainTransceivers
+/// @author Wormhole Project Contributors.
+/// @notice The NttManagerWithPerChainTransceivers abstract contract is an implementation of
```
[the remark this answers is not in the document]
bruce-riley: The comment was out of date.

evm/src/NttManager/NttManagerWithPerChainTransceivers.sol
```diff
@@ -0,0 +1,183 @@
...
+    /// @inheritdoc IManagerBase
+    function enableRecvTransceiverForChain(
+        address transceiver,
+        uint16 forChainId
+    ) external override(ManagerBase, IManagerBase) onlyOwner {
+        _enableTransceiverForChain(transceiver, forChainId, RECV_TRANSCEIVER_BITMAP_SLOT);
+
+        // Make sure the threshold is set to something.
+        if (getPerChainThreshold(forChainId) == 0) {
+            this.setPerChainThreshold(forChainId, 1);
```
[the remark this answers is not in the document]
bruce-riley: No, the default is the global default threshold, which could be set higher than number of transceivers on the chain. I think that's the concern @djb15 has.

After discussing this with @evan-gray, I am now thinking we should *not* use the global default threshold if they configure transceivers for a chain. It seems like that avoids some confusion and saves a storage read.

evm/src/NttManager/NttManagerWithPerChainTransceivers.sol
bruce-riley: I will add the disable methods.

evm/src/NttManager/ManagerBase.sol
bruce-riley: Interesting. I changed this to view and now it is giving me this:

```
Warning (2018): Function state mutability can be restricted to pure
   --> src/NttManager/ManagerBase.sol:533:5:
    |
533 |     function _checkThresholdInvariants(
    |     ^ (Relevant source part starts here and spans across multiple lines).
```

evm/src/NttManager/NttManagerWithPerChainTransceivers.sol
```diff
@@ -0,0 +1,183 @@
...
+
+        TransceiverInfo storage transceiverInfo = _getTransceiverInfosStorage()[transceiver];
+
+        if (!transceiverInfo.registered) {
+            revert NonRegisteredTransceiver(transceiver);
+        }
+
+        mapping(uint16 => _EnabledTransceiverBitmap) storage _bitmaps =
+            _getPerChainTransceiverBitmapStorage(tag);
+        _bitmaps[forChainId].bitmap |= uint64(1 << transceiverInfo.index);
```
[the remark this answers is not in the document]
bruce-riley: Yes, it is a requirement that `setTransceiver` has been called prior to this, otherwise the registered check would fail.

evm/src/NttManager/NttManagerWithPerChainTransceivers.sol
```diff
@@ -0,0 +1,183 @@
...
+    function _enableTransceiverForChain(
+        address transceiver,
+        uint16 forChainId,
+        bytes32 tag
+    ) private {
+        if (transceiver == address(0)) {
+            revert InvalidTransceiverZeroAddress();
+        }
+
+        TransceiverInfo storage transceiverInfo = _getTransceiverInfosStorage()[transceiver];
```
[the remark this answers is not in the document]
bruce-riley: We have to be very careful about adding functionality to the transceiver registry because the `NttManager` contract is running low on code space. That's why `NttManagerWithPerChainTransceiver` inherits from the no-rate-limiting one rather than NttManager.

evm/src/NttManager/NttManagerWithPerChainTransceivers.sol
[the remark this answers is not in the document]
banescusebi: that sounds reasonable. I feel like this aspect should also be captured by the README.md section you added

evm/src/NttManager/NttManagerWithPerChainTransceivers.sol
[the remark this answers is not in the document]
banescusebi: Oh yeah, you're totally right about that. I'm not a big fan of adding more to the contracts. It feels a bit confusing that the logic now contains both per-chain transceivers, however, there are some functions input params and return values don't have a `chainId`.

evm/src/NttManager/NttManagerWithPerChainTransceivers.sol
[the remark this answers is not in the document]
banescusebi: Makes sense. Should this prerequisite be indicated in a comment at the beginning of this function?

bruce-riley:
We've decided to not support falling back to the global defaults, and also switch to functions to get / set the bitmaps per chain, rather than doing enable/disable.

I'm converting this back to draft while I do the rework.

evm/src/NttManager/NttManagerWithPerChainTransceivers.sol
bruce-riley: This comment was made obsolete by the switch to use get/set instead of enable/disable.

evm/src/NttManager/NttManagerWithPerChainTransceivers.sol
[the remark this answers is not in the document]
bruce-riley: Please take another look since this has all changed.

evm/src/interfaces/IManagerBase.sol
```diff
@@ -126,6 +156,13 @@ interface IManagerBase {
         uint8 threshold
     ) external;
 
+    /// @notice Sets the per-chain threshold for the number of attestations required for a message
+    /// to be considered valid. Note that if a threshold is not specified for a chain, the default applies.
```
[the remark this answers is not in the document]
bruce-riley: Please take another look since this has all changed.

evm/src/NttManager/NttManagerWithPerChainTransceivers.sol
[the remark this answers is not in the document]
bruce-riley: Please take another look since this has all changed.

evm/src/NttManager/NttManagerWithPerChainTransceivers.sol
```diff
@@ -0,0 +1,183 @@
...
+
+        if (!transceiverInfo.registered) {
```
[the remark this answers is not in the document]
bruce-riley: Please take another look since this has all changed.

evm/src/NttManager/NttManagerWithPerChainTransceivers.sol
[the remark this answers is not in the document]
bruce-riley: Please take another look since this has all changed.

evm/test/PerChainTransceivers.t.sol
```diff
@@ -0,0 +1,789 @@
...
+    address userD = address(0xABC);
+
+    address relayer = address(0x28D8F1Be96f97C1387e94A53e00eCcFb4E75175a);
+    IWormhole wormhole = IWormhole(0x4a8bc80Ed5a4067f1CCf107057b8270E0cC11A78);
+
+    // This function sets up the following config:
+    // - A manager on each of three chains.
+    // - Two transceivers on each chain, all interconnected as peers.
+    // - On chain one, it sets a default threshold of one and a per-chain threshold of two for chain three.
+    // - On chain three, it sets a default threshold of one and a per-chain threshold of two for chain one.
```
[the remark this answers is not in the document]
bruce-riley: Please take another look since this has all changed.

evm/test/PerChainTransceivers.t.sol
bruce-riley: Please take another look since this has all changed.

evm/README.md
```diff
@@ -275,3 +275,11 @@ The `mint` and `setMinter` methods found in the [`INttToken` Interface](src/inte
 Although the rate limiting feature can be disabled during contract instantiation, it still occupies code space in the `NttManager` contract. If you wish to free up code space for custom development, you can instead instantiate the
 `NttManagerNoRateLimiting` contract. This contract is built without the bulk of the rate limiting code. Note that the current immutable checks do not allow modifying the rate-limiting parameters during migration. This means migrating
 from an instance of `NttManager` with rate-limiting enabled to `NttManagerNoRateLimiting` or vice versa is not officially supported.
+
+#### Per-Chain-Transceivers
+
+The `NttManagerWithPerChainTransceivers` contract which inherits from `NttManagerNoRateLimiting` and allows configuring different transceivers and thresholds for each chain.
```
evan-gray, line 281: nit: rm "which"

evm/README.md
```diff
@@ -275,3 +275,11 @@ The `mint` and `setMinter` methods found in the [`INttToken` Interface](src/inte
...
+You can configure a different set of send and receive transceivers for each chain. If you don't specifically enable any transceivers for a chain, then all transceivers will be used.
```
evan-gray, line 282: Not sure this is true after the most recent change (there's no defaulting anymore).

evm/src/NttManager/NttManagerWithPerChainTransceivers.sol
```diff
@@ -0,0 +1,356 @@
...
+        uint64 indexBitmap,
+        uint8 threshold
+    ) public onlyOwner {
+        _validateTransceivers(indexBitmap);
+
+        // Validate the threshold against the bitmap.
+        uint8 numEnabled = countSetBits(indexBitmap);
+        if (threshold > numEnabled) {
+            revert ThresholdTooHigh(threshold, numEnabled);
+        }
```
djb15, line 165: Is it worth checking the threshold against the set bits count in the union of the recv bitmap and the global enabled transceiver bitmap? Reduces the probability of setting the wrong bitmap

djb15: Similarly, if the admin disables a transceiver globally then the per-chain threshold could be > the union of the enabled transceivers. Worth having an equivalent of `_checkTransceiverInvariants` (might need to override again sadly)?

djb15 (commented):
- Should `setThreshold` be overridden and revert?
- More generally to the above bullet point, we should try to minimise unintended side effects of enabling/disabling transceivers globally/per-chain and also setting global/per-chain thresholds (global thresholds appear to be unused now, but worth being explicit)

evm/src/NttManager/NttManagerWithPerChainTransceivers.sol
```diff
@@ -0,0 +1,356 @@
...
+        uint16[] storage chains = _getEnabledChainsStorage(tag);
+        uint256 zeroIdx = type(uint256).max;
+        uint256 len = chains.length;
+        for (uint256 idx = 0; (idx < len); idx++) {
+            if (chains[idx] == forChainId) {
+                return;
+            }
+            if (chains[idx] == 0) {
+                zeroIdx = idx;
+            }
```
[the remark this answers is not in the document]
bruce-riley: Left over from a previous version of `_removeChain` that used `delete()` which left holes in the array. Removed this.

evm/src/NttManager/NttManagerWithPerChainTransceivers.sol
```diff
@@ -27,61 +73,184 @@ contract NttManagerWithPerChainTransceivers is NttManagerNoRateLimiting {
...
+    }
+
+    function _getMessageAttestations(
+        bytes32 digest
+    ) internal view override returns (uint64) {
+        AttestationInfo memory attInfo = _getMessageAttestationsStorage()[digest];
+        uint64 enabledTransceiverBitmap = _getEnabledTransceiversBitmap();
+        uint64 enabledTransceiversForChain =
+            _getEnabledRecvTransceiversForChain(attInfo.sourceChainId);
+        return attInfo.attestedTransceivers & enabledTransceiverBitmap & enabledTransceiversForChain;
```
banescusebi, line 231: The `_checkTransceiversInvariants()` ensures that you can't remove a transceiver globally, when it's enabled for a chain. Therefore `enabledTransceiversForChain` is a subset of `enabledTransceiverBitmap`.

evm/src/NttManager/NttManagerWithPerChainTransceivers.sol
```diff
@@ -27,61 +73,184 @@ contract NttManagerWithPerChainTransceivers is NttManagerNoRateLimiting {
...
+    }
+
+    function _checkTransceiversInvariants() internal view override {
+        super._checkTransceiversInvariants();
+        _validateTransceivers(SEND_ENABLED_CHAINS_SLOT, SEND_TRANSCEIVER_BITMAP_SLOT);
+        _validateTransceivers(RECV_ENABLED_CHAINS_SLOT, RECV_TRANSCEIVER_BITMAP_SLOT);
```
banescusebi, line 237: This `_checkTransceiversInvariants()` seems quite heavy. Would it be worth doing an analysis at how many chains and how many transceivers per chain do the calling functions hit the block gas limit?

evm/test/PerChainTransceivers.t.sol
```diff
@@ -269,100 +269,286 @@ contract TestPerChainTransceivers is Test, IRateLimiterEvents {
...
-            abi.encodeWithSelector(TransceiverRegistry.InvalidTransceiverZeroAddress.selector)
+            abi.encodeWithSelector(
+                NttManagerWithPerChainTransceivers.InvalidTransceiverIndex.selector, 6
+            )
         );
-        nttManagerChain1.enableSendTransceiverForChain(address(0), chainId2);
+        nttManagerChain1.setRecvTransceiverBitmapForChain(chainId2, 0x40, 0);
 
-        // Can't enable transceiver for an unknown address.
+        // Can't globally remove a transceiver when it's enabled for a chain.
```
banescusebi, line 511: Love this test.

evm/README.md
```diff
@@ -316,8 +316,8 @@ from an instance of `NttManager` with rate-limiting enabled to `NttManagerNoRate
 
 #### Per-Chain-Transceivers
 
-The `NttManagerWithPerChainTransceivers` contract which inherits from `NttManagerNoRateLimiting` and allows configuring different transceivers and thresholds for each chain.
-You can configure a different set of send and receive transceivers for each chain. If you don't specifically enable any transceivers for a chain, then all transceivers will be used.
+The `NttManagerWithPerChainTransceivers` contract inherits from `NttManagerNoRateLimiting` and allows configuring different transceivers and thresholds for each chain.
+You can configure a different set of send and receive transceivers for each chain, as well as the receive threshold for the chain.
```
banescusebi, line 320: Something that might be worth adding to the README.md: 

Even when only using per chain transceivers, one needs to register each transceiver using the `setTransceiver()` method inherited from `ManagerBase`. Note that this only allows registering up to 64 transceivers in total (to be used as send and/or receive transceivers).

banescusebi (commented):
Overall looks pretty solid this way. Just a few nits and questions from my side.

evm/src/NttManager/NttManagerWithPerChainTransceivers.sol
```diff
@@ -27,61 +73,184 @@ contract NttManagerWithPerChainTransceivers is NttManagerNoRateLimiting {
...
+    // =============== Internal Interface Overrides ===================================================
+
     function _isSendTransceiverEnabledForChain(
         address transceiver,
         uint16 forChainId
     ) internal view override returns (bool) {
         uint64 bitmap = _getPerChainTransceiverBitmap(forChainId, SEND_TRANSCEIVER_BITMAP_SLOT);
         uint8 index = _getTransceiverInfosStorage()[transceiver].index;
-        return (bitmap & uint64(1 << index)) != 0;
+        return (bitmap & uint64(1 << index)) > 0;
```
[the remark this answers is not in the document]
djb15: This was me! `!=` needs 2 opcodes vs `>` just needs 1, so we save (a small amount of) gas here. Plus it matches how we do it elsewhere in the codebase

evm/src/NttManager/NttManagerWithPerChainTransceivers.sol
```diff
@@ -27,61 +73,184 @@ contract NttManagerWithPerChainTransceivers is NttManagerNoRateLimiting {
...
+    }
+
+    /// @inheritdoc IManagerBase
+    function isMessageApproved(
+        bytes32 digest
+    ) public view override(ManagerBase, IManagerBase) returns (bool) {
+        uint16 sourceChainId = _getMessageAttestationsStorage()[digest].sourceChainId;
+        uint8 threshold = getThresholdForChain(sourceChainId);
+        return messageAttestations(digest) >= threshold && threshold > 0;
```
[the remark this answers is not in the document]
djb15: Is this an optimisation? Threshold of 0 would be the rare edge case, so surely we'd want to fail early on checking message attestations > threshold before bothering to check that threshold is non-zero?

PR #519, opened 2024-09-18
EVM: NttManagerWithoutRateLimiting

bruce-riley:
This PR breaks up the EVM NTT Manager into two components, the existing `NttManager`, which works as always and a new `NttManagerWithoutRateLimiting`, which does not do rate limiting. This allows integrators to chose whether or not they wish to use rate limiting.

This is done by adding an abstract contract called `NttManagerBase` which contains most of the implementation, except for rate limiting. 

The `NttManagerWithoutRateLimiting` does not implement rate limiting, but it does implement that `IRateLimiter` interface. This is being done to maintain ABI compatibility between the two contracts.

This change increases the size of the `NttManager contract` by 67 bytes (leaving 90 bytes free).
The `NttManagerWithoutRateLimiting` contract has 6689 bytes free.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,6 @@ node_modules
 
 # Tilt
 .wormhole
+
+# Code coverage
+lcov.info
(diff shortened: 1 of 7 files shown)


bruce-riley:
I verified that the output from these commands match the output from the NttManager contract on main.

evan-gray:
Closing this in favor of #520 which supports a smaller diff and an easier migration strategy (for those who disabled rate limiting)

PR #522, opened 2024-09-25
NativeTokenTransfer additional payload

evan-gray:
This PR adds support for an additional, integrator-specified payload on the NativeTokenTransfer. It aims to be backwards compatible with the previous wire format when no additional payload is specified. Please see the runtime-specific READMEs for implementation-specific information.

The primary motivations for keeping the wire format semi-backwards compatible are
- Integrators who upgrade will _not_ have their wire format changed if they do not specify an additional payload.
- Ecosystem parsing will continue to support these integrators without changes
- Updated parsing can be made backwards compatible

As is, if the additional payload length were always written, the existing Solidity and TS SDK implementations would revert before being upgraded.

### EVM

This PR also includes changes to mark specific integration point functions as `virtual` for integrators to more easily customize the NttManager without forking.

Contract size before ([68a7ca4](https://github.com/wormhole-foundation/example-native-token-transfers/commit/68a7ca4132c74e838ac23e54752e8c0bc02bb4a2)):
```
| NttManager                            |   23,755 |        821 |
| NttManagerNoRateLimiting              |   18,403 |      6,173 |
```

After
```
| NttManager                              |   23,908 |        668 |
| NttManagerNoRateLimiting                |   17,150 |      7,426 |
```

### SVM

The existing Rust parsing (used by the Solana contract) does _not_ perform length checks - that means that if an additional payload is passed to a Solana contract, it will be interpreted as the transceiver payload. However, as far as I can tell from this repo, that transceiver payload is never used by the Solana code. Still, integrators should be careful when adding additional payload to update Solana first. The Solana implementation provided in this PR also _always_ expects the same additional payload format to be sent (empty by default).

Using a generic type on `NativeTokenTransfer` was the cleanest way I saw to keep the struct compatible with Anchor serialization / deserialization. Because the struct is written to an account, `additional_payload` could not simply be a `Vec<u8>`, as that would require a `max_len` and technically that value for integrators is `u16::MAX`, which would cause 64k+ allocation of accounts. This way, integrators can at least extend a fixed size account of their concrete type, which is empty by default.

### SDK

The TS SDK implementation provided is backwards compatible.


evm/src/NttManager/NttManager.sol
```diff
@@ -543,6 +571,10 @@ contract NttManager is INttManager, RateLimiter, ManagerBase {
         return seq;
     }
 
+    /// @dev Override this function to provide an additional payload on the NativeTokenTransfer
+    /// @return - The bytes to be sent as the additional payload
+    function _prepareAdditionalPayload() internal pure virtual returns (bytes memory) {}
```
djb15, line 576: I think you mentioned that no transfer context is needed to prepare the payload for the first use case of this feature, but does it make sense to pass in the key user provided info like `amount`, `recipient` and `recipientChain` as args here for the more general use case?

djb15 (commented):
Initial comments on EVM portion, will continue reviewing tomorrow

evm/src/NttManager/NttManager.sol
evan-gray: I agree, will update.

evm/src/libraries/TransceiverStructs.sol
```diff
@@ -146,6 +167,12 @@ library TransceiverStructs {
         (nativeTokenTransfer.sourceToken, offset) = encoded.asBytes32Unchecked(offset);
         (nativeTokenTransfer.to, offset) = encoded.asBytes32Unchecked(offset);
         (nativeTokenTransfer.toChain, offset) = encoded.asUint16Unchecked(offset);
+        if (encoded.length >= offset + 2) {
```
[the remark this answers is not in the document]
evan-gray: Sure thing, will update.

evm/src/NttManager/NttManager.sol
```diff
@@ -210,6 +210,26 @@ contract NttManager is INttManager, RateLimiter, ManagerBase {
...
+        bytes32 digest
+    ) internal virtual {
+        _handleTransfer(sourceChainId, message, digest);
+    }
+
+    function _handleTransfer(
+        uint16 sourceChainId,
+        TransceiverStructs.NttManagerMessage memory message,
+        bytes32 digest
+    ) internal virtual {
```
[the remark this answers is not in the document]
evan-gray: Correct. I’d like to add a test that ensures all these work, will try to get that in soon.

solana/modules/ntt-messages/src/ntt.rs
```diff
@@ -85,6 +102,53 @@ impl Writeable for NativeTokenTransfer {
         to.write(writer)?;
         to_chain.write(writer)?;
 
+        if A::SIZE != Some(0) {
+            let len: u16 = u16::try_from(additional_payload.written_size()).expect("u16 overflow");
+            len.write(writer)?;
+            // TODO: same as above
```
banescusebi, line 108: This seems like a TODO to be fixed in a future PR. Should there be a longer sentence here indicating what needs to be done?

solana/modules/ntt-messages/src/ntt.rs
```diff
@@ -50,22 +51,37 @@ impl Readable for NativeTokenTransfer {
         let to = Readable::read(reader)?;
         let to_chain = Readable::read(reader)?;
 
+        if A::SIZE != Some(0) {
+            // if the size is explicitly zero, this is an empty payload message
+            // and the size field should be skipped
+            // TODO: same as below for ntt additional_payload
```
banescusebi, line 57: This seems like a TODO to be fixed in a future PR. Should there be a longer sentence here indicating what needs to be done?

banescusebi (commented):
Small clarification questions and suggestions

evm/README.md
```diff
@@ -197,9 +199,35 @@ $ anvil --help
...
+## Message Customization
+
+See the [NttManager](../docs/NttManager.md) doc for wire format details.
+
+### NativeTokenTransfer Additional Payload
+
+Your contract can extend `NttManagerNoRateLimiting` to provide an additional payload on the `NativeTokenTransfer` message. Override the following:
+
+- `_prepareAdditionalPayload`
+- `_handleAdditionalPayload`
```
banescusebi, line 211: It could be worth adding a note here to point out that `NttManager` may be extended if rate limiting is off, but the size of `NttManager` is significantly higher (and very close to the contract size limit) than `NttManagerNoRateLimiting`. This is pretty clear from the PR description but it's not clear to a reader of this README who doesn't have that context.

banescusebi (commented):
One comment added together with Dirk over a discussion from our call.

solana/modules/ntt-messages/src/ntt.rs
evan-gray: This refers to a more verbose todo in `transceiver.rs` (and was copied from `ntt.rs`). Updating.

solana/modules/ntt-messages/src/ntt.rs
evan-gray: This refers to a more verbose todo in `transceiver.rs` (and was copied from `ntt.rs`). Updating.

evm/src/NttManager/NttManager.sol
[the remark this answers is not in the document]
evan-gray: Actually, I feel like they can just override `_handleMsg` if they want to use their own `_handleTransfer` which should drop this internal function anyway. I had documented that use case above, just didn't update this.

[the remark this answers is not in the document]
djb15: Nice! I think dropping `virtual` is the safest option here so that's great

evm/src/NttManager/NttManager.sol
evan-gray: Updated the function to return the NTT struct which _might_ be more effective? lmk your thoughts

evm/README.md
```diff
@@ -197,9 +199,37 @@ $ anvil --help
...
+
+See the [NttManager](../docs/NttManager.md) doc for wire format details.
+
+> Note: The size of `NttManager` is significantly higher (and very close to the contract size limit) than `NttManagerNoRateLimiting`. See [No-Rate-Limiting](#no-rate-limiting) for more detail.
+
+### NativeTokenTransfer Additional Payload
+
+Your contract can extend `NttManagerNoRateLimiting` to provide an additional payload on the `NativeTokenTransfer` message. Override the following:
+
+- `_prepareAdditionalPayload`
```
djb15, line 212: Should be updated to `_prepareNativeTokenTransfer` now

solana/modules/ntt-messages/src/ntt.rs
```diff
@@ -50,22 +51,37 @@ impl Readable for NativeTokenTransfer {
...
             + TrimmedAmount::SIZE.unwrap()
             + self.source_token.len()
             + self.to.len()
             + ChainId::SIZE.unwrap()
+            + if A::SIZE != Some(0) {
+                u16::SIZE.unwrap()
+            } else {
+                0
+            }
+            + self.additional_payload.written_size()
```
djb15, line 84: Thoughts about having this last line in the `if` block? Not sure if that's more/less explicit

djb15 (commented):
Is there anything that needs to be done on our side for handling custom manager payloads on Solana? Or are we leaving it to the integrator to write their own prefix handling from scratch? Maybe worth pointing to the best place to do so in the README.

And I guess similarly for the additional payload case, worth pointing integrators to the best place to do this? Or is the expectation that they'd write their own program that reads from the relevant account containing the transfer payload?

evm/src/NttManager/NttManager.sol
djb15: Nice idea, that feels a bit cleaner to me

solana/modules/ntt-messages/src/ntt.rs
banescusebi: I think it makes sense to add it in the `if` block from a code readability perspective.

evm/README.md
banescusebi, line 212:
```suggestion
- `_prepareNativeTokenTransfer`
```

banescusebi: I was confused initially by the comment because I thought you meant the `_handleAdditionalPayload` should be updated to `_prepareNativeTokenTransfer`. So I added the suggested change above for clarity.

solana/modules/ntt-messages/src/ntt.rs
evan-gray: Agree, it also then mirrors the write code.

evan-gray:
@djb15 I added a note about where to generate and handle the custom payloads on Solana. From a quick look, custom manager payloads seems a bit complex for a simple readme section. I would recommend waiting for the Manager/Transceiver program split before attempting that one.

banescusebi (approved):

nvsriram (approved):

djb15 (approved):

evm/src/libraries/TransceiverStructs.sol
```diff
@@ -112,6 +117,22 @@ library TransceiverStructs {
         // The `amount` and `decimals` fields are encoded in reverse order compared to how they are declared in the
         // `TrimmedAmount` type. This is consistent with the Rust NTT implementation.
         TrimmedAmount transferAmount = m.amount;
+        if (m.additionalPayload.length > 0) {
```
toninorair, line 120:
```suggestion
       uint16 additionalPayloadLength = uint16(m.additionalPayload.length)
```
Move assignment to line 120

banescusebi (commented):
Just one comment that I think could help improve maintainability and also help integrators

djb15 (dismissed):
Lgtm. Agreed on the docstring suggestion above

evm/src/NttManager/NttManager.sol
```diff
@@ -573,17 +583,22 @@ contract NttManager is INttManager, RateLimiter, ManagerBase {
     }
 
     /// @dev Override this function to provide an additional payload on the NativeTokenTransfer
+    /// For integrator flexibility, this function is *not* marked pure or view
     /// @param amount TrimmedAmount of the transfer
     /// @param token Address of the token that this NTT Manager is tied to
     /// @param recipient The recipient address
     /// @param recipientChain The Wormhole chain ID for the destination
+    /// @param - The sequence number for the manager message (unused, provided for overriding integrators)
+    /// @param - The message sender (unused, provided for overriding integrators)
```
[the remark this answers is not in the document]
evan-gray: accepted, thanks!

evm/test/mocks/MockNttManagerAdditionalPayload.sol
```diff
@@ -0,0 +1,41 @@
...
+
+    function _prepareNativeTokenTransfer(
+        TrimmedAmount amount,
+        address token,
+        bytes32 recipient,
+        uint16 recipientChain,
+        uint64, // sequence
+        address // sender
+    ) internal override returns (TransceiverStructs.NativeTokenTransfer memory) {
+        bytes memory additionalPayload = abi.encodePacked("banana");
```
djb15, line 25: 🍌

djb15 (approved):

solana/programs/example-native-token-transfers/src/transfer.rs
```diff
@@ -0,0 +1,3 @@
+use ntt_messages::{ntt::EmptyPayload, ntt::NativeTokenTransfer};
+
+pub type NativeTokenTransferConcrete = NativeTokenTransfer<EmptyPayload>;
```
kcsongor, line 3: naming nit: what's the rationale for calling this `Concrete`? Personally I think inlining the RHS everywhere would improve readability

kcsongor (approved):

solana/programs/example-native-token-transfers/src/transfer.rs
evan-gray: It’s a concrete type of the, now generic, NativeTokenTransfer. I was trying to avoid an integrator having to update a bunch of places in the code should they want to use an additional payload. I also figured this approach makes it safer, since it is unsafe (with the current implementation) to mix up the types used (where `EmptyPayload` currently is).

kcsongor: I'm still not a fan of the name, because every instantiation is going to be concrete, so calling this one the concrete one is pretty arbitrary.
`NativeTokenTransferNoPayload` might be a better candidate.

This is cosmetic so let's do it in a follow up PR which should be easier to review than re-reviewing this one.

evan-gray: https://github.com/wormhole-foundation/example-native-token-transfers/issues/526

PR #536, opened 2024-10-17
evm: update _prepareNativeTokenTransfer params

evan-gray:
This PR updates the parameters for `_prepareNativeTokenTransfer` 
- remove `token`, as it is an immutable of `ManagerBase.sol` and always accessible.
- add `refundAddress` in case an integrator chooses to reference that in their additional payload.


kcsongor (commented):
makes sense

kcsongor (approved):

nik-suri (approved):
lgtm

PR #546, opened 2024-11-06
evm: Emit events when updating inbound and outbound limits

nvsriram:
This PR adds:
* `OutboundTransferLimitUpdated` and `InboundTransferLimitUpdated` events
* Emits these events when `_setOutboundLimit` and `_setInboundLimit` are called respectively.

### DIFF
--- a/evm/src/interfaces/IRateLimiterEvents.sol
+++ b/evm/src/interfaces/IRateLimiterEvents.sol
@@ -25,4 +25,26 @@ interface IRateLimiterEvents {
     event OutboundTransferRateLimited(
         address indexed sender, uint64 sequence, uint256 amount, uint256 currentCapacity
     );
+
+    /// @notice Emitted when the outbound transfer limit is updated.
+    /// @dev Topic0
+    ///      0x7e3b0fc388be9d36273f66210aed83be975df3a9adfffa4c734033f498f362cd.
+    /// @param oldLimit The old outbound limit.
+    /// @param newLimit The new outbound limit.
+    event OutboundTransferLimitUpdated(
+        uint256 oldLimit,
+        uint256 newLimit
+    );
+ 
+    /// @notice Emitted when the inbound transfer limit is updated.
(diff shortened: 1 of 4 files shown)


evm/src/interfaces/IRateLimiterEvents.sol
```diff
@@ -25,4 +25,19 @@ interface IRateLimiterEvents {
...
+    /// @param newLimit The new outbound limit.
+    event OutboundTransferLimitUpdated(uint256 oldLimit, uint256 newLimit);
+
+    /// @notice Emitted when the inbound transfer limit is updated.
+    /// @dev Topic0
+    ///      0x739ed886fd81a3ddc9f4b327ab69152e513cd45b26fda0c73660eaca8e119301.
+    /// @param chainId The chain ID the limit is set for.
+    /// @param oldLimit The old inbound limit.
+    /// @param newLimit The new inbound limit.
+    event InboundTransferLimitUpdated(uint16 indexed chainId, uint256 oldLimit, uint256 newLimit);
```
nik-suri, line 42: Curious why you made `chainId` an `indexed` parameter here? 

cc @djb15 for thoughts as well

djb15: I'm guessing so that it's easier to search for the inbound limit updates for specific chain ids?

djb15 (approved):

evm/src/interfaces/IRateLimiterEvents.sol
nvsriram: Yep, and I think other events like [PeerUpdated](https://github.com/wormhole-foundation/example-native-token-transfers/blob/main/evm/src/interfaces/INttManager.sol#L50) also have `chainId` indexed. `InboundTransferLimitUpdated` event would be emitted when `PeerUpdated` is emitted so it felt right to have both events indexed by the same thing.

nik-suri (approved):

PR #550, opened 2024-11-13
solana sdk: Fix getIsApproved to check inbox item

kev1n-peters:
getIsApproved() was returning true when the VAA was posted, causing track() to set the receipt state to DestinationInitiated prematurely. It should check if the inbox item is initialized instead.

Fixes an issue that caused calls to NttAutomaticRoute.complete() to return early and never actually complete the transfer in rare cases (receipt.DestinationInitiated === true and inbox item not created).

### DIFF
--- a/solana/ts/sdk/ntt.ts
+++ b/solana/ts/sdk/ntt.ts
@@ -800,7 +800,9 @@ export class SolanaNtt<N extends Network, C extends SolanaChains>
   }
 
   async getIsApproved(attestation: Ntt.Attestation): Promise<boolean> {
-    if (attestation.payloadName !== "WormholeTransfer") return false;
+    if (attestation.payloadName !== "WormholeTransfer") {
+      throw new Error(`Invalid payload: ${attestation.payloadName}`);
+    }
     const payload = attestation.payload["nttManagerPayload"];
     try {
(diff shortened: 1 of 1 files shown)


solana/ts/sdk/ntt.ts
```diff
@@ -800,14 +800,20 @@ export class SolanaNtt<N extends Network, C extends SolanaChains>
...
   async getIsApproved(attestation: Ntt.Attestation): Promise<boolean> {
-    const digest = (attestation as WormholeNttTransceiver.VAA).hash;
-    const vaaAddress = utils.derivePostedVaaKey(
-      this.core.address,
-      Buffer.from(digest)
-    );
-
-    const info = await this.connection.getAccountInfo(vaaAddress);
-    return info !== null;
+    if (attestation.payloadName !== "WormholeTransfer") return false;
```
artursapek, line 803: maybe better to throw here? this would indicate a bug in the calling code, right?

kev1n-peters: i don't see any harm in throwing, i was just following the pattern used elsewhere but i'll change it

artursapek (approved):

nvsriram (approved):

PR #560, opened 2024-11-27
sdk: quote destination amount truncation fix

kev1n-peters:
Fixes an issue where calling quote with an amount that has more fractional digits than the destination chain supports throws an error. We need to truncate and scale the destination amount to destination chain decimals.

### DIFF
--- a/sdk/route/src/types.ts
+++ b/sdk/route/src/types.ts
@@ -18,6 +18,8 @@ export namespace NttRoute {
   // Currently only wormhole attestations supported
   export type TransceiverType = "wormhole";
 
+  export const TRIMMED_DECIMALS = 8;
+
   export type TransceiverConfig = {
     type: TransceiverType;
     address: string;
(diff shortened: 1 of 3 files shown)


sdk/route/src/types.ts
```diff
@@ -212,4 +214,16 @@ export namespace NttRoute {
...
   }
+
+  export function getTrimmedAmount(
+    amt: amount.Amount,
+    dstTokenDecimals: number
+  ): amount.Amount {
+    // remove dust to avoid `TransferAmountHasDust` revert reason
+    const truncatedAmount = amount.truncate(
+      amt,
+      Math.min(amt.decimals, dstTokenDecimals, NttRoute.TRIMMED_DECIMALS)
```
kev1n-peters, line 225: See: https://github.com/wormhole-foundation/native-token-transfers/blob/ef24f9f0065c22ec8f51775ed059501c45bbf152/solana/modules/ntt-messages/src/trimmed_amount.rs#L90

nvsriram (approved):

artursapek (approved):

PR #563, opened 2024-11-30
tilt: solana-test-validator multi-platform support

evan-gray:
This PR adds support for solana-test-validator within Docker / Tilt for M-series Macs. See also https://github.com/wormhole-foundation/wormhole/pull/4175 and https://github.com/wormhole-foundation/wormhole/pull/4181 which made this possible.

It also changes the `sed` line to one that works on Mac and Linux / WSL. To confirm that line still works in ci, I did a run with the guardian log level on `info` [here](https://github.com/wormhole-foundation/native-token-transfers/actions/runs/12204869868/job/34050904852?pr=563) where you can search for `8bf0b547c96edc5c1d512ca25c5c1d1812a180438a0046e511d1fb61561d5cdf` and the other emitters added to confirm they are being accounted (search for `gacct` guardian logs).

Bonus: this changes the result of `solana/Dockerfile` to only contain the required artifacts, reducing the image size from 24.82 GB to 5.27 MB!

oh, and you can now run `tilt up -- --evm2 --generic_relayer --solana_watcher --wormchain --guardiand_loglevel=warn` on your Mac just like in CI and it passes!

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -2,7 +2,7 @@ load('ext://namespace', 'namespace_create', 'namespace_inject')
 load('ext://git_resource', 'git_checkout')
 
 git_checkout('https://github.com/wormhole-foundation/wormhole.git#main', '.wormhole/', unsafe_mode=True)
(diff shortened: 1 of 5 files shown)


evan-gray:
Should be passable on Mac with 4 CPU and 8 GB RAM allocated (at least for the run, not sure on the build which likely uses more). This test run was done with 8 CPU 24 GB RAM.

image

nvsriram (dismissed):

johnsaigle (commented):
Changes look OK but it seems like the solana tests are failing.

nvsriram (approved):

nik-suri:
@johnsaigle the CLI check is not a required check. There is an open PR from @nvsriram to fix this: https://github.com/wormhole-foundation/native-token-transfers/pull/573

Issue #504, opened 2024-08-27
Emit source chain in the destination chain redeem/MsgAttestedTo events


nik-suri:
Emitting the source chain in these events can help 3rd party systems quickly identify where a certain transfer is coming from. Otherwise, 3rd party systems need to do extra work to find the body of the message and properly deserialize it to obtain the source chain.

PR #566, opened 2024-12-03
evm: Add source chain topic to `TransferRedeemed` and `MessageAttestedTo` events

nvsriram:
This PR adds:
* `sourceChain` topic to `MessageAttestedTo` and `TransferRedeemed` events
* Update event emit with propagated `sourceChain`
* Update test and docs to account for new event signature


evm/src/NttManager/ManagerBase.sol
```diff
@@ -423,7 +423,7 @@ abstract contract ManagerBase is
         _setTransceiverAttestedToMessage(digest, _getTransceiverInfosStorage()[transceiver].index);
 
         emit MessageAttestedTo(
-            digest, transceiver, _getTransceiverInfosStorage()[transceiver].index
+            chainId, digest, transceiver, _getTransceiverInfosStorage()[transceiver].index
```
djb15, line 426: Isn't `chainId` the current chain, not the source chain id of the message?

evm/src/interfaces/IRateLimiter.sol
```diff
@@ -57,13 +57,15 @@ interface IRateLimiter {
...
     /// @dev
+    ///    - sourceChain: the chain of the sender.
     ///    - recipient: the recipient of the transfer.
     ///    - amount: the amount of the transfer, trimmed.
     ///    - txTimestamp: the timestamp of the transfer.
     ///    - recipientChain: the chain of the recipient.
     ///    - sender: the sender of the transfer.
     ///    - transceiverInstructions: additional instructions to be forwarded to the recipient chain.
     struct OutboundQueuedTransfer {
+        uint16 sourceChain;
```
djb15, line 68: Any new fields in a struct should be added at the end of the struct. Same goes for below

djb15 (commented):
High level Q: what's the motivation for this change? Why do we need the source chain id in these events?

evm/src/NttManager/ManagerBase.sol
nvsriram: You're right, `sourceChainId` is not in the scope of `_setTransceiverAttestedToMessage` function so will pass it in as a param from the calling function.

nvsriram:
> High level Q: what's the motivation for this change? Why do we need the source chain id in these events?

Main context behind the change is to address this issue: https://github.com/wormhole-foundation/native-token-transfers/issues/504.

djb15 (commented):
Is there the intention for integrators to potentially upgrade existing deployments?

I'd love to see a couple of tests that involve having queued transfers (both inbound and outbound) with a previous deployment, before upgrading to this functionality and trying to release the queued transfers. You may have to do this by manually writing into a slot with the old struct format to fake a queued transfer before an upgrade; that way your test doesn't have to deal with upgrades directly.

For reference, this is why I suggested you move the the new fields to the end of the structs as otherwise the reads of existing queued transfers will be borked; moving to the end of the struct should work but a test to verify would be great.

PR #568, opened 2024-12-03
solana: Add SPL multisig support

nvsriram:
This PR adds:
* `initialize_multisig` and `release_inbound_multisig_mint` to act as multisig variants of `initialize` and `release_inbound_mint` respectively
* Refactor out common structs and function logic to avoid duplication
* Update test to verify independent minting capability after transferring authority to multisig
* Update IDL


kcsongor (commented):
overall looks good, but most of the code is duplication from the original versions. I wonder if we could refactor to share the logic?

solana/ts/sdk/ntt.ts
```diff
@@ -640,17 +641,30 @@ export class SolanaNtt<N extends Network, C extends SolanaChains>
...
+          this.program,
+          {
+            ...args,
+            payer,
+            owner: payer,
+            chain: this.chain,
+            tokenProgram: mintInfo.owner,
+          },
+          this.pdas
+        );
```
nonergodic, line 667:
```suggestion
const createInstruction = args.multisig
  ? NTT.createInitializeMultisigInstruction
  : NTT.createInitializeInstruction;

const ix = await createInstruction(
  this.program,
  {
    ...args,
    payer,
    owner: payer,
    chain: this.chain,
    tokenProgram: mintInfo.owner,
    ...(args.multisig && { multisig: args.multisig }),
  },
  this.pdas
);
```
Making this DRYer - not sure if the suggestion actually works, but it gets the idea across.

solana/ts/sdk/ntt.ts
```diff
@@ -987,12 +1005,19 @@ export class SolanaNtt<N extends Network, C extends SolanaChains>
           chain: emitterChain,
           revertOnDelay: false,
         };
-        const releaseIx =
+        let releaseIx =
           config.mode.locking != null
-            ? NTT.createReleaseInboundUnlockInstruction(
+            ? NTT.createReleaseInboundUnlockInstruction(this.program, config, {
```
nonergodic, line 1010: I guess there's multiple places that could be DRYer.

nonergodic (dismissed):
Only nits.

solana/ts/sdk/ntt.ts
nvsriram: This is definitely easier to follow but TS screams at me with an error when I tried this:
```
...
Types of property 'multisig' are incompatible.
   Type 'PublicKey | undefined' is not assignable to type 'PublicKey'.
      Type 'undefined' is not assignable to type 'PublicKey'.
```
I tried passing the `multisig` as not part of `args` object but even then it had the same error.

nvsriram: The issue seems to be that the condition deciding`createInstruction` (i.e., `args.multisig`) is not "propagated" when `createInstruction` is used. 
So despite using the same condition (without any changes to `args.multisig` in between), TS is not able to figure out which version of `createInstruction` is used (it assumes both could be used) and thus gives me the error.

solana/programs/example-native-token-transfers/src/instructions/release_inbound.rs
```diff
@@ -141,12 +134,87 @@ pub fn release_inbound_mint<'info>(
...
+        ctx.accounts.common.token_authority.to_account_info(),
+        ctx.remaining_accounts,
+        inbox_item.amount,
+        ctx.accounts.common.mint.decimals,
+        &[&[
+            crate::TOKEN_AUTHORITY_SEED,
+            &[ctx.bumps.common.token_authority],
+        ]],
+    )?;
+    Ok(())
```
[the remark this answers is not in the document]
nvsriram: I don't think that is possible without typecasting as `onchain::invoke_transfer_checked` returns `ProgramResult` but we expect our function to return `Result`.

solana/tests/anchor.test.ts
```diff
@@ -285,21 +295,26 @@ describe("example-native-token-transfers", () => {
...
       transaction.sign(payer);
-      await sendAndConfirmTransaction(connection, transaction, [payer], {
-        commitment: "confirmed",
-      });
+      await anchor.web3.sendAndConfirmTransaction(
+        connection,
+        transaction,
+        [payer],
+        {
+          commitment: "confirmed",
```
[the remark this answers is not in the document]
nvsriram: I usually prefer an explicit commitment level for more transparency but if removing it makes it easier to follow, then I can remove it.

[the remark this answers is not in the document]
nonergodic: It's not about making something easier to follow, but rather things can get mixed up. If you instantiate your connection to use a different commitment level then all hell can break loose. E.g. you query blockhashes that exist for a lower level but aren't yet known at a higher finality level. By always relying on the commitment level of the connection, you avoid this sort of mishap (and I've seen this happen in practice several times where you get rather vexing errors because the code looks like everything should be fine).

solana/ts/sdk/ntt.ts
nonergodic: Yeah, I figured that there might be a TS issue there (hence my uncertainty) but I'd assume there's still a way to do away with a bunch of repetition. Overall it's not a big deal ofc.

solana/programs/example-native-token-transfers/src/instructions/release_inbound.rs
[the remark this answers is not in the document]
nonergodic: Oh, I didn't consider that. I failed to realize that the ? operator performs the `.from` conversion.

solana/tests/anchor.test.ts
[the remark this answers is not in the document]
nvsriram: I agree. That makes sense, thanks for clarifying!

solana/tests/anchor.test.ts
nonergodic, line 1: Seems slightly sketch that `TOKEN_PROGRAM` is hardcoded to `spl.TOKEN_2022_PROGRAM_ID`. Arguably the tests should try both variants to ensure that there aren't any hidden dependencies on token 2022 (anything that's valid in the old token program should automatically be fine in the new one ofc).

nonergodic (approved):
Again, only nits. If you decide to address them, I promise I'll be quick with rubberstamping those changes this time around.

solana/programs/example-native-token-transfers/src/spl_multisig.rs
```diff
@@ -0,0 +1,37 @@
...
+/// Thus, we have to write our own wrapper to use with `InterfaceAccount`
+
+#[derive(Clone, Debug, Default, PartialEq)]
+pub struct SplMultisig(spl_token_2022::state::Multisig);
+
+impl AccountDeserialize for SplMultisig {
+    fn try_deserialize_unchecked(buf: &mut &[u8]) -> anchor_lang::Result<Self> {
+        spl_token_2022::state::Multisig::unpack(buf)
+            .map(SplMultisig)
+            .map_err(Into::into)
```
nonergodic, line 16:
```suggestion
        Ok(SplMultiSig(spl_token_2022::state::Multisig::unpack(buf)?))
```

solana/programs/example-native-token-transfers/src/spl_multisig.rs
nonergodic, line 1: This entire thing should probably go somewhere else. If it isn't in the latest Anchor version, it would be good to PR it there too.

nonergodic (commented):
Two more remarks.

solana/programs/example-native-token-transfers/src/instructions/initialize.rs
```diff
@@ -96,24 +90,77 @@ pub struct InitializeArgs {
...
-        bump: ctx.bumps.config,
-        mint: ctx.accounts.mint.key(),
-        token_program: ctx.accounts.token_program.key(),
-        mode: args.mode,
-        chain_id: ChainId { id: args.chain_id },
-        owner: ctx.accounts.deployer.key(),
+    // NOTE: this check was moved into the function body to reuse the `Initialize` struct
+    // in the multisig variant while preserving ABI
+    if !(args.mode == Mode::Locking
+        || ctx.accounts.mint.mint_authority.unwrap() == ctx.accounts.token_authority.key())
```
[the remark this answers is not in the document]
nvsriram: The reasoning behind this is to match the constraint but I agree the latter is more comprehensible without that context.

solana/programs/example-native-token-transfers/src/spl_multisig.rs
nvsriram: It isn't in the latest [anchor-spl](https://docs.rs/anchor-spl/latest/anchor_spl/token_interface/index.html) version and there is also an open issue https://github.com/coral-xyz/anchor/issues/3134 for the same.
According to the issue, it is not being implemented as it would cause breaking changes. So I don't think it makes sense to bloat the `token_interface` with a struct that is never used/ has similar helper support as the rest.

In this case though, I could move the file to anywhere else you think makes more sense.

solana/tests/anchor.test.ts
nvsriram: I found no easy way to test both variants in an integration test. I [tried it previously](https://github.com/wormhole-foundation/native-token-transfers/pull/488/files) by running multiple `*.test.ts` tests via a script but it wasn't successful.

It is hardcoded in this case but I am treating it as a configurable parameter in local testing (like `VERSION`). I will move it up so it is in a more consistent position. However, if there's an easier way to achieve this that I missed, I can give that a try.

solana/tests/anchor.test.ts
```diff
@@ -24,12 +24,6 @@ import {
 import { SolanaWormholeCore } from "@wormhole-foundation/sdk-solana-core";
 import * as fs from "fs";
 
-import {
-  PublicKey,
-  SystemProgram,
-  Transaction,
-  sendAndConfirmTransaction,
-} from "@solana/web3.js";
```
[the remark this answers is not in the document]
nvsriram: This doesn't compile as is - it treats `PublicKey` as `typeof anchor.web3.PublicKey` when trying to destructure.

solana/programs/example-native-token-transfers/src/instructions/release_inbound.rs
```diff
@@ -62,6 +63,7 @@ pub struct ReleaseInboundArgs {
 pub struct ReleaseInboundMint<'info> {
     #[account(
         constraint = common.config.mode == Mode::Burning @ NTTError::InvalidMode,
+        constraint = common.mint.mint_authority.unwrap() == common.token_authority.key()
```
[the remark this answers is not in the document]
nvsriram: I agree, I believe the early checks were not there previously so will remove it as they are redundant.

solana/programs/example-native-token-transfers/src/spl_multisig.rs
nonergodic: Well, the [wormhole rust sdk](https://github.com/wormhole-foundation/wormhole-sdk-rs) would be a good candidate if nothing else, if only it wasn't essentially abandoned. So I don't have a good answer except "somebody has to do the work".

nonergodic (approved):
:+1:

solana/programs/example-native-token-transfers/src/instructions/release_inbound.rs
```diff
@@ -77,19 +78,7 @@ pub fn release_inbound_mint<'info>(
...
-    if !released {
-        if args.revert_on_delay {
-            return Err(NTTError::CantReleaseYet.into());
-        } else {
-            return Ok(());
-        }
-    }
-
-    assert!(inbox_item.release_status == ReleaseStatus::Released);
+    let inbox_item = release_inbox_item(&mut ctx.accounts.common.inbox_item, args.revert_on_delay)?;
```
kcsongor, line 81: we have to be careful here! the original code had an early return on the `!released && !revert_on_delay` code path, but now that the conditional is moved into the `release_inbox_item` function, only that function will early return, making this call succeed in that case, and the function proceed.
This in turn bypasses the rate limit and the transceiver threshold (since it would allow releasing funds when try_release returns false on the `NotApproved` branch)

The `assert!(inbox_item.release_status == ReleaseStatus::Released);` in the original code was unreachable unless we introduced a logical bug. We should move that back into this function below `release_inbox_item` just to be extra sure (and right now it would actually be reachable)

(It also means we likely don't have test coverage for this)

solana/programs/example-native-token-transfers/src/instructions/initialize.rs
```diff
@@ -37,12 +37,6 @@ pub struct Initialize<'info> {
     )]
     pub config: Box<Account<'info, crate::config::Config>>,
 
-    #[account(
-        constraint =
-            args.mode == Mode::Locking
-            || mint.mint_authority.unwrap() == token_authority.key()
-            @ NTTError::InvalidMintAuthority,
-    )]
     pub mint: Box<InterfaceAccount<'info, token_interface::Mint>>,
```
kcsongor, line 40: maybe add a note here that the mint is unconstrained in the accounts struct and is the responsibility of the handler to constrain it

solana/programs/example-native-token-transfers/src/instructions/release_inbound.rs
```diff
@@ -133,10 +124,82 @@ pub fn release_inbound_mint<'info>(
...
+            @ NTTError::InvalidMultisig,
+    )]
+    pub multisig: InterfaceAccount<'info, SplMultisig>,
+}
+
+pub fn release_inbound_mint_multisig<'info>(
+    ctx: Context<'_, '_, '_, 'info, ReleaseInboundMintMultisig<'info>>,
+    args: ReleaseInboundArgs,
+) -> Result<()> {
+    let inbox_item = release_inbox_item(&mut ctx.accounts.common.inbox_item, args.revert_on_delay)?;
```
kcsongor, line 151: same as above, let's add back the assertion

solana/programs/example-native-token-transfers/src/spl_multisig.rs
```diff
@@ -0,0 +1,35 @@
+use anchor_lang::{prelude::*, solana_program::program_pack::Pack, Ids, Owners};
```
kcsongor, line 1: very cool!

kcsongor (changes requested):

solana/programs/example-native-token-transfers/src/instructions/release_inbound.rs
nvsriram: Ah, right!! Good catch!!
Just adding the `assert` would not be enough though as that would return early with an `Err` as opposed to `Ok`.

nvsriram: How do you feel about having `release_inbox_item` return a `Result<Option<_>>`? 
Like so:
```rust
fn release_inbox_item(
    inbox_item: &mut InboxItem,
    revert_on_delay: bool,
) -> Result<Option<&mut InboxItem>> {
    if inbox_item.try_release()? {
        Ok(Some(inbox_item))
    } else if revert_on_delay {
        Err(NTTError::CantReleaseYet.into())
    } else {
        Ok(None)
    }
}
```

And now I can handle it accordingly like so:
```rust
let inbox_item = release_inbox_item(&mut ctx.accounts.common.inbox_item, args.revert_on_delay)?;
if inbox_item.is_none() {
    return Ok(());
}
let inbox_item = inbox_item.unwrap();
assert!(inbox_item.release_status == ReleaseStatus::Released);
```

kcsongor (approved):

nonergodic (dismissed):

kcsongor (approved):

nonergodic (approved):

PR #569, opened 2024-12-03
sdk: automatic route is available fix

kev1n-peters:
Need to check if the route is actually available on the contract.

### DIFF
--- a/sdk/route/src/automatic.ts
+++ b/sdk/route/src/automatic.ts
@@ -114,6 +114,14 @@ export class NttAutomaticRoute<N extends Network>
     request: routes.RouteTransferRequest<N>,
     params: Tp
   ): Promise<Vr> {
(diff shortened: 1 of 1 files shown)


nvsriram (approved):

artursapek (commented):
hmmm... is it better to do this inside `quote`? I feel like it belongs in there. the relayer has nothing to do with validating the transfer parameters.

kev1n-peters:
> hmmm... is it better to do this inside `quote`? I feel like it belongs in there. the relayer has nothing to do with validating the transfer parameters.

I think it makes more sense to have in quote

artursapek (approved):

emreboga:
Fixes https://github.com/wormhole-foundation/wormhole-connect/issues/3030

nvsriram (approved):

evgeniko (approved):

PR #570, opened 2024-12-03
solana: Allow transferring mint authority

nvsriram:
This PR adds:
* Two ways to transfer out mint authority like `transfer_ownership`:
  * One step version: `set_token_authority_one_step_unchecked` ix
  * Two step version: `set_token_authority` and `claim_token_authority` ix (and `revert_token_authority` to reset authority back to `token_authority`)
* `createSetTokenAuthorityInstruction` helper function in `NTT` namespace
* Update IDL

### DIFF
--- a/solana/programs/example-native-token-transfers/src/instructions/admin.rs
+++ b/solana/programs/example-native-token-transfers/src/instructions/admin.rs
@@ -1,4 +1,5 @@
 use anchor_lang::prelude::*;
+use anchor_spl::{token_2022::spl_token_2022::instruction::AuthorityType, token_interface};
 use ntt_messages::chain_id::ChainId;
 use wormhole_solana_utils::cpi::bpf_loader_upgradeable::{self, BpfLoaderUpgradeable};
 
@@ -151,6 +152,51 @@ pub fn claim_ownership(ctx: Context<ClaimOwnership>) -> Result<()> {
             &[&[b"upgrade_lock", &[ctx.bumps.upgrade_lock]]],
         ),
         &crate::ID,
+    )
+}
+
+// * Set token authority
+#[derive(Accounts)]
+pub struct SetTokenAuthority<'info> {
+    #[account(
+        constraint = config.owner == owner.key()
+    )]
+    pub config: Account<'info, Config>,
+
+    pub owner: Signer<'info>,
+
+    #[account(
+        mut,
+        address = config.mint,
+    )]
+    /// CHECK: the mint address matches the config
+    pub mint: InterfaceAccount<'info, token_interface::Mint>,
+
+    pub token_program: Interface<'info, token_interface::TokenInterface>,
+
+    #[account(
+        seeds = [crate::TOKEN_AUTHORITY_SEED],
+        bump,
+    )]
+    /// CHECK: The seeds constraint enforces that this is the correct account.
+    pub token_authority: UncheckedAccount<'info>,
+
+    /// CHECK: This is unsafe as is so should be used with caution
+    pub new_authority: UncheckedAccount<'info>,
+}
+
+pub fn set_token_authority_one_step_unchecked(ctx: Context<SetTokenAuthority>) -> Result<()> {
+    token_interface::set_authority(
+        CpiContext::new_with_signer(
+            ctx.accounts.token_program.to_account_info(),
+            token_interface::SetAuthority {
+                account_or_mint: ctx.accounts.mint.to_account_info(),
+                current_authority: ctx.accounts.token_authority.to_account_info(),
+            },
+            &[&[crate::TOKEN_AUTHORITY_SEED, &[ctx.bumps.token_authority]]],
+        ),
+        AuthorityType::MintTokens,
+        Some(ctx.accounts.new_authority.key()),
     )
 }
 
--- a/solana/programs/example-native-token-transfers/src/instructions/admin.rs
+++ b/solana/programs/example-native-token-transfers/src/instructions/admin.rs
@@ -159,6 +159,43 @@ pub fn claim_ownership(ctx: Context<ClaimOwnership>) -> Result<()> {
 // * Set token authority
 
 #[derive(Accounts)]
+pub struct AcceptTokenAuthority<'info> {
+    #[account(
+        has_one = mint,
+        constraint = config.paused @ NTTError::NotPaused,
+    )]
+    pub config: Account<'info, Config>,
+
+    #[account(mut)]
+    pub mint: InterfaceAccount<'info, token_interface::Mint>,
+
+    #[account(
+        seeds = [crate::TOKEN_AUTHORITY_SEED],
+        bump,
+    )]
+    /// CHECK: The constraints enforce this is valid mint authority
+    pub token_authority: UncheckedAccount<'info>,
+
+    pub current_authority: Signer<'info>,
+
+    pub token_program: Interface<'info, token_interface::TokenInterface>,
+}
+
+pub fn accept_token_authority(ctx: Context<AcceptTokenAuthority>) -> Result<()> {
+    token_interface::set_authority(
+        CpiContext::new(
+            ctx.accounts.token_program.to_account_info(),
+            token_interface::SetAuthority {
+                account_or_mint: ctx.accounts.mint.to_account_info(),
+                current_authority: ctx.accounts.current_authority.to_account_info(),
+            },
+        ),
+        AuthorityType::MintTokens,
+        Some(ctx.accounts.token_authority.key()),
+    )
+}
+
+#[derive(Accounts)]
 pub struct SetTokenAuthority<'info> {
     #[account(
         has_one = owner,
@@ -170,7 +207,6 @@ pub struct SetTokenAuthority<'info> {
     pub owner: Signer<'info>,
 
     #[account(mut)]
-    /// CHECK: the mint address matches the config
     pub mint: InterfaceAccount<'info, token_interface::Mint>,
 
     #[account(
@@ -244,6 +280,8 @@ pub fn set_token_authority_one_step_unchecked(
     )
 }
 
+// * Claim token authority
+
 #[derive(Accounts)]
 pub struct RevertTokenAuthority<'info> {
     #[account(
@@ -253,7 +291,6 @@ pub struct RevertTokenAuthority<'info> {
     pub config: Account<'info, Config>,
 
     #[account(mut)]
-    /// CHECK: the mint address matches the config
     pub mint: InterfaceAccount<'info, token_interface::Mint>,
 
     #[account(
@@ -264,7 +301,7 @@ pub struct RevertTokenAuthority<'info> {
     pub token_authority: UncheckedAccount<'info>,
 
     #[account(mut)]
-    /// CHECK: the constraint enforces that this is the correct address
+    /// CHECK: the `pending_token_authority` constraint enforces that this is the correct address
     pub rent_payer: UncheckedAccount<'info>,
 
     #[account(
--- a/solana/programs/example-native-token-transfers/src/instructions/admin.rs
+++ b/solana/programs/example-native-token-transfers/src/instructions/admin.rs
@@ -10,6 +10,7 @@ use crate::{
     config::Config,
     error::NTTError,
     peer::NttManagerPeer,
+    pending_token_authority::PendingTokenAuthority,
     queue::{inbox::InboxRateLimit, outbox::OutboxRateLimit, rate_limit::RateLimitState},
     registered_transceiver::RegisteredTransceiver,
 };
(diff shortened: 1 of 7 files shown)


kcsongor (approved):
very nice!

solana/programs/example-native-token-transfers/src/instructions/admin.rs
nonergodic, line 1: Currently, there's still no way for the program to safely accept the token authority back, i.e.:
```
#[derive(Accounts)]
pub struct AcceptTokenAuthority<'info> {
    pub config: Account<'info, Config>,

    #[account(
        mut,
        address = config.mint,
    )]
    /// CHECK: the mint address matches the config
    pub mint: InterfaceAccount<'info, token_interface::Mint>,

    #[account(
        seeds = [crate::TOKEN_AUTHORITY_SEED],
        bump,
    )]
    /// CHECK: The constraints enforce this is valid mint authority
    pub token_authority: UncheckedAccount<'info>,

    /// CHECK: This account will be the signer in the [claim_token_authority] instruction.
    pub current_authority: Signer<'info>,
}

pub fn accept_token_authority(
    ctx: Context<AcceptTokenAuthority>,
) -> Result<()> {
    token_interface::set_authority(
        CpiContext::new(
            ctx.accounts.token_program.to_account_info(),
            token_interface::SetAuthority {
                account_or_mint: ctx.accounts.mint.to_account_info(),
                current_authority: ctx.accounts.current_authority.to_account_info(),
            }
        ),
        AuthorityType::MintTokens,
        Some(ctx.accounts.token_authority.key()),
    )
}
```

nonergodic, line 1: Currently, there's no enforcement that the program remains paused as long as the token authority is borrowed, i.e. the admin can transfer out the authority and then unpause.

solana/programs/example-native-token-transfers/src/instructions/admin.rs
```diff
@@ -154,6 +156,167 @@ pub fn claim_ownership(ctx: Context<ClaimOwnership>) -> Result<()> {
...
+#[derive(Accounts)]
+pub struct RevertTokenAuthority<'info> {
+    #[account(
+        constraint = config.paused @ NTTError::NotPaused,
+    )]
+    pub config: Account<'info, Config>,
+
+    #[account(
+        mut,
+        address = pending_token_authority.rent_payer @ NTTError::IncorrectRentPayer,
```
nonergodic, line 256: As a personal preference, I would rename the account to `rent_payer` and then check it with the `pending_token_authority` account via a `has_one` constraint. (A `payer` that is not a signer will raise eyebrows regardless, but having the name be `rent_payer` and hence at least slightly different from the idiomatic `payer` will help with the confusion.

Additionally, I'd move the account closer to the `pending_token_authority` account to make their grouping more intuitive.

solana/programs/example-native-token-transfers/src/instructions/admin.rs
```diff
@@ -154,6 +156,167 @@ pub fn claim_ownership(ctx: Context<ClaimOwnership>) -> Result<()> {
...
+        mut,
+        address = config.mint,
+    )]
+    /// CHECK: the mint address matches the config
+    pub mint: InterfaceAccount<'info, token_interface::Mint>,
+
+    #[account(
+        seeds = [crate::TOKEN_AUTHORITY_SEED],
+        bump,
+        address = mint.mint_authority.unwrap() @ NTTError::InvalidMintAuthority
```
nonergodic, line 181: I'd either get rid of this check (since it is ultimately unnecessary, and I don't like diluting code with unnecessary checks thus making it unclear what is or isn't critical), or at least also have it in the revert version so it is at least consistent.

solana/programs/example-native-token-transfers/src/instructions/admin.rs
```diff
@@ -154,6 +156,167 @@ pub fn claim_ownership(ctx: Context<ClaimOwnership>) -> Result<()> {
...
+        has_one = owner,
+        constraint = config.paused @ NTTError::NotPaused,
+    )]
+    pub config: Account<'info, Config>,
+
+    pub owner: Signer<'info>,
+
+    #[account(
+        mut,
+        address = config.mint,
```
nonergodic, line 173: should also be checked via a `has_one` constrained of the `config` account for consistency.

nonergodic (changes requested):

solana/programs/example-native-token-transfers/src/instructions/admin.rs
nvsriram: On startup, this is [done separately](https://github.com/wormhole-foundation/native-token-transfers/blob/main/solana/tests/anchor.test.ts#L240) via `spl.setAuthority` BEFORE calling `ntt.initialize`.

The `newAuthority` can just call `spl.setAuthority` again to give back the authority to the `tokenAuthority` without needing a separate instruction.

nvsriram: Yes, that's true.
1. I'm not really sure what the best way to enforce that would be - we could add a constraint on `unpause` but it still seems iffy. eg:- with https://github.com/wormhole-foundation/native-token-transfers/pull/568, the actual expected token authority is not a given.
2. All instructions that depend on the `token_authority` like transfers would just fail. So it should not cause any issues imo.
3. As resetting token authority is done via `spl.setAuthority`, I don't see a need to enforce the order of operations i.e., token authority be transferred back first and only then `unpause`. As long as token authority is transferred back before any instruction that requires it to be the case, it should be good. 
Having said that, we now could have cases where `tokenAuthority` being the authority when unpaused is not a given.

solana/programs/example-native-token-transfers/src/instructions/admin.rs
nvsriram: This is not unnecessary with regards to the `set_token_authority` (checked) ix as we init the `pending_token_authority` without doing any transfer so we need to ensure the mint authority is valid.
It is unnecessary wrt the `set_token_authority_one_step_unchecked` though as the token program will check itself.

I can remove this constraint from `SetTokenAuthority` and add it in the `SetTokenAuthorityChecked` struct to remove redundant checks like so:
```rust
#[derive(Accounts)]
pub struct SetTokenAuthorityChecked<'info> {
#[account(
    constraint = common.token_authority.key() == common.mint.mint_authority.unwrap() @ NTTError::InvalidMintAuthority
)]
pub common: SetTokenAuthority<'info>,
...
}
```

solana/programs/example-native-token-transfers/src/instructions/admin.rs
nvsriram: Would this not be redundant checks as we are checking the equality both ways?

I think the repo has cases of `address = config.mint` without `has_one` in [broadcast_id](https://github.com/wormhole-foundation/native-token-transfers/blob/main/solana/programs/example-native-token-transfers/src/transceivers/wormhole/instructions/broadcast_id.rs#L15), [release_inbound](https://github.com/wormhole-foundation/native-token-transfers/blob/main/solana/programs/example-native-token-transfers/src/instructions/release_inbound.rs#L39), etc. so I went with this check here. If there's a constraint preference in using `address` over `has_one` or vice versa, we can stick to that.

solana/programs/example-native-token-transfers/src/instructions/admin.rs
nonergodic: Yes, but that's an unsafe way to hand over the authority. If you specify the wrong address as the new authority, you are wrecked.

solana/programs/example-native-token-transfers/src/instructions/admin.rs
nonergodic: Sorry, my comment was imprecise. My "also" was not meant as "in addition to" but as in "also via `has_one` like the `owner`". Right now checking the `owner` one way and the `mint` another seems strange. I think that the `address` check has the advantage that it actually is located with the account that is being checked, while the `has_one` is more terse. So either both should use `has_one` or neither, I'd argue.

solana/programs/example-native-token-transfers/src/instructions/admin.rs
nonergodic: Yeah, good point.

solana/programs/example-native-token-transfers/src/instructions/admin.rs
nonergodic: Yeah, I don't have a strong recommendation either, just that atm the handling isn't actually consistent - ultimately the pause check is atm more a reminder that prevents you from tripping up. But otoh this should be handled by the contract's SDK not by the on-chain program.

solana/programs/example-native-token-transfers/src/instructions/admin.rs
nvsriram: That makes sense. For now, I'll arbitrarily choose the `has_one` constraint over the `address` and keep that the case for this file at least.

solana/programs/example-native-token-transfers/src/instructions/admin.rs
nvsriram: Makes sense. For the time being, I'll leave it as is (without enforcement) but we can come back to it later in a different PR directed specifically at this.

nvsriram: I agree it is technically unsafe but in practice, wouldn't users use the SDK exposed PDA helpers i.e., `ntt.pdas.tokenAuthority()`?
Yes, if they decided to not use these helpers and just use the actual `PublicKey`s instead, they could brick it up.

nvsriram: Thinking ahead, with https://github.com/wormhole-foundation/native-token-transfers/pull/568 we'd also need to add a similar `AcceptTokenAuthorityMultisig` and ensure the signature threshold is 1 and that the `tokenAuthority` PDA is one of the signers. But since there's no easy way to currently deserialize a SPL multisig account in Anchor, we'd have to write our own deserialization functions and use that to verify.

tl;dr: 
To implement a SPL multisig variant (for consistency), we need to add a lot of code (to ensure safety).

Is all this extra code worth it when they can just call `spl.setAuthority` directly?

nvsriram: I pondered on it for a while and I do like being able to offer a safe way to accept the authority. While it is already there on the SDK side (via `ntt.pdas`), I think having it on the program side as well makes it more complete so I've added `AcceptTokenAuthority` ix to this PR.

re: SPL multisig support, I wrote a custom wrapper for it so the eventual `AcceptTokenAuthorityMultisig` ix shouldn't be too difficult to add.

The trade-off between adding all this code and users being able to use `spl.setAuthority` directly instead is still a concern but I don't really see an easy way to fix it than hoping they read the docs.

solana/programs/example-native-token-transfers/src/instructions/admin.rs
```diff
@@ -154,6 +156,200 @@ pub fn claim_ownership(ctx: Context<ClaimOwnership>) -> Result<()> {
...
+        bump = pending_token_authority.bump,
+        has_one = rent_payer @ NTTError::IncorrectRentPayer,
+        close = rent_payer
+     )]
+    pub pending_token_authority: Account<'info, PendingTokenAuthority>,
+
+    pub token_program: Interface<'info, token_interface::TokenInterface>,
+
+    pub system_program: Program<'info, System>,
+}
```
nonergodic, line 319: This is currently permissionless (no particular signer required), which allows for grieving by closing the `pending_token_authority` account before the handover can be completed. So this should also require the owner's signature.

solana/ts/lib/ntt.ts
```diff
@@ -710,6 +713,55 @@ export namespace NTT {
...
+  export async function createAcceptTokenAuthorityInstruction(
+    program: Program<NttBindings.NativeTokenTransfer<IdlVersion>>,
+    config: NttBindings.Config<IdlVersion>,
+    args: {
+      currentAuthority: PublicKey;
+    },
+    pdas?: Pdas
+  ) {
+    pdas = pdas ?? NTT.pdas(program.programId);
+    return await program.methods
```
nonergodic, line 725:
```suggestion
    return program.methods
```

solana/ts/lib/ntt.ts
```diff
@@ -710,6 +713,55 @@ export namespace NTT {
...
+    config: NttBindings.Config<IdlVersion>,
+    args: {
+      rentPayer: PublicKey;
+      owner: PublicKey;
+      newAuthority: PublicKey;
+    },
+    pdas?: Pdas
+  ) {
+    pdas = pdas ?? NTT.pdas(program.programId);
+    return await program.methods
```
nonergodic, line 748:
```suggestion
    return program.methods
```

nonergodic (changes requested):

nonergodic (approved):

kcsongor (approved):

PR #572, opened 2024-12-04
tilt: speedier ci

evan-gray:
This PR makes several speed improvements to the tilt tests.

Initially, the test suite took about 530 seconds. That's nearly 9 minutes!

1. The tests had a 1 minute sleep on relays, but after #538 the relay uses the same consistency level (200 = instant) so that's no longer necessary. New time 400s, about 130s savings.
2. Switch the Solana consistency level to Confirmed when one of the tilt feature flags is enabled. New time 320s, about 80s savings.
3. Batch the accountant VAAs - they need to be in a particular order, but there's no reason not to submit them all in the same transaction as long as it's not above the gas limit. New time 280s, about 40s savings.
4. Parallelize the peer setup - each chain's transactions are run sequentially with a single wallet, but those can be done in parallel. New time 220s, about 60s savings.
5. Parallelize tests - I initially split it into two files, but it seemed a little easier to split into one for now. The code changes accommodate either approach. The run time was the about the same either way, except I got an open handle with the separate files. New time 120s, about 100s savings.

Now the tests take just about 2 minutes, or less than a quarter of the original time!

This also avoids building all of the monorepo Solana contracts and simply starts with the end result of having a core bridge with its accounts, the rest of the programs and accounts weren't used anyway. In a future PR, I would like to similarly improve EVM and Wormchain.


nvsriram (approved):

johnsaigle (approved):
Nice improvements, I think the only thing to consider is that we'll need to update `bridge.so` here if there's ever a bridge update, otherwise we'll be testing against an old version.

evan-gray:
> Nice improvements, I think the only thing to consider is that we'll need to update `bridge.so` here if there's ever a bridge update, otherwise we'll be testing against an old version.

That's correct, though any breakage of the core bridge would break all integrators, so I figured that was a safe-enough assumption to make. I plan to have a similar approach to speed up `eth-devnet` and may re-organize and document this there.

PR #573, opened 2024-12-05
ci: Debug `CLI / test-solana` job

nvsriram:
`CLI / test-solana` job has been failing sporadically (race condition probably) in recent PRs so this PR attempts to fix it

### DIFF
--- a/solana/ts/sdk/ntt.ts
+++ b/solana/ts/sdk/ntt.ts
@@ -619,7 +619,7 @@ export class SolanaNtt<N extends Network, C extends SolanaChains>
       );
     } catch (e) {
       // This might happen if e.g. the program is not deployed yet.
-      const version = "3.0.0";
+      const version = "2.0.0";
       return version;
     }
   }


cli/src/index.ts
```diff
@@ -472,6 +472,9 @@ yargs(hideBin(process.argv))
                 argv["binary"]
             );
 
+            // we sleep here to ensure the program has enough time to be deployed
+            // and we can fetch the version from the program itself
+            await new Promise((resolve) => setTimeout(resolve, 2000));
```
evan-gray, line 477: Is there a less arbitrary way to detect this? Can it be known that the program will always take less than 2 seconds to be deployed?

nvsriram: There is no guarantee that the program will always take less than 2 seconds to be deployed or in this case for the RPC to catch up to the deployed program (I could probably update the comment). 
This sleep happens _after_ [deploy proc has exited](https://github.com/wormhole-foundation/native-token-transfers/blob/main/cli/src/index.ts#L1403). So this sleep just acts as a buffer for the RPC to query the correct deployed program - similar to the [sleep before initialize](https://github.com/wormhole-foundation/native-token-transfers/blob/main/cli/src/index.ts#L1417).

The more arbitrary option I considered was a spinloop inside `nttFromManager` if an optional `expectedVersion` parameter is passed in:
```ts
async function nttFromManager<N extends Network, C extends Chain>(
    ch: ChainContext<N, C>,
    nativeManagerAddress: string,
    expectedVersion?: string
): Promise<{ ntt: Ntt<N, C>; addresses: Partial<Ntt.Contracts> }> {
    let onlyManager: Ntt<N, C>;
    onlyManager = await ch.getProtocol("Ntt", {
        ...
    });
    while (expectedVersion && getVersion(ch.chain, onlyManager) !== expectedVersion) {
       await new Promise((resolve) => setTimeout(resolve, 2000));
       onlyManager = await ch.getProtocol("Ntt", {
         ...
       });
    }
    ...
}
```
And it could be called in `upgrade`:
```ts
const { ntt: upgraded } = await nttFromManager(ch, chainConfig.manager, toVersion ?? undefined);
```

**Concerns with this approach**:

1. This CI job fail is only due to the edge case where the CLI release has not caught up to the latest NTT manager version so the complexity added with a spinloop and debugging might be overkill?
2. Calling `getProtocol` in a loop might not be the most optimal way to fetch versions
3. Not sure if the strict equality is the best way to compare versions (in case of semver etc.)

evan-gray: I think my proposal in #574 fixes the underlying problem without the arbitrary wait.

PR #574, opened 2024-12-12
cli: use finalized commitment level for solana program deploy

evan-gray:
testing an alternate approach to #573. I think the test is simply failing because the deploy is using `confirmed` and the read is using `finalized` (or something like that). My hunch is that setting the deploy to `finalized` should incorporate the appropriate wait time.


nvsriram (approved):
LGTM!

kev1n-peters (approved):

PR #587, opened 2025-02-03
solana: Handle transferring mint authority using SPL Multisig

nvsriram:
### Context:

Currently, there are instructions to safely handle transferring token mint authority to-and-from the NTT Manager - `token_authority` PDA.

However, with SPL Multisig support, these instructions need to be modified to handle all cases.

### List of Abbreviations:

* `Multisig(...)` = any generic SPL Multisig with signers …
* `Multisig(TA, ...)` = SPL Multisig with m = 1 and `token_authority` PDA as one of its required signers
* `TA` = `token_authority` PDA
* `TA`/`Multisig(TA, ...)` = Representation of NTT Manager
   * Could be `token_authority` PDA or SPL Multisig with m = 1 and `token_authority` PDA as one of its required signers

#### Transfer Mint Authority OUT:

  * **`set_token_authority_one_step_unchecked`**: `TA`/`Multisig(TA, ...)` -> any non-signer
  * **`set_token_authority`**: `TA`/`Multisig(TA, ...)` -> non-signer `S`
    * **`claim_token_authority`**: claimed by `S` as signer
    * **`claim_token_authority_to_multisig`**: claimed by `S` = `Multisig(...)`
       *  `ctx.remaining_accounts` act as required signers of `Multisig(...)`
    * **`revert_token_authority`**: reverted back to `TA`/`Multisig(TA, ...)`

#### Transfer Mint Authority IN:

  * **`accept_token_authority`**: signer `S` -> `TA`/`Multisig(TA, ...)`
  * **`accept_token_authority_from_multisig`**: `Multisig(...)` -> `TA`/`Multisig(TA, ...)`
     *  `ctx.remaining_accounts` act as required signers of `Multisig(...)`

### Implementation Details:

* When the NTT Manager is `Multisig(TA, ...)`: `Option<InterfaceAccount<'info, SplMultisig>>` is used to avoid additional explicit instructions
* When `Multisig(...)` needs to sign/validate: `ctx.remaining_accounts` is treated as required signers

### tl;dr:

This PR adds:
* 7 mint authority update instructions to handle all SPL Multisig cases (described above)
* Add NTT namespace TS helpers for the instructions
* Update TS test to cover multiple mint authority transfer instances
* `admin.rs` and TS test case refactor
* Update IDL


solana/programs/example-native-token-transfers/src/instructions/admin.rs
```diff
@@ -343,21 +448,113 @@ pub struct ClaimTokenAuthority<'info> {
...
+        token_interface::spl_token_2022::processor::Processor::validate_owner(
+            &ctx.accounts.common.token_program.key(),
+            &multisig.key(),
+            &multisig,
+            multisig.data_len(),
+            ctx.remaining_accounts,
+        )?;
+    }
+
+    if let Some(multisig_token_authority) = &ctx.accounts.common.multisig_token_authority {
```
nonergodic, line 519: Consider `match`.

solana/tests/anchor.test.ts
nonergodic, line 1: The code in here could really profit from a bunch of DRY-ness. See e.g. [TBRv3 tests](https://github.com/XLabs/arbitrary-token-transfers/blob/8c97cc32cf71ead44de3861635baa4b971bc4d86/solana/tests/token-bridge-relayer-tests.ts#L386-L388) that use [various helpers](https://github.com/XLabs/arbitrary-token-transfers/blob/8c97cc32cf71ead44de3861635baa4b971bc4d86/solana/tests/utils/helpers.ts#L49-L75) under the hood.

solana/ts/lib/ntt.ts
```diff
@@ -849,29 +849,94 @@ export namespace NTT {
...
       })
       .instruction();
   }
 
+  export async function createAcceptTokenAuthorityFromMultisigInstruction(
+    program: Program<NttBindings.NativeTokenTransfer<IdlVersion>>,
+    config: NttBindings.Config<IdlVersion>,
+    args: {
+      currentMultisigAuthority: PublicKey;
+      additionalSigners: PublicKey[];
```
nonergodic, line 877:
```suggestion
      additionalSigners: readonly PublicKey[];
```

solana/ts/lib/ntt.ts
```diff
@@ -914,12 +981,79 @@ export namespace NTT {
...
+      .instruction();
+  }
+
+  export async function createClaimTokenAuthorityToMultisigInstruction(
+    program: Program<NttBindings.NativeTokenTransfer<IdlVersion>>,
+    config: NttBindings.Config<IdlVersion>,
+    args: {
+      rentPayer: PublicKey;
+      newMultisigAuthority: PublicKey;
+      additionalSigners: PublicKey[];
```
nonergodic, line 1026:
```suggestion
      additionalSigners: readonly PublicKey[];
```

nonergodic (dismissed):
Only nits.

One other thing that I noticed that isn't part of the changes here:
The address constraint on the owner signer in `RevertTokenAuthority` has no custom error, while the other address checks do. I assume this is because usually these type of constraints are checked via `has_one` which isn't possible here - but I figured I'd flag it here regardless (might be worth a short comment).

nonergodic (approved):
Code is very neat now!
(at least in as far as Anchor code can be neat :grimacing:)

Regarding the refactor:
I think I'd probably have the ownership ixs and the token authority ixs in two separate files, but that's about it.

What I started doing in the Solidity SDK where I'm facing a similar problem of "lots of code, but not actually a lot of complexity" is just put [an overview comment at the top of the file](https://github.com/wormhole-foundation/wormhole-solidity-sdk/blob/5159f6e82696ed86026d59c526d5da85f2a962d3/src/libraries/VaaLib.sol#L94-L135) that says "look, all this garbage below is really just accomplishes the following in roughly this manner".

As such, it would imho be justifiable to just keep all the code in admin.rs because the file is so entirely uninspiring and unsurprising.

I think you can really just choose here what you like best.

kcsongor (approved):

nvsriram:
> Regarding the refactor: I think I'd probably have the ownership ixs and the token authority ixs in two separate files, but that's about it.

Yeah, I refactored it the same way - I think lengthy related instructions in individual files makes sense.

solana/tests/anchor.test.ts
nvsriram: Done, just added a `helpers` file with a similar setup and refactored the main test.

nonergodic (approved):
good test cleanup!

kcsongor (approved):

PR #589, opened 2025-02-07
solana: Add `set_threshold` ix

nvsriram:
Currently there is no way to update the NTT manager `threshold` (# of transceivers that must attest to a transfer before it is accepted) - it defaults to 1 - or deregister a transceiver. With multi-transceiver support, having a way to manage transceivers and their threshold is essential.

This PR adds:
* `set_threshold` ix to update the threshold (`0 < threshold <= # of enabled transceivers`)
* `deregister_transceiver` ix to deregister (disable transceiver in enabled transceiver bitmap):
   * Also decrements threshold (down to minimum 1) if # of enabled transceivers < threshold 
* Update `register_transceiver` ix to also act as re-enable:
   * If `RegisteredTransceiver` account is init for the first time, assign new id
   * Otherwise, reuse old id
* Rust tests to verify edge cases
* `createSetThresholdInstruction` and `createDeregisterTransceiverIx` TS helpers as a part of NTT namespace and `SolanaNtt` class respectively
* Update IDL

### DIFF
--- a/solana/programs/example-native-token-transfers/src/bitmap.rs
+++ b/solana/programs/example-native-token-transfers/src/bitmap.rs
@@ -48,8 +48,11 @@ impl Bitmap {
             .expect("Bitmap length must not exceed the bounds of u8")
     }
 
-    pub fn len(self) -> usize {
-        BM::<128>::from_value(self.map).len()
+    pub fn len(self) -> u8 {
+        BM::<128>::from_value(self.map)
+            .len()
+            .try_into()
+            .expect("Bitmap length must not exceed the bounds of u8")
     }
 
     pub fn is_empty(self) -> bool {
--- a/solana/programs/example-native-token-transfers/src/error.rs
+++ b/solana/programs/example-native-token-transfers/src/error.rs
@@ -63,6 +63,8 @@ pub enum NTTError {
     InvalidMultisig,
     #[msg("ThresholdTooHigh")]
     ThresholdTooHigh,
+    #[msg("InvalidTransceiverProgram")]
+    InvalidTransceiverProgram,
 }
 
 impl From<ScalingError> for NTTError {
--- a/solana/programs/example-native-token-transfers/src/error.rs
+++ b/solana/programs/example-native-token-transfers/src/error.rs
@@ -61,6 +61,8 @@ pub enum NTTError {
     IncorrectRentPayer,
     #[msg("InvalidMultisig")]
     InvalidMultisig,
+    #[msg("ThresholdTooHigh")]
+    ThresholdTooHigh,
 }
 
 impl From<ScalingError> for NTTError {
--- a/solana/programs/example-native-token-transfers/src/instructions/admin/mod.rs
+++ b/solana/programs/example-native-token-transfers/src/instructions/admin/mod.rs
@@ -124,6 +124,45 @@ pub fn register_transceiver(ctx: Context<RegisterTransceiver>) -> Result<()> {
     Ok(())
 }
 
+#[derive(Accounts)]
+pub struct DeregisterTransceiver<'info> {
+    #[account(
+        mut,
+        has_one = owner,
+    )]
+    pub config: Account<'info, Config>,
+
+    #[account(mut)]
+    pub owner: Signer<'info>,
+
+    #[account(executable)]
+    /// CHECK: transceiver is meant to be a transceiver program. Arguably a `Program` constraint could be
+    /// used here that wraps the Transceiver account type.
+    pub transceiver: UncheckedAccount<'info>,
+
+    #[account(
+        seeds = [RegisteredTransceiver::SEED_PREFIX, transceiver.key().as_ref()],
+        bump,
+        constraint = config.enabled_transceivers.get(registered_transceiver.id)? @ NTTError::DisabledTransceiver,
+    )]
+    pub registered_transceiver: Account<'info, RegisteredTransceiver>,
+}
+
+pub fn deregister_transceiver(ctx: Context<DeregisterTransceiver>) -> Result<()> {
+    ctx.accounts
+        .config
+        .enabled_transceivers
+        .set(ctx.accounts.registered_transceiver.id, false)?;
+
+    // decrement threshold if too high
+    let num_enabled_transceivers = ctx.accounts.config.enabled_transceivers.len();
+    if num_enabled_transceivers < ctx.accounts.config.threshold {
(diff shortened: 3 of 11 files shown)


solana/programs/example-native-token-transfers/src/instructions/admin.rs
```diff
@@ -469,6 +469,49 @@ pub fn register_transceiver(ctx: Context<RegisterTransceiver>) -> Result<()> {
...
+
+pub fn deregister_transceiver(ctx: Context<DeregisterTransceiver>) -> Result<()> {
+    ctx.accounts
+        .config
+        .enabled_transceivers
+        .set(ctx.accounts.registered_transceiver.id, false)?;
+
+    // decrement threshold if too high
+    let num_enabled_transceivers = u8::try_from(ctx.accounts.config.enabled_transceivers.len())
+        .expect("Bitmap length must not exceed the bounds of u8");
```
nonergodic, line 507: This makes no sense to me, neither here nor in the Bitmap class itself.

Bitmap uses a u128 underneath, so it is guaranteed that the count of 1s can be at most 128 < 255. But even [the implementation of Bitmap doesn't simply unwrap but has a pointless error check](https://github.com/wormhole-foundation/native-token-transfers/blob/3311787ab22087f5c10ab08edb6a2a5e3f7afd77/solana/programs/example-native-token-transfers/src/bitmap.rs#L46-L48).

I.e. it's not that it "must not exceed the bounds of u8" but rather that it can't.

And bitmap itself should also reflect that by returning a `u8` from `len`. In lieu of that, the try_from here should really just be a cast, imho.

solana/programs/example-native-token-transfers/src/instructions/admin.rs
```diff
@@ -469,6 +469,49 @@ pub fn register_transceiver(ctx: Context<RegisterTransceiver>) -> Result<()> {
...
+    if num_enabled_transceivers < ctx.accounts.config.threshold {
+        // threshold should be at least 1
+        ctx.accounts.config.threshold = num_enabled_transceivers.max(1);
```
nonergodic, line 510: I find it inconsistent that registering a transceiver does not increase the threshold, but deregistering decreases it.

solana/programs/example-native-token-transfers/src/instructions/admin.rs
```diff
@@ -544,3 +587,25 @@ pub fn set_paused(ctx: Context<SetPaused>, paused: bool) -> Result<()> {
...
+    #[account(
+        mut,
+        has_one = owner,
+        constraint = threshold <= u8::try_from(config.enabled_transceivers.len()).unwrap() @ NTTError::ThresholdTooHigh
+    )]
+    pub config: Account<'info, Config>,
+}
+
+pub fn set_threshold(ctx: Context<SetThreshold>, threshold: u8) -> Result<()> {
+    if threshold == 0 {
```
nonergodic, line 606: Right now you could set the threshold to any value above 128, making it literally impossible to attest anything. If this is considered a feature, it should be documented, if it is not, it should be prohibited.

nonergodic:
I also now realize why I felt confused:
Enabled transceivers suggests that there can be such a thing as a disabled transceiver. But really, there are only registered transceivers, so the naming is suboptimal.

The deregister function makes it seem like you just want to recoup rent, but really it's primarily about disabling the transceiver.

kcsongor (commented):
overall the changes look good, but a couple of comments:
- on the EVM side, we don't support deregistering transceivers, only disabling them. We should consider doing the same here. the rent saving is marginal here (given the frequency of this operation), and it might be good to be able to re-enable a transceiver (re-registering it will give it a new id).
- I think we shouldn't modify the threshold when disabling a transceiver. arguably the contract should be paused when modifying transceiver/threshold configurations, but if it's not, then as the admin modifies the config, there will be temporary scenarios where less than the intended threshold of transceivers can attest a message

solana/programs/example-native-token-transfers/src/instructions/admin.rs
nvsriram: It wouldn't be possible to set it to above 128 because of the `ThresholdTooHigh` [constraint](https://github.com/wormhole-foundation/native-token-transfers/pull/589/files#diff-1a508b38581cdff755d859a43ca778a99cff954250e4f364212a3f16ad4ff9ecR600):

``` rust
#[account(
      mut,
      has_one = owner,
      constraint = threshold <= u8::try_from(config.enabled_transceivers.len()).unwrap() @ NTTError::ThresholdTooHigh
  )]
  pub config: Account<'info, Config>,
```

nvsriram: Threshold is limited to the number of enabled transceivers.

nvsriram:
> * on the EVM side, we don't support deregistering transceivers, only disabling them. We should consider doing the same here. the rent saving is marginal here (given the frequency of this operation), and it might be good to be able to re-enable a transceiver (re-registering it will give it a new id).

Are you suggesting a `reenable_transceiver` instruction or updating `register_transceiver` to double as reenable?

Currently, `register` is defined with `init`. To make it so we can re-enable, it would be `init_if_needed`. Similarly, I'd have to handle the cases depending on whether it is initialized or not to avoid updating its id.

solana/programs/example-native-token-transfers/src/instructions/admin.rs
nvsriram: There may be cases where we wouldn't need threshold to be maxed out (# of enabled transceivers).
For example, if they want a `1-of-n` configuration where as long as they receive `1` attestation from any of the `n` transceivers, they are good to redeem.

However, in deregister there is no benefit to having `n + 1` threshold with `n` transceivers - this will never work. In fact, this could cause issues where `transfer` is called with insufficient transceivers set up.

nvsriram:
> * I think we shouldn't modify the threshold when disabling a transceiver. arguably the contract should be paused when modifying transceiver/threshold configurations, but if it's not, then as the admin modifies the config, there will be temporary scenarios where less than the intended threshold of transceivers can attest a message

This is what EVM does (decrement threshold if too high) so it is mostly just lifted from there.

I agree that disabling could be limited to when contract is paused so could add a constraint for that. 
However, say the owner pauses, deregisters a transceiver but threshold is not updated. Then they unpause and transfer, it will still go through as there is no `ThresholdTooHigh` check...

nonergodic (approved):

PR #591, opened 2025-02-21
ci: Fix `call didn't revert at a lower depth than cheatcode call depth` forge error in EVM tests

nvsriram:
With Foundry v1 release, some forge tests have been failing with this error. 
This PR fixes this by adding `/// forge-config: default.allow_internal_expect_revert = true` above the test function.

See https://book.getfoundry.sh/cheatcodes/expect-revert#description for reference:
> Selectively by using an inline configuration entry where it is DEEMED SAFE:
> Add `/// forge-config: default.allow_internal_expect_revert = true` above the test function.

### DIFF
--- a/evm/test/TrimmedAmount.t.sol
+++ b/evm/test/TrimmedAmount.t.sol
@@ -67,6 +67,7 @@ contract TrimmingTest is Test {
         }
     }
 
+    /// forge-config: default.allow_internal_expect_revert = true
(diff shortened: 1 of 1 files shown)


nonergodic (approved):

kcsongor (approved):
nice

djb15 (approved):

PR #593, opened 2025-03-05
solana: Remove `*_multisig` instruction variants

nvsriram:
This PR adds:
* Remove `initialize_multisig` and `release_inbound_mint_multisig` ix
* Add `Option<'info, SplMultisig<'info>>` account to `Initialize` and `ReleaseInboundMint` struct
* Update cargo and TS tests
* Derive multisig token authority from the token mint authority
* Update IDL

### DIFF
--- a/solana/programs/example-native-token-transfers/src/instructions/release_inbound.rs
+++ b/solana/programs/example-native-token-transfers/src/instructions/release_inbound.rs
@@ -54,7 +54,7 @@ pub struct ReleaseInbound<'info> {
 
 #[derive(AnchorDeserialize, AnchorSerialize)]
 pub struct ReleaseInboundArgs {
-    pub revert_on_delay: bool,
+    pub revert_when_not_ready: bool,
 }
 
 // Burn/mint
@@ -75,8 +75,8 @@ pub struct ReleaseInboundMint<'info> {
 }
 
 /// Release an inbound transfer and mint the tokens to the recipient.
-/// When `revert_on_delay` is true, the transaction will revert if the
-/// release timestamp has not been reached. When `revert_on_delay` is false, the
+/// When `revert_when_not_ready` is true, the transaction will revert if the
+/// release timestamp has not been reached. When `revert_when_not_ready` is false, the
 /// transaction succeeds, but the minting is not performed.
 /// Setting this flag to `false` is useful when bundling this instruction
 /// together with [`crate::instructions::redeem`] in a transaction, so that the minting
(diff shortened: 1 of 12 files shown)


nvsriram:
As this PR introduces breaking changes to `initialize` and `release_inbound_mint`, this PR also covers breaking changes to `release_inbound_*` - introduced by https://github.com/wormhole-foundation/native-token-transfers/pull/336.

solana/programs/example-native-token-transfers/src/instructions/release_inbound.rs
```diff
@@ -258,14 +245,21 @@ pub fn release_inbound_unlock<'info>(
...
-    } else if revert_on_delay {
-        Err(NTTError::CantReleaseYet.into())
+    } else if revert_when_not_ready {
+        match inbox_item.release_status {
+            ReleaseStatus::NotApproved => Err(NTTError::TransferNotApproved.into()),
+            ReleaseStatus::ReleaseAfter(_) => Err(NTTError::CantReleaseYet.into()),
+            // Unreachable: if released, [`InboxItem::try_release`] will return an Error immediately
+            // rather than Ok(bool).
+            ReleaseStatus::Released => Err(NTTError::TransferAlreadyRedeemed.into()),
+        }
```
nonergodic, line 262: This would probably be cleaner with a `match` up top with 3 arms for `Ok(true)`, `Ok(false)`, and `Err`  because it should allow avoiding the unreachable part.

nonergodic (approved):

nonergodic:
Btw, I like the streamlining of integrating the multisig handling in their respective instructions instead of doing the ix multiplexing :+1:

solana/programs/example-native-token-transfers/src/instructions/release_inbound.rs
nvsriram: That still wouldn't be too clean as the `Ok(false)` branch could be due to `ReleaseStatus::NotApproved` or `ReleaseStatus::ReleaseAfter` when `revert_when_not_ready` is false. So that'd result in a nested match.

nonergodic: Ah, I see what you're saying, you'd still have to look at the `release_status`. Hmm...

nonergodic: Yeah, then maybe this is the best you can do.

nik-suri (approved):
Stamping this PR to unblock merge. 2nd review will come via security review by Asymmetric Research.

PR #598, opened 2025-03-15
move cli logic -> cli-core so it can be used as a node package

elee1766:


kcsongor:
When running `./cli/install.sh` locally on this branch, I get

```
bun link v1.1.20 (ae194892)
error: GET https://registry.npmjs.org/@wormhole-foundation%2fntt-cli-core - 404
error: @wormhole-foundation/ntt-cli-core@* failed to resolve
```

elee1766:
> When running `./cli/install.sh` locally on this branch, I get
> 
> ```
> bun link v1.1.20 (ae194892)
> error: GET https://registry.npmjs.org/@wormhole-foundation%2fntt-cli-core - 404
> error: @wormhole-foundation/ntt-cli-core@* failed to resolve
> ```

correct. install cannot work because the bun install function can only fetch from npm, which i cannot publish to.

kcsongor:
hm, isn't there a way to set the workspace up in the top level directory so that it gets installed locally?

elee1766:
i don't know because i dont think the top level repo is bun. bun seems to only be used for the installed cli?

PR #607, opened 2025-04-10
solana: Fix rate limit reset on peer upgrade

nvsriram:
Currently when peer is upgraded, the rate limit is reset.
This PR sets rate limit to new rate limit if the rate limit is uninitialized/unused. Otherwise it updates the rate limit.


mdulin2 (approved):

kcsongor (approved):
nice!

PR #608, opened 2025-04-11
solana: Make `initialize_lut` instruction permissioned by `owner`

nvsriram:
Since `initialize_lut` instruction is permissionless, an attacker could theoretically backrun the lookup table with garbage values for `sequence` and `fee_collector` accounts as they are unconstrained in this call and are only verified by the core bridge. This could cause transactions that depend on the LUT being initialized correctly to fail due to size constraints. This PR makes `initialize_lut` permissioned using the `owner`.

### DIFF
--- a/solana/programs/example-native-token-transfers/src/instructions/admin/mod.rs
+++ b/solana/programs/example-native-token-transfers/src/instructions/admin/mod.rs
@@ -149,7 +149,6 @@ pub struct DeregisterTransceiver<'info> {
     )]
     pub config: Account<'info, Config>,
 
-    #[account(mut)]
     pub owner: Signer<'info>,
 
     #[account(executable)]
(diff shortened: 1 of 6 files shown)


mdulin2 (approved):

kcsongor:
Could you describe in more detail what the backrunning exploit looks like? Each call to `initialize_lut` creates a fresh lut, and it's only written to by the program.

nvsriram:
> Could you describe in more detail what the backrunning exploit looks like? Each call to `initialize_lut` creates a fresh lut, and it's only written to by the program.

Currently, the client calls `NTT.getAddressLookupTable()` to read the `address` field in the `lut` PDA.
With the `initialize_lut` instruction is permissionless, an attacker could call the instruction and potentially set the `address` to a new LUT with incorrect/outdated `Entries` values.
Thus, the client would be using a misconfigured LUT. The consequence of this could be DoS issues for bundled transactions whose sizes are kept below limit due to LUT being configured correctly.

mdulin2 (approved):

kcsongor (approved):

PR #612, opened 2025-05-01
[wip] NttOwner support

elee1766:
the NTT dashboard uses an NttOwner contract to own the ntt deployment, in order to faciliate the calling of setPeers in one transaction instead of multiple. 

this is good, however, the cli does not support this. 

this pr adds two things:

1. an additional flag to allow the skipping of chains in push. this is important if the owner of one side of the deployment does not have control over another side, but would like to update their limites

2. by checking using eip165, we can determine if the owner implements INttOwner. if it does, then we provide an alternative route to signSendWait using type assertions. 

this is still wip though, as the same override must be set for pushDeployment, and it is not yet set.

really what should happen is i should create a version of signSendWait for each chain that is aware of the ownership status, and can override the same way i am here. this way, we can reuse that same logic for pushDeployment.

### DIFF
--- a/cli/.gitignore
+++ b/cli/.gitignore
@@ -1,3 +1,7 @@
+
+deployment.json
+id.json
+
 # Based on https://raw.githubusercontent.com/github/gitignore/main/Node.gitignore
 
 # Logs
@@ -173,3 +177,4 @@ dist
 
 # Finder (MacOS) folder config
 .DS_Store
+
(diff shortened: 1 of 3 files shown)


kcsongor (dismissed):
very nice!

elee1766:
that evm cli test seems to be failing on master with the same error

dvgui (approved):

PR #617, opened 2025-05-15
solana: Skip upgrade authority CPI if current owner is `upgrade_lock`

nvsriram:
This PR adds a check to ensure current owner is not `upgrade_lock` and only then performs the CPI call to `set_upgrade_authority_checked`.


djb15 (dismissed):

solana/programs/example-native-token-transfers/src/instructions/admin/transfer_ownership.rs
```diff
@@ -55,21 +55,26 @@ pub struct TransferOwnership<'info> {
...
+                    current_authority: ctx.accounts.owner.to_account_info(),
+                    new_authority: ctx.accounts.upgrade_lock.to_account_info(),
+                },
+                &[&[b"upgrade_lock", &[ctx.bumps.upgrade_lock]]],
+            ),
+            &crate::ID,
+        )
+    } else {
+        Ok(())
+    }
```
djb15, line 77: Nit: You don't really need this else block, you could just return Ok() at the end of the instruction

djb15 (approved):

nik-suri (approved):

PR #619, opened 2025-05-19
ci:  Use `release.anza.xyz` instead of `release.solana.com` to install Solana CLI

nvsriram:
Solana CI jobs using `release.solana.com` to install Solana CLI seem to fail with `solana-keygen: command not found` error as `https://release.solana.com` is not accessible. This PR updates CI to use `https://release.anza.xyz` instead

### DIFF
--- a/.github/workflows/solana.yml
+++ b/.github/workflows/solana.yml
@@ -68,7 +68,7 @@ jobs:
         env:
           SOLANA_VERSION: ${{ steps.solana.outputs.version }}
         run: |
-          sh -c "$(curl -sSfL https://release.solana.com/v${SOLANA_VERSION}/install)"
+          sh -c "$(curl -sSfL https://release.anza.xyz/v${SOLANA_VERSION}/install)"
           ~/.local/share/solana/install/active_release/bin/sdk/sbf/scripts/install.sh
 
       - name: cargo build-sbf && cargo test-sbf && cargo test


kcsongor (approved):
nice!

.github/actions/setup-anchor/action.yml
```diff
@@ -0,0 +1,39 @@
...
+        path: |
+          ~/.cache/solana/
+          ~/.local/share/solana/
+        key: solana-cli-${{ runner.os }}-build-${{ inputs.solana-cli-version }}
+    - name: Install Solana CLI tools
+      run: |
+        sh -c "$(curl -sSfL https://release.anza.xyz/v${{ inputs.solana-cli-version }}/install)"
+      shell: bash
+    - name: Update PATH
+      run: echo "/home/runner/.local/share/solana/install/active_release/bin" >> $GITHUB_PATH
```
evgeniko, line 35: is the hardcoded path here fine?
Maybe we use the output of the installer script to have a dynamic location?

evgeniko (approved):

.github/actions/setup-anchor/action.yml
nvsriram: Yeah, it should be good for now. I think even the Solana Cargo SBF CI job also uses the same hardcoded path.
We could parse the output of the installer script to fetch the command. A potential issue with that is if the output of the installer script changes and we end up running the wrong command.

PR #626, opened 2025-05-23
Feature/node compatible cli

elee1766:
so this should supposedly be able to be used as both a cli utility and a library

i am testing with `npm build && npm pack` and `npm i -g <output>`


elee1766:
for this to work, I think that the packages all need to actually be bumped and republished? there were some incompatibilities

PR #628, opened 2025-05-30
cli: Add commands to create valid SPL Multisig and set mint authority

nvsriram:
This PR adds:
* Add `create_spl_multisig <multisigMemberPubkey...>` command that mimics `spl-token create-multisig` to create valid SPL Multisig.
* Add `set-mint-authority <newAuthority>` command that sets mint authority based on version:
  * If version < `3.x.x`: calls `spl.setAuthority` directly
  * If version >= `3.x.x`: calls `acceptTokenAuthority`
  * Also handles cases where `newAuthority` is a valid SPL Multisig via the `--multisig` flag
* Handle TODO to factor out function to get chain context
* Update GitHub links and README docs
* Format file, rename imports, reorder imports

### DIFF
--- a/cli/src/index.ts
+++ b/cli/src/index.ts
@@ -1124,11 +1124,100 @@ yargs(hideBin(process.argv))
             tokenProgram
           );
           console.log(ata.toBase58());
-                    })
-                .demandCommand()
         }
-    )
-    .help()
+      )
+      .command(
+        "create-spl-multisig <multisigMemberPubkey...>",
+        "create a valid SPL Multisig (see docs for more info)",
+        (yargs) =>
+          yargs
+            .positional("multisigMemberPubkey", {
+              describe:
+                "public keys of the members that can independently mint",
+              type: "string",
+              demandOption: true,
+            })
+            .option("path", options.deploymentPath)
+            .option("payer", { ...options.payer, demandOption: true })
+            .example(
+              "$0 solana create-spl-multisig Sol1234... --payer <SOLANA_KEYPAIR_PATH>",
+              "Create multisig with Sol1234... having independent mint privilege alongside NTT token-authority"
+            )
+            .example(
+              "$0 solana create-spl-multisig Sol1234... Sol3456... Sol5678... --payer <SOLANA_KEYPAIR_PATH>",
+              "Create multisig with Sol1234..., Sol3456..., and Sol5678... having mint privileges alongside NTT token-authority"
+            ),
+        async (argv) => {
+          const path = argv["path"];
+          const deployments: Config = loadConfig(path);
+          const chain: Chain = "Solana";
+          const network = deployments.network as Network;
+          
+          if (!fs.existsSync(argv["payer"])) {
+            console.error("Payer not found. Specify with --payer");
+            process.exit(1);
+          }
+          const payerKeypair = Keypair.fromSecretKey(
+            new Uint8Array(
+              JSON.parse(fs.readFileSync(argv["payer"]).toString())
+            )
+          );
+          
+          if (!(chain in deployments.chains)) {
+            console.error(`Chain ${chain} not found in ${path}`);
+            process.exit(1);
+          }
+          const chainConfig = deployments.chains[chain]!;
+          const ch = getChainContext(network, chain, overrides);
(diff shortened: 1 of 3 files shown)


nvsriram:
For local testing, I used a demo Solana devnet deployment: [8y2hh2wGnagy8wc8Fe8xsjUtLSHnzFbiPmFNrext4jHm](https://explorer.solana.com/address/8y2hh2wGnagy8wc8Fe8xsjUtLSHnzFbiPmFNrext4jHm?cluster=devnet)
I tested the commands first using version `2.0.0` to ensure the correct errors were thrown. Then after upgrading to local `3.0.0` version, tested the commands again.

nvsriram:
> Could you separate the work for the TODO refactor from the new functionality to make this easier to review?

The TODO refactor refers to this commit: https://github.com/wormhole-foundation/native-token-transfers/pull/628/commits/a385a5b41c28eccd9af6fae60bd860a552c0d6a5. As it is very small in scope. I added it with this PR. However, if you think it makes it easier if it were moved out, I can undo that commit.

On a similar note, I've split the commits based on features so it might be easier to review the commits one-by-one.

dvgui:
@nvsriram similar to what @tonyjin asked. I would prefer to have a PR with specific additions for this feature, and we can open a new one for code indentation and maybe some housekeeping, like reorganising functions to more files for better readability and structure.

dvgui (dismissed):

nvsriram:
> @nvsriram similar to what @tonyjin asked. I would prefer to have a PR with specific additions for this feature, and we can open a new one for code indentation and maybe some housekeeping, like reorganising functions to more files for better readability and structure.

Sounds good, will do that then. 
For this PR, I'll revert all commits except https://github.com/wormhole-foundation/native-token-transfers/pull/628/commits/c0f0a91dcf8003f749c2c9bcdc0c3c81068d8fc3 (needed to solve import error), https://github.com/wormhole-foundation/native-token-transfers/pull/628/commits/25d413c5741ef228636e1af238b14a27e66a14bd (`create-spl-multisig`), and https://github.com/wormhole-foundation/native-token-transfers/pull/628/commits/5cb032ee2750853d4873806065ccf3293baa3fed (`set-mint-authority`).
I'll move the remaining commits to another follow-up PR.

evgeniko (approved):
lgtm

cli/src/index.ts
```diff
@@ -883,6 +884,281 @@ yargs(hideBin(process.argv))
...
+                        );
+                        const solanaNtt = ntt as SolanaNtt<typeof network, SolanaChains>;
+                        const tokenAuthority = NTT.pdas(chainConfig.manager).tokenAuthority();
+              
+                        // check if SPL-Multisig is supported for manager version
+                        const major = Number(solanaNtt.version.split(".")[0]);
+                        if (major < 3) {
+                          console.error(
+                            "SPL Multisig token mint authority is only supported for versions >= 3.x.x"
+                          );
```
evgeniko, line 944:
```suggestion
                          );
                              console.error("Use 'ntt upgrade' to upgrade the NTT contract to a specific version.");
```

cli/src/index.ts
```diff
@@ -883,6 +884,281 @@ yargs(hideBin(process.argv))
...
+                        );
+                        if (!mintInfo.mintAuthority) {
+                          console.error(
+                            "Token has fixed supply and no further tokens may be minted"
+                          );
+                          process.exit(1);
+                        }
+                        if (!mintInfo.mintAuthority.equals(payerKeypair.publicKey)) {
+                          console.error(
+                            `Current mint authority (${mintInfo.mintAuthority.toBase58()}) does not match payer (${payerKeypair.publicKey.toBase58()}). Retry with current authority`
```
evgeniko, line 1049: why does the payer need to be the same as the current mint authority?
Besides that, how would I specify the pda of the ntt manager here, that is usually the current mint authority as a payer?

nvsriram: I've updated it so that it exits with an error comment that links to the demo repo if the current authority is the `token-authority` PDA or a valid SPL Multisig.

evgeniko (approved):

PR #634, opened 2025-06-17
solana/ts/sdk: token 2022 get ATA fix

kev1n-peters:
We need to pass in the correct token program when fetching the ATA.


nvsriram (approved):
LGTM!

emreboga (approved):

PR #637, opened 2025-06-19
ts/sdk: executor per token gas limit overrides

kev1n-peters:
It's useful to override the gasLimit set by the executor for certain tokens that might use more gas when redeeming.


nvsriram (approved):
LGTM!

emreboga (approved):
LGTM!

PR #403, opened 2024-04-12
solana: deterministic builds via Docker

kcsongor:

PR #646, opened 2025-06-27
solana: Deterministic builds via Docker

nvsriram:
This PR was initially based on https://github.com/wormhole-foundation/native-token-transfers/pull/403. However, due to reasons mentioned in [this comment](https://github.com/wormhole-foundation/native-token-transfers/pull/646#issuecomment-3014103037), the Dockerfile changes were reverted.

tl;dr: 
This PR adds:
* Add `artifacts-mainnet`, `artifacts-solana-devnet`, and `artifacts-tilt-devnet` Makefile target to generate verifiable build artifacts using `solana-verify build`
* Remove `prod-build` Makefile target

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -12,7 +12,7 @@ docker_build(
     context = "./",
     only = ["./sdk", "./solana"],
     ignore=["./sdk/__tests__", "./sdk/Dockerfile", "./sdk/ci.yaml", "./sdk/**/dist", "./sdk/node_modules", "./sdk/**/node_modules"],
-    target = "builder",
+    target = "dev-builder",
     dockerfile = "./solana/Dockerfile",
 )
 docker_build(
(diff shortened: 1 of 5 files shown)


nvsriram:
### Solutions considered

**Custom Docker stage** ❌

Instead of spinning up Docker ourselves and building it, `make prod-build` that calls `anchor build --verifiable` under the hood does the same thing. Hence, we don't need the Dockerfile changes to generate the artifacts.


**`anchor build --verifiable`** ❌ 

This fails when trying to build for Anchor `v0.29.0`. This is because [Anchor v0.29.0 Dockerfile](https://github.com/solana-foundation/anchor/blob/v0.29.0/docker/build/Dockerfile#L41) uses the old domain (`release.solana.com`) to install Solana CLI which now fails. Hence, we remove the `prod-build` target as well.


**`solana-verify build`** ✅ 

This works similar to `anchor build --verifiable` but [uses the new recommended domain](https://github.com/Ellipsis-Labs/solana-verifiable-build/blob/master/docker/v1.18.26.Dockerfile) (`release.anza.xyz`) to install Solana CLI. We thus use this as the preferred solution.

nik-suri (approved):

evgeniko (approved):

PR #647, opened 2025-06-30
Fix nonce synchronization issues in EVM deployments by adding slow flag to forge

evgeniko:
EVM deployments were failing with nonce errors on certain networks (e.g. Mezo) due to forge sending transactions too quickly without waiting for confirmation between transactions.
Example error:
`server returned an error response: error code -32000: invalid nonce; got 3, expected 2: invalid sequence: invalid sequence`

Added --slow flag to forge script commands in both `deployEvm` and `upgradeEvm` functions. This flag forces forge to wait for transaction confirmation before sending the next one, ensuring proper nonce synchronization.

### DIFF
--- a/cli/src/index.ts
+++ b/cli/src/index.ts
@@ -1470,7 +1470,7 @@ async function upgradeEvm<N extends Network, C extends EvmChains>(
 --sig "upgrade(address)" \
 ${ntt.managerAddress} \
 ${signerArgs} \
---broadcast \
+--broadcast --slow \
 ${verifyArgs} | tee last-run.stdout`, {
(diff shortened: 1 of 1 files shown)


dvgui (dismissed):

nik-suri:
@evgeniko adding on to what @kcsongor said -- could we solve this by exposing the `--slow` flag as an option so CLI users can apply it on a case-by-case basis whenever needed?

evgeniko:
Added helper function that adds slow flag only if Mezo is used as a hardcoded value.

kcsongor (approved):

PR #649, opened 2025-07-03
cli: Handle Solana initialization flow for undeployed program and SPL Multisig mint authority

nvsriram:
This PR adds:
* Update `set-mint-authority` command to work as described in [this comment](https://github.com/wormhole-foundation/native-token-transfers/pull/649#issuecomment-3033224330)
* Update `create-spl-multisig` command to work similar to `set-mint-authority` and is described in the same comment
* Update `deploySolana` to handle NTT initialization with current mint authority being valid SPL Multisig
* Add `checkSolanaValidSplMultisig` helper that checks if address corresponds to valid SPL Multisig

### DIFF
--- a/cli/src/index.ts
+++ b/cli/src/index.ts
@@ -1015,28 +1015,37 @@ yargs(hideBin(process.argv))
                     })
                 .command("create-spl-multisig <multisigMemberPubkey...>",
                     "create a valid SPL Multisig (see https://github.com/wormhole-foundation/native-token-transfers/tree/main/solana#spl-multisig-support for more info)",
-                    (yargs) =>
-                        yargs
-                          .positional("multisigMemberPubkey", {
-                              describe:
-                                "public keys of the members that can independently mint",
-                              type: "string",
-                              demandOption: true,
-                          })
-                          .option("path", options.deploymentPath)
-                          .option("payer", { ...options.payer, demandOption: true })
-                          .example(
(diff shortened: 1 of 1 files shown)


evgeniko:
What do you think about adding some consistent formatting rules to not run into indentation with different PRs? Will make readability of new PRs also easier.
Maybe we add a prettierrc or eslintrc.json?

nvsriram:
> What do you think about adding some consistent formatting rules to not run into indentation with different PRs? Will make readability of new PRs also easier. Maybe we add a prettierrc or eslintrc.json?

Yes, I agree. This was something I had messed around with briefly. However, I believe that would be its own cleanup PR as it affects all TS files across all directories.

evgeniko (dismissed):

evgeniko:
tested:
`set-mint-authority` :
1.Passing only one of --token and --manager -- should error saying pass both
2. Undeployed program: Not passing both -- should error saying chain not found
3- Undeployed program: Passing --multisig with no input -- should error saying this value is missing
4. Undeployed program: Passing --multisig with invalid input -- should show version warning and request confirmation (if --yes is not passed). On confirming, this should fail saying SPL Multisig is not valid
5. Undeployed program: Passing --multisig and --yes with valid input (token & manager flags) -- should set to valid SPL Multisig without confirmation request
6. Different Undeployed program: Passing nothing extra should set to token_authority PDA
7. Deployed program: Without passing --token and --manager -- should set to token_authority PDA
8. Deployed program: Without passing --token and --manager and just --multisig with valid input -- should set to valid SPL Multisig

PR #653, opened 2025-07-14
Update INttToken.sol

nik-suri:
Clarify that `setMinter` is not a required function for integration with NTT.


evgeniko (dismissed):

dvgui (commented):
A couple of suggestions:

    The `mint` function is actually required for all spoke chain tokens, as well as for tokens operating in burn-and-mint manager mode. It is only not required for hub-and-spoke manager chain tokens and for the hub itself. Clarifying this in the comments would help avoid confusion for integrators.
    For `setMinter`, I agree it’s not required and should be presented as optional or removed. There are multiple legitimate ways to implement minter management, such as using role-based access control or other custom approaches, not just a single setter. It might be helpful to mention these alternatives in the comment to encourage flexibility for implementers.

```solidity
// NOTE: The `setMinter` method is not present in the standard ERC20 interface.
//       This is an optional function for integration with NTT. There are multiple ways to manage token minters (e.g., role-based access control, custom implementations).
//       Providing flexibility to change or assign multiple minters is recommended, but the exact mechanism can vary.
```

nik-suri:
Thanks @dvgui! I agree and I've updated both comments based on your feedback above.

evgeniko (approved):

PR #657, opened 2025-07-18
cli: make set-mint-authority a top-level command with chain parameter support

evgeniko:


kcsongor:
https://github.com/wormhole-foundation/native-token-transfers/commit/59b6e785dcb06d64be267ec5a34501a221250efd here's a branch I started which has some overlap, but should show the general direction

cli/src/index.ts
```diff
@@ -956,6 +956,324 @@ yargs(hideBin(process.argv))
...
+                }
+            }
+            // undeployed case
+            else {
+                // verify new authority is valid SPL Multisig
+                if (multisigTokenAuthority) {
+                    // NOTE: the `acceptTokenAuthority` instruction would have done this check normally.
+                    // However, since the program is not yet deployed, we do this check here.
+                    const isMultisigTokenAuthority =
+                        await checkSolanaValidSplMultisig(
```
nvsriram, line 1234: If SPL Multisig works the same way for Fogo, we should probably rename the function to `checkSvmValidSplMultisig` to make it more generic.

```suggestion
                        await checkSvmValidSplMultisig(
```

nvsriram (changes requested):
I believe as a consequence, the original [helper function](https://github.com/wormhole-foundation/native-token-transfers/pull/657/files#diff-4e5d53b79780ec257e845296172b0e131f58a02e2203429882c5141c8af678ecR2806) would also need to be renamed to `checkSvmValidSplMultisig` along with all other references of it in this file.

PR #658, opened 2025-07-21
set wormhole address in solana binary by patching it, to enable other SVM chains to work with the CLI

evgeniko:
From @kcsongor in commit e43c778

"cli: set wormhole address in solana binary by patching it
The problem: solana binaries contain program addresses embedded into
them. Specifically, NTT includes the address of the Wormhole program
it's linked against. Which Wormhole address to include is controlled via
feature flags in the wormhole-anchor-sdk crate, which supports one of
mainnet, devnet, and localhost. For other SVM chains, like Fogo, we
can't build this binary correctly, because there is no way to tell it
which core address to use.

One solution is to roll out an updated version of the
wormhole-anchor-sdk, which allows setting the Wormhole address via an
environment variable. That is likely the prudent solution, but it will
need to be backported to older NTT versions (in case a Fogo deployer
wants to deploy, say, 3.0.0, instead of the main HEAD).

Another solution, which we adopt in this commit, is simply patching the
binary. In other words: compile the binary with the solana flag, then
replace the solana address with the desired address. This seems to work
just fine, there is no checksum signing on the binaries or anything. It
also makes compilation a bit faster, because we don't need to recompile
anything between different targets.

Not sure if we should keep this, but will do for now."

### DIFF
--- a/cli/src/index.ts
+++ b/cli/src/index.ts
@@ -2002,7 +2002,10 @@ async function deploySolana<N extends Network, C extends SolanaChains>(
             console.error("Core bridge address not found in Solana config");
             process.exit(1);
         }
-        await patchSolanaBinary(binary, wormhole, solanaAddress);
+        
+        if (ch.chain !== "Solana") {
+            await patchSolanaBinary(binary, wormhole, solanaAddress);
+        }
         await checkSolanaBinary(binary, wormhole, providedProgramId, version ?? undefined);
 
         // if buffer.json doesn't exist, create it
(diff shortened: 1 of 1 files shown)


nvsriram (commented):
Is there a reason why this we check this again when it is already done at the start of this function [here](https://github.com/wormhole-foundation/native-token-transfers/pull/658/files#diff-4e5d53b79780ec257e845296172b0e131f58a02e2203429882c5141c8af678ecR1833-R1837)? Or am I missing something?

nvsriram:
> Is there a reason why we check this again when it is already done at the start of this function [here](https://github.com/wormhole-foundation/native-token-transfers/pull/658/files#diff-4e5d53b79780ec257e845296172b0e131f58a02e2203429882c5141c8af678ecR1833-R1837)? Or am I missing something?

Nvm, I misread what the check was doing.

PR #664, opened 2025-07-31
add celo testnet executor address

evgeniko:


nik-suri (approved):
Confirmed that this is the verified `NttManagerWithExecutor` contract on Celo testnet

PR #670, opened 2025-08-12
cli: fix mint authority check for locking mode

evgeniko:
checking for mint authority should not be done for locking mode deployments on Solana


nvsriram (approved):
LGTM!

PR #681, opened 2025-08-22
[chore]: Bump wormhole ts sdk to 3.4.4

bradleyplaydon:
- Bump wormhole ts sdk to 3.4.3 
- Remove bun lock
- Remove "resolutions" from root package.json as we don't use pnpm or yarn in this repo yet, this was causing a mismatch of @mysten/sui dep when running npm install which broke the build
- Replaced import with exported IDL const
- Add temporary manual increment to sign function in EvmNativeSigner class

-----------
### Blocked
- More issues where signSendWait is used, can't guarantee they will work without passing a nonce.

### DIFF
--- a/evm/ts/package.json
+++ b/evm/ts/package.json
@@ -48,10 +48,10 @@
     "ethers": "^6.5.1"
   },
   "peerDependencies": {
-    "@wormhole-foundation/sdk-base": "^3.4.3",
-    "@wormhole-foundation/sdk-definitions": "^3.4.3",
-    "@wormhole-foundation/sdk-evm": "^3.4.3",
-    "@wormhole-foundation/sdk-evm-core": "^3.4.3"
+    "@wormhole-foundation/sdk-base": "^3.4.4",
+    "@wormhole-foundation/sdk-definitions": "^3.4.4",
+    "@wormhole-foundation/sdk-evm": "^3.4.4",
+    "@wormhole-foundation/sdk-evm-core": "^3.4.4"
   },
   "devDependencies": {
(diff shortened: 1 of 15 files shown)


sdk/examples/package.json
```diff
@@ -32,10 +32,10 @@
     "tsx": "^4.7.2"
   },
   "dependencies": {
-    "@wormhole-foundation/sdk": "^2.4.0",
+    "@wormhole-foundation/sdk": "^3.4.3",
     "@wormhole-foundation/sdk-definitions-ntt": "1.0.0",
```
priyajeet, line 36: We should also make sure all the 1.0.0 and 1.0.6 versions of "@wormhole-foundation/sdk are also updated to 2.0.1 since that will reduce mysten/sui errors too

evm/ts/package.json
```diff
@@ -48,10 +48,10 @@
     "ethers": "^6.5.1"
   },
   "peerDependencies": {
-    "@wormhole-foundation/sdk-base": "^2.4.0",
-    "@wormhole-foundation/sdk-definitions": "^2.4.0",
-    "@wormhole-foundation/sdk-evm": "^2.4.0",
-    "@wormhole-foundation/sdk-evm-core": "^2.4.0"
+    "@wormhole-foundation/sdk-base": "^3.4.3",
```
priyajeet, line 51: Should update the 1.0.0 dep in deps

priyajeet (dismissed):

priyajeet (dismissed):

evan-gray:
closing in favor of https://github.com/wormhole-foundation/native-token-transfers/pull/702

PR #690, opened 2025-09-04
Feature/blockscout verification

aapclark:


cli/src/index.ts
```diff
@@ -4188,6 +4179,30 @@ function retryWithExponentialBackoff<T>(
...
+        case "blockscout":
+            const verifierUrl = configuration.get(chain, "verifier_url", { reportError: true });
+            if (!verifierUrl) {
+                console.error(`verifier_url is required when using blockscout verifier for ${chain}`);
+                process.exit(1);
+            }
+            return ["--verify", "--verifier", "blockscout", "--verifier-url", verifierUrl, "--verifier-api-key", apiKey];
+        case "etherscan":
+        default:
+            return ["--verify", "--verifier", "etherscan", "--verifier-api-key", apiKey];
```
evan-gray, line 4202: is there a reason this shouldn't be the same as the return before?

```suggestion
            return ["--verify", "--etherscan-api-key", apiKey];
```

cli/src/index.ts
```diff
@@ -4188,6 +4179,30 @@ function retryWithExponentialBackoff<T>(
     return attempt(0);
 }
 
+function buildVerifierArgs(chain: Chain): string[] {
+    const verifier = configuration.get(chain, "verifier", { reportError: true });
+    const apiKey = configuration.get(chain, "scan_api_key", { reportError: true });
+
+    if (!apiKey) {
```
evan-gray, line 4186: I didn't think blockscout required an API key per their [docs](https://docs.blockscout.com/devs/verification/foundry-verification). If that is the case, this check should be moved to the default case.

evan-gray (approved):
not confident on the specifics but seems like a reasonable change.

cli/src/index.ts
aapclark: The docs say the key is optional which I took to mean it is up to the explorer host to determine whether or not to require an API key. If that's not the case, then I'm happy to make the above changes.

XRPL-EVM's explorer doesn't require a key.

dvgui (approved):

evan-gray (approved):

PR #694, opened 2025-09-10
docs: Update and clarify invariants

johnsaigle:
- Make Pause invariant more explicit in terms of what pausing should limit
- Make it clear that the transceiver count should never go back to 0 after initial deployment
- Add information about registering vs enabling transceivers, why transceivers can't be deleted, and how it relates to business logic


nvsriram (dismissed):
LGTM!

djb15 (dismissed):

PR #711, opened 2025-10-14
cli: remove SR-related code

evgeniko:
- remove all standard relaying related code from the CLI and deployment scripts as it is now handled by the executor
- add hasExecutorDeployed function to check if the used chain has an Executoer deployed, otherwise ntt transfers won't be relayed


nvsriram (approved):
LGTM!

PR #714, opened 2025-10-16
sdk/sui: fix for rpc returning Currency<T> instead of CoinMedata<T>

kev1n-peters:
The sui rpc now returns Currency<T> instead of CoinMetadata<T>. Sui NTT contracts still require us to pass the CoinMetadata object ID.


kcsongor (approved):

package.json
```diff
@@ -40,5 +40,10 @@
     "sdk/route",
     "sdk/examples",
     "cli"
-  ]
+  ],
+  "overrides": {
+    "ts-jest": {
+      "@types/jest": "^29.5.12"
```
kev1n-peters: Was getting `Missing: @types/jest@27.5.2 from lock file` without this. The wormchain SDK has ^27 as a peer dep. Didn't happen to me locally, and the lockfile was indeed updated, confusing.

kcsongor (approved):

PR #663, opened 2025-07-30
solana: Update `ntt_transceiver` program to use Wormhole shims

nvsriram:
This PR is based on this transceiver prototype commit: https://github.com/wormhole-foundation/native-token-transfers/commit/66940fe0751d1bc76f0ed4c31ef78252557f10d1.

This PR adds:
* Update `ntt_transceiver` program to work with Wormhole Post Message and Wormhole Verify VAA shim programs
* Update `SolanaNtt` and `SolanaNttWormholeTransceiver` classes to handle shim variants
* Add program binaries and IDLs for interacting with the shim programs
* Add `TestWormholePostMessageShim` utility helper class
* Fix Cargo and Anchor test errors
* Update IDL

PR #718, opened 2025-10-22
sdk: Add `svmShims` to `Ntt.Contracts` to configure Wormhole Shims

nvsriram:
This PR adds:
* Update `Ntt.Contracts` to add a `svmShims` field that works like how [the current Solana shimOverrides argument in the constructor works](https://github.com/wormhole-foundation/native-token-transfers/pull/663#issuecomment-3137468968) — the main difference being that it is now backwards compatible
* Update `SolanaNtt` and `SolanaNttWormholeTransceiver` classes to use the `svmShims` field
* Add default shim addresses as a `SolanaChains`-keyed constant that can be reference by `SolanaNttWormholeTransceiver`
* Update tests

### DIFF
--- a/solana/ts/sdk/ntt.ts
+++ b/solana/ts/sdk/ntt.ts
@@ -72,7 +72,7 @@ export class SolanaNttWormholeTransceiver<
     readonly manager: SolanaNtt<N, C>,
     readonly program: Program<NttBindings.Transceiver<IdlVersion>>,
     readonly version: string = "3.0.0",
-    readonly svmShims: Ntt.Contracts["svmShims"] = {}
+    readonly svmShims: Ntt.Contracts["svmShims"] = undefined
   ) {
     this.programId = program.programId;
     this.pdas = NTT.transceiverPdas(program.programId);


sdk/definitions/src/ntt.ts
```diff
@@ -42,6 +44,31 @@ export namespace Ntt {
...
     };
     quoter?: string;
+    svmShims?: {
+      postMessageShimOverride?: string;
+      verifyVaaShimOverride?: string;
+    };
+  };
+
+  export const DEFAULT_SVM_SHIM_ADDRESSES: {
+    [chain in SolanaChains]: {
```
kev1n-peters: Can we change this to `[chain in Chains]?` actually so we don't have to import from `@wormhole-foundation/sdk-solana` above? That package is not in dependencies in this package.

nvsriram: Sure, but I think `Chains` would not limit it to just the SVM chains...
How about I replace it with something like:
```typescript
export const DEFAULT_SVM_SHIM_ADDRESSES: {
    [chain in platform.PlatformToChains<"Solana">]: {
```

solana/ts/sdk/ntt.ts
```diff
@@ -76,21 +72,37 @@ export class SolanaNttWormholeTransceiver<
     readonly manager: SolanaNtt<N, C>,
     readonly program: Program<NttBindings.Transceiver<IdlVersion>>,
     readonly version: string = "3.0.0",
-    readonly shimOverrides: WormholeShimOverrides | null = {}
+    readonly svmShims: Ntt.Contracts["svmShims"] = {}
```
kev1n-peters: We should pass this in from the manager in its constructor (the manager should have access to it from the config). We should make sure that it gets passed as undefined when it's not set in the config

nvsriram: Right now, `svmShims` is being passed from the manager [in its constructor](https://github.com/wormhole-foundation/native-token-transfers/pull/718/files#diff-b07cf1c6f7cea79d625ed64b75833e0124127bf393b0a463f9a8297f635c5cbdR623) via the config.
But I will update the default to be `undefined` so that it gets passed in as `undefined`

kev1n-peters (approved):

kev1n-peters:
Confirmed the changes work in Connect:
transfer out: https://explorer.solana.com/tx/4yHeDg5YqFdotVFgsGxLsPakpiman4vfTqW9TENuH9peB4LBMMNrXPKkU1SLpK7SkoaxPA9B7vQYJnukhbDd1Dmf
transfer in: (in-progress)

evgeniko (approved):

PR #848, opened 2026-03-23
feat(solana): Make transceiver peers upgradable

nvsriram:
Currently, the `set_transceiver_peer` instruction can only be called once since it uses `init`. This means if the transceiver peer needs to be modified, then the transceiver needs to be redeployed. This PR changes the `set_transceiver_peer` instruction to be `init_if_needed` so that we can update previously set transceiver peers.

This also adds unit tests for `update_peer` and `update_transceiver_peer`

## Summary by CodeRabbit

* **Tests**
  * Added tests for peer configuration updates and transceiver peer address updates
  * Enhanced test infrastructure and initialization to ensure on-chain account verification and registrations run during test startup

* **Chores**
  * Enabled mainnet feature flag in build, test, and lint targets
  * Adjusted dependency/configuration to allow conditional account initialization during operations

### DIFF
--- a/solana/Makefile
+++ b/solana/Makefile
@@ -16,7 +16,7 @@ build: cargo-build anchor-build
 
 
 cargo-build:
-	BPF_OUT_DIR="$(pwd)/target/deploy" cargo build-sbf
+	BPF_OUT_DIR="$(pwd)/target/deploy" cargo build-sbf --features "mainnet"
 
 # After building, remove the generics from the idl file. This is necessary as of anchor 0.29.0,
 # because the javascript library does not support generics yet, and just panics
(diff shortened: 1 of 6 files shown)


kcsongor (approved):

djb15:
I believe preventing transceiver peer updates is deliberate...see [here](https://github.com/wormhole-foundation/example-native-token-transfers/blob/a13746a430fbd4a033aef0ff8180d0d49fdbbab3/evm/src/Transceiver/WormholeTransceiver/WormholeTransceiverState.sol#L166-L168)

johnsaigle:
Closing this as not planned following Dirk's comment above.

PR #852, opened 2026-03-25
fix: increase bun:test default timeout

emreboga:
Every test in this file spawns CLI subprocesses, but only some had the 30s timeout, the rest used Bun's 5s default. We need to fix that by using `setDefaultTimeout(SUBPROCESS_TIMEOUT)` for the entire file.

## Summary by CodeRabbit

* **Tests**
  * Streamlined test timeout configuration by implementing a global default timeout, reducing redundant per-test specifications while maintaining consistent test execution.


evan-gray (approved):
worth a shot

DBosley (approved):

PR #854, opened 2026-03-30
fix(sdk): deprecate NttAutomaticRoute in favor of NttExecutorRoute

evgeniko:
## Summary

- Mark `nttAutomaticRoute` / `NttAutomaticRoute` as `@deprecated` with pointer to `nttExecutorRoute` / `NttExecutorRoute`
- Add runtime `console.warn` in `EvmNtt.quoteDeliveryPrice` when `automatic: true` is passed
- Update `sdk/examples/src/route.ts` to use `nttExecutorRoute`

## Context

The EVM on-chain special/standard relayer pricing was removed from `WormholeTransceiver` in 36570ce4 ("SR removal and CCL"), but older deployed transceivers (v1.0.0/v1.1.0) still have `isSpecialRelayingEnabled` active with stale pricing. This causes `quoteDeliveryPrice` with `automatic: true` to return inflated fees (e.g. 0.6 ETH for an ETH→Solana transfer) via the unmaintained special relayer contract.

The executor route (`NttExecutorRoute`) replaced this path and quotes via an off-chain service, returning correct fees. The Solana automatic relay path (via the on-chain NttQuoter program) is unaffected and still valid.

## Summary by CodeRabbit

* **Deprecations**
  * Automatic NTT routing for EVM chains is deprecated; on-chain relayer pricing is no longer maintained and may yield stale or inflated delivery quotes. Migrate to executor-based routing with quote support.
* **Documentation**
  * Deprecated APIs now include deprecation notes and runtime warnings directing users to the executor-based approach.
* **Examples**
  * Sample configurations updated to demonstrate the executor-based routing pattern.

### DIFF
--- a/evm/ts/src/ntt.ts
+++ b/evm/ts/src/ntt.ts
@@ -529,6 +529,7 @@ export class EvmNtt<N extends Network, C extends EvmChains>
         "EvmNtt.quoteDeliveryPrice with automatic: true queries on-chain " +
           "special/standard relayer pricing which is no longer maintained " +
           "and may return inflated fees. " +
+          "Any messages sent via the Standard Relayer will not be automatically delivered to the destination chain past April 1st, 2026. " +
           "For automatic EVM transfers, use nttExecutorRoute and obtain quotes via route.quote() instead."
       );
     }
--- a/evm/ts/src/ntt.ts
+++ b/evm/ts/src/ntt.ts
@@ -524,6 +524,15 @@ export class EvmNtt<N extends Network, C extends EvmChains>
     dstChain: Chain,
     options: Ntt.TransferOptions
   ): Promise<bigint> {
+    if (options.automatic) {
+      console.warn(
+        "EvmNtt.quoteDeliveryPrice with automatic: true queries on-chain " +
+          "special/standard relayer pricing which is no longer maintained " +
+          "and may return inflated fees. " +
+          "For automatic EVM transfers, use nttExecutorRoute and obtain quotes via route.quote() instead."
+      );
+    }
+
     await this.initTransceiverIndices();
 
     // Bypass the manager's quoteDeliveryPrice because it has a bug: it sizes the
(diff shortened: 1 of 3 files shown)


tonyjin (dismissed):

evm/ts/src/ntt.ts
```diff
@@ -524,6 +524,15 @@ export class EvmNtt<N extends Network, C extends EvmChains>
     dstChain: Chain,
     options: Ntt.TransferOptions
   ): Promise<bigint> {
+    if (options.automatic) {
+      console.warn(
+        "EvmNtt.quoteDeliveryPrice with automatic: true queries on-chain " +
+          "special/standard relayer pricing which is no longer maintained " +
+          "and may return inflated fees. " +
```
nik-suri, line 531: Suggest adding "Any messages sent via the Standard Relayer will not be automatically delivered to the destination chain."

sdk/route/src/automatic.ts
```diff
@@ -40,13 +40,25 @@ type Q = routes.Quote<Op, Vp>;
 
 type R = NttRoute.AutomaticTransferReceipt;
 
+/**
+ * @deprecated Use {@link nttExecutorRoute} instead.
+ * When the source chain is EVM, this route relies on on-chain special/standard relayer pricing
+ * which is no longer maintained. Older EVM transceiver deployments may return stale or inflated
+ * delivery quotes.
```
nik-suri: Instead of "Older EVM transceiver deployments may return stale or inflated delivery quotes", let's say: "Messages sent via the Standard Relayer will not be automatically delivered to the destination chain past April 1st, 2026."

sdk/route/src/automatic.ts
```diff
@@ -40,13 +40,25 @@ type Q = routes.Quote<Op, Vp>;
...
     static override config = config;
   }
   return NttRouteImpl;
 }
 
+/**
+ * @deprecated Use {@link NttExecutorRoute} instead.
+ * When the source chain is EVM, this route relies on on-chain special/standard relayer pricing
+ * which is no longer maintained. Older EVM transceiver deployments may return stale or inflated
+ * delivery quotes.
```
nik-suri: Same comment as the above

sdk/route/src/automatic.ts
evgeniko: Good call on adding the delivery deadline, I'll add it, but I would still keep the stale quotes note too since that's the main issue integrators have been hitting. An explanation for why the quote is very high is important here imo.

PR #856, opened 2026-04-09
fix: use u64::MAX as default Solana outbound limit

bengtlofgren:
Sets limits on solana/svm to match that of evm

## Summary by CodeRabbit

* **Chores**
  * Updated outbound limit configuration to use the maximum 64-bit unsigned integer value.


kcsongor (approved):
great patch

PR #861, opened 2026-04-21
feat(cli): support precompile token deployment + Tempo gas limit

douglasgalico:
Chains with precompile/system contract tokens (e.g. Tempo) cause `OpcodeNotFound` errors in Forge's local EVM. When simulation fails with `OpcodeNotFound` or `StaticcallFailed`, the CLI now:

1. Runs forge script without --broadcast to generate dry-run tx JSON
2. Fixes gas limits via eth_estimateGas for CREATE txs
3. Broadcasts with --resume using corrected gas values

Also:
- Add Tempo per-chain gas limit override (2M) in signer.ts
- Bump SDK to 4.17.1 (sendWait fix for custom tx types)
- Improve e2e waitForRpc to detect early Anvil process exit

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -51,25 +51,25 @@
         "typechain": "^8.3.2",
       },
       "peerDependencies": {
-        "@wormhole-foundation/sdk-base": "^4.17.0",
-        "@wormhole-foundation/sdk-definitions": "^4.17.0",
-        "@wormhole-foundation/sdk-evm": "^4.17.0",
-        "@wormhole-foundation/sdk-evm-core": "^4.17.0",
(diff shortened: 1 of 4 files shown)


bengtlofgren (dismissed):
cool, but man can we get rid of carets across the org lol

bengtlofgren (dismissed):

kcsongor (commented):
the logic seems sensible but can we refactor the deploy evm function to just support doing this? rebuilding the same command multiple times looks a bit brittle

kcsongor (approved):

bengtlofgren (approved):

PR #873, opened 2026-05-06
ci: Add zizmor and fix violations

johnsaigle:


johnsaigle:
Superseded by org-wide configuration.

PR #875, opened 2026-05-07
solana: multi-tenant deployment

kcsongor:


johnsaigle:
@claude

johnsaigle (changes requested):
Needs rebase

solana/programs/example-native-token-transfers/src/instructions/release_inbound.rs
```diff
@@ -29,7 +29,7 @@ pub struct ReleaseInbound<'info> {
     pub recipient: InterfaceAccount<'info, token_interface::TokenAccount>,
 
     #[account(
-        seeds = [crate::TOKEN_AUTHORITY_SEED],
+        seeds = [crate::TOKEN_AUTHORITY_SEED, config.key().as_ref()],
```
johnsaigle, line 41: My understanding is that the `inbox_item` should also be validated against the instance (config) key. Otherwise, it seems possible to pass an InboxItem on Instance A and pass it through the release flow on Instance B. The recipient encoded in the message would not be altered, but the funds would be minted/unlocked out of the wrong instance. (This relies on both instances using the same token mint also -- maybe that disqualifies this scenario? If so, we should document explicitly)

It looks like OutboxItem was updated so I think we just need something similar for the InboxItem.

Here's a test I was playing with -- if useful, we could modify the logic slightly to ensure that the inbox item can't be used on the wrong instance.

(test under the `multi_instance.test.ts` file)
```ts

  it("demonstrates cross-instance release_inbound_mint uses an inbox item from another instance", async () => {
    const vaaForB = makeInboundVaa(
      "xinst-release-bug",
      instanceB.publicKey,
      payer.publicKey
    );

    // Prepare a valid, approved InboxItem for instance B, but do not release it.
    await redeemWithoutRelease(nttB, vaaForB, sender, signer);

    const recipientA = spl.getAssociatedTokenAddressSync(
      testMintA.address,
      payer.publicKey,
      true,
      TOKEN_PROGRAM
    );
    await spl.getOrCreateAssociatedTokenAccount(
      $.connection,
      payer,
      testMintA.address,
      payer.publicKey,
      true,
      undefined,
      undefined,
      TOKEN_PROGRAM,
      spl.ASSOCIATED_TOKEN_PROGRAM_ID
    );
    await assert.tokenBalance($.connection, recipientA).equal(0);

    // This should be rejected: config/mint/token_authority/custody are from A,
    // while inbox_item is the approved PDA for B. The current program accepts it.
    const maliciousReleaseIx = await nttA.program.methods
      .releaseInboundMint({ revertOnDelay: false, revertWhenNotReady: false })
      .accounts({
        common: {
          payer: payer.publicKey,
          config: { config: nttA.pdas.configAccount() },
          inboxItem: nttB.pdas.inboxItemAccount(
            "Ethereum",
            vaaForB.payload.nttManagerPayload
          ),
          recipient: recipientA,
          mint: testMintA.address,
          tokenAuthority: nttA.pdas.tokenAuthority(),
          tokenProgram: TOKEN_PROGRAM,
          custody: await NTT.custodyAccountAddress(
            nttA.pdas,
            await nttA.getConfig()
          ),
        },
        multisigTokenAuthority: multisigTokenAuthorityA,
      })
      .instruction();

    await $.sendAndConfirm(maliciousReleaseIx, payer);

    // 25_000 at 8 decimals released to a 9-decimal mint becomes 250_000 units.
    await assert.tokenBalance($.connection, recipientA).equal(250_000);
  });

File: solana/programs/example-native-token-transfers/src/transceivers/wormhole/instructions/receive_message.rs [hunk 2]
```diff
@@ -47,6 +47,7 @@ pub struct ReceiveMessage<'info> {
         space = 8 + ValidatedTransceiverMessage::<TransceiverMessageData<NativeTokenTransfer<Payload>>>::INIT_SPACE,
         seeds = [
             ValidatedTransceiverMessage::<TransceiverMessageData<NativeTokenTransfer<Payload>>>::SEED_PREFIX,
+            config.key().as_ref(),
```
johnsaigle, line 59: I think we ought to add a check here such that the `recipient_ntt_manager` matches the instance key. Otherwise I think you can take a VAA intended for an Instance A and create the transceiver message for Instance B. That incorrectly creates the transceiver message

I believe the impact here is low-to-nil for a few reasons:
* The exist seeds include the emitter chain and ID as seeds
* Solana only allows one peer per Wormhole emitter chain
* The IDs increase monotically on the sender side

This _should_ suffice to make the account resulting from the seeds truly unique. However if all of the above weren't true, it would be possible to e.g. relay messages sent from EVM peers intended for given SVM instances to the wrong SVM instance, and potentially front-run transceiver message accounts and DoS them via the `init` constraint.

Let me know if you agree with this read. IMO this is definitely worth a comment because understanding the security here relies on keeping a lot of NTT subtlety in your head.

johnsaigle (commented):
Two potential issues that I'd appreciate clarification on. 🙏🏻 

I have a few design questions as well -- maybe they would be best recorded as GitHub issues or else in a design document somewhere.

* Have you considered whether these changes a) are compatible with Fogo or b) how this change impacts SVM <-> SVM interactions? I haven't considered this in my review because I wanted to check whether it's worth spending time on first.

solana/programs/example-native-token-transfers/src/instructions/release_inbound.rs
kcsongor: this is valid, added a check for this!

solana/programs/example-native-token-transfers/src/instructions/initialize.rs
```diff
@@ -20,22 +16,17 @@ pub struct Initialize<'info> {
...
-    pub deployer: Signer<'info>,
-
-    #[account(
-        seeds = [crate::ID.as_ref()],
-        bump,
-        seeds::program = bpf_loader_upgradeable_program,
-    )]
-    program_data: Account<'info, ProgramData>,
+    /// The owner of the new instance. Distinct from the program's upgrade
+    /// authority — see the v4 trust-model note in the README.
```
johnsaigle, line 20: There are no README changes in this PR. Maybe the design document could be included in this PR? Easy to do in a follow-up PR though.

solana/programs/example-native-token-transfers/src/instructions/initialize.rs
```diff
@@ -52,18 +43,17 @@ pub struct Initialize<'info> {
...
         space = 8 + OutboxRateLimit::INIT_SPACE,
-        seeds = [OutboxRateLimit::SEED_PREFIX],
+        seeds = [OutboxRateLimit::SEED_PREFIX, config.key().as_ref()],
         bump,
     )]
     pub rate_limit: Account<'info, OutboxRateLimit>,
 
     #[account(
-        seeds = [crate::TOKEN_AUTHORITY_SEED],
+        seeds = [crate::TOKEN_AUTHORITY_SEED, config.key().as_ref()],
```
johnsaigle, line 52: @claude Verify whether the TS files in the SDK are correctly deriving the token and mint authorities near e.g. `cli/src/commands/set-mint-authority.ts:152`.

This change makes it so the instance (config) key is part of the seeds. The relevant code is not obviously shown in the diff so when proving/disproving this claim, quote the relevant code directly and include line numbers.

cli/src/commands/solana.ts
```diff
@@ -201,7 +217,8 @@ export function createSolanaCommand(
               const [, , ntt] = await pullChainConfig(
                 network,
                 { chain, address: toUniversal(chain, chainConfig.manager) },
-                overrides
+                overrides,
+                chainConfig.instance
```
johnsaigle, line 241: Further down in this file, the multisig is initialized with the token_authority based on the manager key. Should this include the instance key?

solana/programs/example-native-token-transfers/src/instructions/initialize.rs
johnsaigle: Lol Claude emoji reacted but didn't complete I guess?

solana/programs/example-native-token-transfers/src/instructions/redeem.rs
```diff
@@ -39,7 +39,8 @@ pub struct Redeem<'info> {
         // check that the message is targeted to this chain
         constraint = ValidatedTransceiverMessage::<NativeTokenTransfer<Payload>>::message(&transceiver_message.try_borrow_data()?[..])?.ntt_manager_payload().payload.to_chain == config.chain_id @ NTTError::InvalidChainId,
         // check that we're the intended recipient
-        constraint = ValidatedTransceiverMessage::<NativeTokenTransfer<Payload>>::message(&transceiver_message.try_borrow_data()?[..])?.recipient_ntt_manager() == crate::ID.to_bytes() @ NTTError::InvalidRecipientNttManager,
+        // v4: the on-the-wire manager identity is the instance's `config` pubkey.
```
mdulin2, line 42: I think we need an additional seed check here. 

Previously, when a transceiver could have only a single instance associated with it, this was safe. Now, I think a maliciously crafted `transceiver_message` from another deployment on the same transceiver could be used. The `id`, recipient manager, source manager, and destination chain are all controllable via a crafted VAA, since this data is just in the body.

solana/programs/ntt-transceiver/src/wormhole/instructions/release_outbound.rs
mdulin2, line 32: `MarkOutboxItemAsReleased` already checks that `outbox_item.manager == config.key()`. Could we add this here just for a sanity check? It's in the standalone version already.

mdulin2, line 44: Could we add a seed derivation for this? It's already checked in the `MarkOutboxItemAsReleased` on the CPI. But, it'd be good to add a second layer as a sanity check; it's already done in the standalone version of this.

solana/programs/example-native-token-transfers/src/transceivers/wormhole/instructions/receive_message.rs
johnsaigle: This should be applied to both files:

```
solana/programs/example-native-token-transfers/src/transceivers/wormhole/instructions/receive_message.rs
solana/programs/ntt-transceiver/src/wormhole/instructions/receive_message.rs
```

cli/src/commands/set-mint-authority.ts
johnsaigle: Does this need an instance key now?

cli/src/query.ts
johnsaigle: I think this needs an instance.key as well? Otherwise the query results will return the v3 style emitter which could lead to bad v4 peer registrations on other chains.

solana/programs/example-native-token-transfers/src/instructions/redeem.rs
mdulin2: Seeds look correct to me. Thanks for the change!

solana/programs/example-native-token-transfers/src/instructions/initialize.rs
kcsongor: added the README changes!

cli/src/query.ts
kcsongor: added!

kcsongor:
> * Have you considered whether these changes a) are compatible with Fogo or b) how this change impacts SVM <-> SVM interactions? I haven't considered this in my review because I wanted to check whether it's worth spending time on first.

Yes, no Fogo specific changes here. Added a paragraph in the README that explains these changes conform to the NTT spec in a backwards compatible way, so no SVM <-> SVM funny business (and more generally, a v4 NTT instance can be registered against older EVM/Sui/SVM/etc. NTTs)

johnsaigle (changes requested):
Hey @kcsongor, I noticed some other bugs in the TS portion. Could you please review the below issues and fix them if you agree?

* `push` registers the shared Solana program ID (manager), not the v4 instance. This can register peers that send transfers to an address a v4 instance cannot redeem.
[cli/src/validation.ts:L278](https://github.com/wormhole-foundation/native-token-transfers/blob/34b3a5f36470d0e1ae1635ae6809aad1a66a5f2c/cli/src/validation.ts#L278)

* push bypasses `canUpgrade`. A config push can apply a v3-to-v4 upgrade path and strand custody.
[cli/src/config-mgmt.ts:L83](https://github.com/wormhole-foundation/native-token-transfers/blob/34b3a5f36470d0e1ae1635ae6809aad1a66a5f2c/cli/src/config-mgmt.ts#L83)

* `canUpgrade` permits v4-to-v3 downgrade.
[cli/src/upgradeBarriers.ts:L55](https://github.com/wormhole-foundation/native-token-transfers/blob/34b3a5f36470d0e1ae1635ae6809aad1a66a5f2c/cli/src/upgradeBarriers.ts#L55)

* `upgradeSolana` does not pass the existing instance. This can create a new random authority and block v4 burning-mode upgrades.
Evidence: [cli/src/solana/deploy.ts:L732](https://github.com/wormhole-foundation/native-token-transfers/blob/34b3a5f36470d0e1ae1635ae6809aad1a66a5f2c/cli/src/solana/deploy.ts#L732)

* Manual commands omit `chainConfig.instance`. v4 peer setup and emergency redeem flows cannot build instance-scoped Solana state.
[cli/src/commands/manual.ts:L111](https://github.com/wormhole-foundation/native-token-transfers/blob/34b3a5f36470d0e1ae1635ae6809aad1a66a5f2c/cli/src/commands/manual.ts#L111)

* `addSolanaInstance` suppresses registration failures. add-chain can persist an instance that cannot transfer.
[cli/src/solana/deploy.ts:L705](https://github.com/wormhole-foundation/native-token-transfers/blob/34b3a5f36470d0e1ae1635ae6809aad1a66a5f2c/cli/src/solana/deploy.ts#L705)

johnsaigle:
Two more:

* token-transfer.ts drops `chainConfig.instance` when building both manual and executor routes. Any transfer involving a v4 Solana deployment fails because the SDK cannot construct instance-scoped state; the executor also derives legacy peer/config/LUT PDAs. [cli/src/commands/token-transfer.ts:L730](https://github.com/wormhole-foundation/native-token-transfers/blob/34b3a5f36470d0e1ae1635ae6809aad1a66a5f2c/cli/src/commands/token-transfer.ts#L730-L736)

* v4 deployment guidance omits `--instance` from the suggested set-mint-authority command. Following it transfers mint authority to the legacy PDA that the v4 instance cannot sign for, potentially stranding minting.
[cli/src/solana/deploy.ts:L495](https://github.com/wormhole-foundation/native-token-transfers/blob/34b3a5f36470d0e1ae1635ae6809aad1a66a5f2c/cli/src/solana/deploy.ts#L486-L497)

PR #884, opened 2026-05-18
ci: address zizmor workflow security audit findings

kev1n-peters:


.github/workflows/cli.yml
```diff
@@ -76,11 +81,13 @@ jobs:
...
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          persist-credentials: false
 
       - name: Restore cached v${{ matrix.version }} artifacts
         id: cache
-        uses: actions/cache@v5
+        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5 # zizmor: ignore[cache-poisoning]
```
kev1n-peters, line 92: zizmor sees `docker/build-push-action` below and assumes it publishes artifacts (so it flags "poisoned cache gets published"), but it doesn't actually publish anything

johnsaigle: Maybe worth in-lining this as a comment? It'd help us track it in case this is a bug in zizmor that gets fixed later.

johnsaigle (dismissed):

johnsaigle (approved):

.github/workflows/cli.yml
evan-gray: I do think this may be part of the larger issue where commits from forks don't work.

djb15 (approved):
I think the zizmor ignores for the RPC secrets are acceptable given the limited blast radius of stealing an RPC API key

evan-gray (approved):

PR #899, opened 2026-06-09
feat: migrate Sui SDK to @mysten/sui v2 gRPC and ESM-only

kev1n-peters:
Migrate the Sui packages to @mysten/sui ^2.17.0 (gRPC SuiGrpcClient, transport-agnostic core API, flat-json object shape). Because v2 is ESM-only, all dual-build workspaces drop CommonJS and ship ESM-only, and jest runs in native-ESM mode.

- sui/ts: rewrite ntt.ts/nttWithExecutor.ts/utils.ts for gRPC flat json (getObject json, listDynamicFields, getDynamicField, simulateTransaction, movePackageService.getDatatype, listCoins); base64 vector<u8>; owner union
- bump @wormhole-foundation/sdk-* to 5.0.0-beta.0 monorepo-wide; pin sdk-sui/sdk-sui-core exact via root overrides (bun caret resolves beta ranges to the latest stable tag, which dragged in @mysten/sui v1)
- ESM-only packaging for all dual-build workspaces; remove tsconfig.cjs.json
- migrate jest configs (sui/evm/route/definitions/xrpl) to native ESM

Addresses CORE-2802


sui/ts/src/ntt.ts
```diff
@@ -323,7 +326,7 @@ export class SuiNtt<N extends Network, C extends SuiChains>
   }
 
   async getTokenDecimals(): Promise<number> {
-    const coinMetadata = await this.provider.getCoinMetadata({
+    const { coinMetadata } = await this.provider.getCoinMetadata({
       coinType: this.contracts.ntt!["token"],
     });
```
bengtlofgren, line 332: What if `coin.decimals` actually is 0 - then we throw error hmm?

bengtlofgren (approved):

bengtlofgren (changes requested):
I would prefer to fix the decimals thing if we can

bengtlofgren (changes requested):
I would prefer to fix the decimals thing if we can

sui/ts/src/ntt.ts
kev1n-peters: fixed

bengtlofgren (approved):

douglasgalico (approved):

PR #907, opened 2026-06-18
feat(cli): XRPL subcommand group

M-Picco:
XRP Ledger commands used when preparing an NTT deployment on XRPL. XRPL has no smart contracts, so NTT relies on a Guardian-controlled custody model. These commands cover (1) creating/configuring the underlying XRPL token, (2) setting up the custody account and handing it off to the manager-set multisig, and (3) operating a live deployment (emitter derivation, VAA decoding, relaying).

For more detailed information see the README.md on the cli/src/commands/xrpl directory


cli/src/commands/xrpl/README.md
```diff
@@ -0,0 +1,347 @@
...
+- `--dst-addr` — destination address (hex32; recipient NTT manager for `ern1`)
+- `--src-manager` / `--manager` — source NTT manager emitter (hex32), or derive it
+  from the manager r-address/hex
+- `--token` (+ `--currency`/`--issuer`/`--mpt-id`) — defaults to the token inferred
+  from the tx's delivered amount
+- `--executor` (required) — Executor XRPL address to pay
+- `--executor-api`, `--guardian-api` — API base URLs (default: testnet)
+- `--gas-limit`, `--msg-value`, `--relay-instructions` — relay sizing
+- `--poll-interval`, `--poll-timeout` — VAA polling (ms)
+- `--seed` (or env `SEED`)
```
douglasgalico, line 303: Should we suggest the gas value/limits that will be needed for different operations within the sequencer?

cli/src/deployments.ts
```diff
@@ -28,6 +28,13 @@ export type HyperCoreConfig = {
   weiDecimals?: number;
 };
 
+// XRPL custody-account config. Kept as a dedicated top-level section (not under
+// `chains`) since the XRPL NTT account is set up incrementally and doesn't fit
+// the full ChainConfig shape.
+export type XrplConfig = {
+  manager?: string;
```
douglasgalico, line 35: To enrich with token Data: https://github.com/wormholelabs-xyz/bridge-monorepo/blob/d91d6de27b930ab86d21640df62d87bc3476fbe3/apps/portal-bridge/src/token-config/tokens.ntt.testnet.ts#L86-L137

douglasgalico (dismissed):

cli/src/commands/xrpl/README.md
```diff
@@ -0,0 +1,372 @@
...
+```
+ntt xrpl register-peer -n Testnet --peer-chain Solana --peer-address 0x… --admin-seed sEd7...
+```
+
+### `rotate-admin`
+
+Rotates the custody account's admin by publishing an `XADM` RotateAdmin (`0x02`)
+core message. Signed by the **current admin** (`--admin-seed`).
+
+> ⚠️ **Not yet supported by the Sequencer.** The message is published but will
```
evan-gray: is this accurate? I see a `handle_rotate_admin` command in the sequencer code

cli/src/commands/xrpl/init.ts
```diff
@@ -0,0 +1,179 @@
...
+        })
+        .option("ticket-count", {
+          describe: "Number of tickets available",
+          type: "number",
+          demandOption: true,
+        })
+        .option("app", {
+          describe: "App type",
+          type: "string",
+          choices: ["NTT", "WTT", "Core"] as const,
```
evan-gray:
```suggestion
          choices: ["NTT", "WTT", "CORE"] as const,
```

see https://github.com/wormholelabs-xyz/ripple/blob/0e1e5b55e2e6534d0eee2eb4f992d9cf7b0fe565/solana/xrpl-sequencer/programs/xrpl-sequencer/src/instructions/initialize.rs#L154-L158

though perhaps since this is the NTT CLI, it should just be hardcoded to NTT

cli/src/commands/xrpl/init.ts
```diff
@@ -0,0 +1,179 @@
...
+        })
+        .option("mpt-id", {
+          describe: "MPT issuance ID, 48-char hex [--token mpt]",
+          type: "string",
+        })
+        .option("core-account", {
+          describe:
+            "Wormhole Core (GMP) account to send the onboarding message to",
+          type: "string",
+          default: DEFAULT_TESTNET_CORE_ACCOUNT,
```
evan-gray: for this and the other functions that use it, I think this should not use default in this way, because when the network is mainnet it would still default to the testnet core account. instead, check this value in the function and or it with something like `getDefaultCoreAccountForNetwork(network)` which can return `DEFAULT_TESTNET_CORE_ACCOUNT` for testnet and otherwise throws (until there's a mainnet core account)

cli/src/commands/xrpl/relay.ts
```diff
@@ -0,0 +1,430 @@
...
+function resolveUniversalAddr(chain: Chain, addr: string): `0x${string}` {
+  const hex = addr.replace(/^0x/i, "");
+  if (/^[0-9a-fA-F]{64}$/.test(hex)) {
+    return `0x${hex.toLowerCase()}`;
+  }
+  return toUniversal(chain, addr).toString() as `0x${string}`;
+}
+
+function buildRequest(opts: {
+  requestType: "ern1" | "erv1";
```
evan-gray: seems odd to me that these are lowercase when in the spec / encoding they are uppercase. why not use `RequestPrefix.ERV1` and `RequestPrefix.ERN1` as the types?

cli/src/commands/xrpl/README.md
```diff
@@ -0,0 +1,372 @@
...
+onboarding payload encoding in [`../../xrpl/onboarding.ts`](../../xrpl/onboarding.ts),
+and the manager-set fetch in [`../../xrpl/manager-set.ts`](../../xrpl/manager-set.ts).
+The operational commands add emitter/token-id derivation
+([`tokenId.ts`](../../xrpl/tokenId.ts)), VAA/payload decoding
+([`payloads.ts`](../../xrpl/payloads.ts)), and the Executor/guardian clients +
+request layouts ([`executor.ts`](../../xrpl/executor.ts),
+[`guardian.ts`](../../xrpl/guardian.ts),
+[`executorLayouts.ts`](../../xrpl/executorLayouts.ts)).
+
+## Commands
```
evan-gray, line 20: it might be nice, in a future PR, to include a complete list of the series of example commands to do a fresh deployment of each type. e.g. WETH from Ethereum -> fresh IOU NTT deployment on XRPL

cli/src/commands/xrpl/relay.ts
```diff
@@ -0,0 +1,430 @@
...
+        })
+        .option("executor", {
+          describe: "Executor XRPL address to send the request payment to",
+          type: "string",
+          demandOption: true,
+        })
+        .option("executor-api", {
+          describe: "Executor API base URL",
+          type: "string",
+          default: DEFAULT_EXECUTOR_API,
```
evan-gray: similar issue with the default core address

cli/src/commands/xrpl/relay.ts
```diff
@@ -0,0 +1,430 @@
...
+        })
+        .option("guardian-api", {
+          describe: "Guardian / Wormholescan API base URL",
+          type: "string",
+          default: DEFAULT_GUARDIAN_API,
```
evan-gray: similar issue with the default core address

cli/src/xrpl/executor.ts
```diff
@@ -0,0 +1,74 @@
...
+
+export const DEFAULT_EXECUTOR_API = "https://executor-testnet.labsapis.com";
+
+export interface QuoteResponse {
+  signedQuote: `0x${string}`;
+  estimatedCost: string; // drops (stringified)
+}
+
+/** POST /v0/quote — fetch a signed quote + estimated cost for a relay. */
+export async function fetchQuote(opts: {
```
evan-gray: can/should this not reuse `fetchSignedQuote` from `sdk/route/src/executor`?

cli/src/xrpl/executor.ts
```diff
@@ -0,0 +1,74 @@
...
+  failureMessage?: string;
+  txs?: { txHash: string; chainId: number }[];
+  [k: string]: unknown;
+}
+
+/**
+ * POST /v0/status/tx — trigger indexing of `txHash` and return its relay status.
+ * `chainId` is the chain the request was emitted on (66 for XRPL).
+ */
+export async function submitStatusTx(opts: {
```
evan-gray: ditto re `fetchStatus`

cli/src/xrpl/executorLayouts.ts
```diff
@@ -0,0 +1,227 @@
+// Executor RequestForExecution layouts (binary-layout).
+//
+// Ported from ripple/xrpl-client/src/integration/executorLayouts.ts and
```
evan-gray: I thought all this was included in the wormhole ts sdk? if not, it probably should be but can be done in a later PR

cli/src/xrpl/guardian.ts
```diff
@@ -0,0 +1,68 @@
+// Thin client for the Guardian / Wormholescan signed-VAA API.
+//
+// Ported from ripple/xrpl-client/src/integration/fetch-vaa.ts. The guardian
```
evan-gray: are these not included in the wormhole ts sdk?

cli/src/xrpl/manager-set.ts
```diff
@@ -0,0 +1,108 @@
+import { ethers } from "ethers";
+import { decodeAccountID, deriveAddress } from "xrpl";
+
+// Fetches the Wormhole "delegated manager set" — the set of signer keys that
```
evan-gray, line 4: another good candidate for the ts sdk, but probably only relevant for NTT CLI for the time being.

cli/src/xrpl/onboarding.ts
```diff
@@ -0,0 +1,147 @@
...
+      );
+    }
+    case "mpt": {
+      if (!/^[0-9a-fA-F]{48}$/.test(token.mptId)) {
+        throw new Error("expected mpt issuance id length of 48 (hex)");
+      }
+      return (
+        dec +
+        padRight(
+          TOKEN_TYPE_MPT + token.mptId.toLowerCase(),
```
evan-gray: why `toLowerCase` here?

cli/src/xrpl/onboarding.ts
```diff
@@ -0,0 +1,147 @@
...
+/** Encode a non-negative integer as a 1-byte (2 hex char) value. */
+function u8Hex(n: number): string {
+  if (!Number.isInteger(n) || n < 0 || n > 0xff) {
+    throw new Error(`value ${n} does not fit in one unsigned byte`);
+  }
+  return n.toString(16).padStart(2, "0");
+}
+
+/** Encode a non-negative integer as a big-endian 8-byte (16 hex char) value. */
+function u64Hex(n: number): string {
```
evan-gray: I would expect u64 would need to be bigint or string

cli/src/xrpl/onboarding.ts
```diff
@@ -0,0 +1,147 @@
...
+  if (!Number.isInteger(n) || n < 0) {
+    throw new Error(`value ${n} must be a non-negative integer`);
+  }
+  return BigInt(n).toString(16).padStart(16, "0");
```
evan-gray: no upper limit check?

cli/src/commands/xrpl/init.ts
M-Picco: Will opt for an NTT constant value

cli/src/commands/xrpl/init.ts
M-Picco: Good catch

cli/src/xrpl/onboarding.ts
M-Picco: I think in the end it doesn't matter the casing since it's all hex. I'll remove it

cli/src/xrpl/guardian.ts
douglasgalico: Nope, just the basic platform and signer atm

cli/src/commands/xrpl/README.md
douglasgalico: Agree, from our E2E it was clear all the fine tuned parameters for each type. I can port them here :)

douglasgalico (dismissed):

cli/src/xrpl/executor.ts
```diff
@@ -0,0 +1,21 @@
+import type { Network } from "@wormhole-foundation/sdk-connect";
+
+const DEFAULT_EXECUTOR_APIS: Partial<Record<Network, string>> = {
```
evan-gray: these are already defined in the ts sdk https://github.com/wormhole-foundation/wormhole-sdk-ts/blob/065ac8d036be9408b41b8627415583bd5a303690/core/base/src/constants/executor.ts#L6 let's not duplicate them. perhaps this file could be avoided altogether.

cli/src/xrpl/guardian.ts
```diff
@@ -0,0 +1,89 @@
...
+// watcher synthesizes a VAA from an XRPL payment; this fetches it by
+// (chain, emitter, sequence) and polls until it is available.
+
+import type { Network } from "@wormhole-foundation/sdk-connect";
+
+export const DEFAULT_GUARDIAN_API = "https://api.testnet.wormholescan.io";
```
evan-gray: these are already defined here: https://github.com/wormhole-foundation/wormhole-sdk-ts/blob/065ac8d036be9408b41b8627415583bd5a303690/connect/src/config.ts#L15 - can those be used instead?

cli/src/xrpl/guardian.ts
```diff
@@ -0,0 +1,89 @@
...
+    if (!res.ok) return null;
+    const data = (await res.json()) as { vaaBytes?: string };
+    return data.vaaBytes ?? null;
+  } catch {
+    return null;
+  }
+}
+
+/** Poll the guardian API for a signed VAA until it appears or timeout. */
+export async function pollSignedVaa(opts: {
```
evan-gray: certainly we shouldn't need to reproduce VAA fetching logic. can we use https://github.com/wormhole-foundation/wormhole-sdk-ts/blob/065ac8d036be9408b41b8627415583bd5a303690/connect/src/whscan-api.ts#L159 ?

cli/src/xrpl/executorLayouts.ts
```diff
@@ -0,0 +1,141 @@
...
+import { deserialize, serialize } from "binary-layout";
+import { ethers } from "ethers";
+
+const DEFAULT_EXECUTOR_API = "https://executor-testnet.labsapis.com";
+
+const DEFAULT_EXECUTOR_APIS: Partial<Record<Network, string>> = {
+  Testnet: DEFAULT_EXECUTOR_API,
+};
+
+export function getDefaultExecutorApiForNetwork(network: Network): string {
```
evan-gray: this and the above consts are duplicated from executor.ts

cli/src/commands/xrpl/init.ts
```diff
@@ -0,0 +1,178 @@
...
+    builder: (yargs: any) =>
+      withCommon(yargs)
+        .option("admin", {
+          describe: "Admin XRPL r-address (registers peers / rotates admin)",
+          type: "string",
+          demandOption: true,
+        })
+        .option("initial-ticket", {
+          describe: "First ticket ID in the pre-allocated range",
+          type: "number",
```
evan-gray: I don't think it's necessarily safe for this to be a number. as a u64 it should be a bigint (or string as a parameter, then `BigInt()`ed)

cli/src/commands/xrpl/init.ts
```diff
@@ -0,0 +1,178 @@
...
+          demandOption: true,
+        })
+        .option("ticket-count", {
+          describe: "Number of tickets available",
+          type: "number",
```
evan-gray: likely safe as a number, but since its type is u64, should also be bigint

cli/src/xrpl/onboarding.ts
```diff
@@ -0,0 +1,177 @@
...
+/**
+ * Build the XRPLAppOnboarding payload (hex, no 0x prefix):
+ *   prefix("XRPL")[4] + admin[20] + app(left-padded)[32] + initial_ticket[8] +
+ *   ticket_count[8] + init_data
+ */
+export function buildOnboardingPayload(params: {
+  admin: string;
+  app: string;
+  initialTicket: number;
+  ticketCount: number;
```
evan-gray: tracking that these should be changed as well - u64 should never be represented as `number`

cli/src/xrpl/onboarding.ts
```diff
@@ -0,0 +1,177 @@
...
+  return n.toString(16).padStart(2, "0");
+}
+
+const U64_MAX = (1n << 64n) - 1n;
+
+/**
+ * Encode a non-negative integer that fits in a u64 as a big-endian 8-byte
+ * (16 hex char) value. Accepts a number, bigint, or decimal string.
+ */
+function u64Hex(n: string | number | bigint): string {
```
evan-gray: I recommend removing `number` here, as it lacks the precision for full u64 representation.

evan-gray (changes requested):

evan-gray (dismissed):

douglasgalico (dismissed):

evan-gray (approved):

douglasgalico (approved):

PR #915, opened 2026-06-25
fix: xrpl executor route tx-fetch

douglasgalico:


xrpl/ts/src/nttWithExecutor.ts
```diff
@@ -93,39 +99,62 @@ export class XrplNttWithExecutor<N extends Network, C extends XrplChains>
...
     ) as unknown as UnsignedTransaction<N, C>;
 
-    // After the NTT payment is submitted and confirmed, look up the result
-    // to compute the messageId = (ledgerIndex << 32) | txIndex
     if (!this.provider.isConnected()) {
       await this.provider.connect();
     }
     const senderAddr = sender.toString();
+
+    // The memo data uniquely identifies the payment we yielded (XRPL stores
```
evan-gray, line 107: is it unique if they sent two with the same payment? I would think the txBody / hash would uniquely identify it

douglasgalico: Hash would be the best, but we can't precompute while inside the generator as it wasn't signed yet - that would require us changing the way transfer generator works across all platforms.

An mid-point that can fix this "twin bug" is passing a second memo data with a UUID and using it as part of match

M-Picco: According to the watcher, the [memo data](https://github.com/wormhole-foundation/wormhole/blob/3888993f80d9a7e63742103eea6e482b4de87181/node/pkg/watchers/xrpl/parse.go#L828) could have the same info for two distinct transfers

Could we use the tx's `Sequence`? We'd need to do a `client.autoFill(tx)` before yielding to the client/user. However, I'm wondering if that'd introduce the risk of the client calling the autofill again, messing with the expected sequence

evan-gray: https://github.com/wormhole-foundation/native-token-transfers/issues/916#issue-4747890239

evan-gray (approved):

M-Picco (approved):

PR #917, opened 2026-06-26
Sui Rate Limit Fixes

mdulin2:
There were two bugs in the rate limit code: 

* Missing timestamp update on refill.
* Wrong scaling on outbound refill.

This PR fixes those bugs, and adds tests for the rate limiter.

### DIFF
--- a/sui/packages/ntt/tests/ntt_tests.move
+++ b/sui/packages/ntt/tests/ntt_tests.move
@@ -904,4 +904,1119 @@ module ntt::ntt_tests {
         test_scenario::return_to_address(admin, admin_cap);
         scenario.end();
     }
+
+
+    #[test]
+    fun test_transfer_with_refill() {
+        let (_, user_a, _, _) = ntt_scenario::test_addresses();
+        let mut scenario = test_scenario::begin(user_a);
+        ntt_scenario::setup(&mut scenario);
+
+        scenario.next_tx(user_a);
+
+        // Take state and clock
+        let mut state = ntt_scenario::take_state(&scenario);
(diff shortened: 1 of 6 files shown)


kcsongor (dismissed):

sui/packages/ntt/tests/ntt_tests.move
```diff
@@ -904,4 +904,1119 @@ module ntt::ntt_tests {
...
+                .borrow_inbound_rate_limit_mut();
+
+        let inbound_last_tx_after     = inbound.last_tx_timestamp();
+
+        // Outbound and inbound rate limit comparisons            
+        assert!(inbound_last_tx_after != inbound_last_tx_before);             
+        assert!(inbound_last_tx_after > inbound_last_tx_before); 
+        assert!(outbound_last_tx_after != outbound_last_tx_before);             
+        assert!(outbound_last_tx_after > outbound_last_tx_before); 
+        assert!(outbound_last_tx_after == inbound_last_tx_after);
```
djb15: Is it possible to be more precise around the assertions in this test? Since we already knew the rate limits were updated, but we didn't know that they were updated by the wrong amount

djb15 (commented):
The source changes look like we discussed and it's nice to see all the new tests. Just one question about whether we can make the test a bit more precise

sui/packages/ntt/tests/ntt_tests.move
mdulin2: Should be resolved in the latest commit.

djb15 (approved):

kcsongor (approved):

PR #918, opened 2026-06-29
Add peer check to the executeMsg (evm)

### LINKED ISSUE #890: evm: sourceNttManagerAddress is not validated when a partially attested message is released during a transceiver threshold decrease
Ordinarily `_verifyPeer()` in the attestation flow ties the sourceNttManagerAddress to the payload indirectly. However, if:
* a message has some but not all attestations AND
* the threshold decreases during transceiver managements, AND
* someone calls `executeMsg` for that message with a different sourceNttManagerAddress

then the message could be processed with an incorrect manager address.

This has implications in only two places:
* The event emitted on release
* `handleAdditionalPayload` -- this is integrator-specific logic which the core maintainers of this project do not have control of or visibility into.

Integrators relying on `sourceNttManagerAddress` while handling additional payloads should be aware that this data may not be trustworthy in the unlikely edge-cases described above.

mdulin2:
Performs the peer check on the `executeMsg()` flow for EVM. I added an internal function with the logic in order to not duplicate the `_verifyPeer()` check in two spots.

### DIFF
--- a/evm/src/NttManager/NttManager.sol
+++ b/evm/src/NttManager/NttManager.sol
@@ -41,7 +41,7 @@ contract NttManager is INttManager, RateLimiter, ManagerBase {
     using TrimmedAmountLib for uint256;
     using TrimmedAmountLib for TrimmedAmount;
 
-    string public constant NTT_MANAGER_VERSION = "2.0.0";
+    string public constant NTT_MANAGER_VERSION = "2.0.1";
 
     // =============== Setup =================================================================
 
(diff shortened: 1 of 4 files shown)


djb15:
Resolves https://github.com/wormhole-foundation/native-token-transfers/issues/890

djb15 (dismissed):
This looks good to me. Critically `executeMsgInner` has the correct visibility, `attestationReceived` still has the `whenNotPaused` modifier, and functionality is otherwise untouched

bemic (approved):
The change does not affect functionality. New test cases added. Visibility and modifiers are matching the original implementation. Looks good to me.

johnsaigle (changes requested):
I think this needs a manual rebase before it'll pass CI. I'm unable to do it in the front-end

johnsaigle (approved):
Changes look good -- the PR just needs help to get through CI

djb15 (approved):

PR #921, opened 2026-07-02
refactor: new ticket count and onboarding init payload

douglasgalico:


cli/src/commands/xrpl/reserve-tickets.ts
```diff
@@ -13,8 +13,8 @@ import {
 } from "../../xrpl/helpers";
 import { withCommon } from "./common";
 
-// XRPL caps an account at 250 tickets.
-const MAX_TICKETS = 250;
+// XRPL caps an account at 175 tickets.
+const MAX_TICKETS = 175;
```
johnsaigle: Can you provide some more context for this change, maybe in the PR description?

I'm unclear why this value is being modified. The docs seem to list 250 as the correct limit: https://xrpl.org/docs/concepts/accounts/tickets

douglasgalico: That's related to the Sequencer change at TicketEntry that grows from 27 -> 59 byes (as includes source_emitter). Then to keep under 10kIB MAX_PERMITTED_DATA_INCREASE for CPI create_account the TICKET_LEDGER_CAPACITY was reduced there from 250 to 170 (fixed the 175*)

PR #923, opened 2026-07-10
feat: json structured output

douglasgalico:


cli/src/output.ts
```diff
@@ -0,0 +1,74 @@
+// Structured-output mode for the NTT CLI.
+//
+// Enabled by `--json` (anywhere in argv) or `WL_NTT_JSON=1`. While active,
+// every stdout write is redirected to stderr — the only thing that lands on
```
johnsaigle, line 4: I think that solana.ts logs a private key, so this change means that the private key could get written to STDERR. This might not be a good idea in case the error string gets saved somewhere

douglasgalico: That's true, we can force emitting directly to stdOut for this case

kcsongor (dismissed):

cli/src/commands/push.ts
```diff
@@ -284,7 +286,17 @@ export function createPushCommand(overrides: WormholeConfigOverrides<Network>) {
           argv["dangerously-transfer-ownership-in-one-step"],
           overrides
         );
+        chainsTouched.push(chain);
       }
+      emitResult("push", {
```
johnsaigle, line 317: It looks like this will emit the OK true case even if some of the above cases of errors, e.g. the cases of "Solana wormhole transceiver can only be set on Solana chains" and errors from `registerSolanaTransceiver`.

To me this seems misleading though I don't know the full context of the feature. Maybe it would be better to review the above logic and avoid returning the success message if errors occur?

douglasgalico: good catch! Changed to accumulate these non-fatal errors and use the right exit code (and terminate process) if found.

johnsaigle: It looks like `pushDeployment()` still runs when there are errors above?

I think it makes sense to fail before continuing once any required registration/LUT step fails.

douglasgalico: adjusted!

johnsaigle (dismissed):
Looks OK from a security perspective

cli/test/solana.sh
```diff
@@ -236,7 +236,7 @@ else
 fi
 ntt status || true
 
-ntt push --payer "$keypair" --yes
+ntt push --payer "$keypair" --yes || true
```
johnsaigle: This hides the error completely and makes this line of the test always succeed. Can you explain why this is necessary? It seems like it could easily conceal a legitimate failure.

johnsaigle (changes requested):

cli/test/solana.sh
douglasgalico: Just following how we do on `ntt status` above, but can be improved

douglasgalico: added a better check

kcsongor (approved):

johnsaigle (approved):

